PR #2, opened 2021-12-01
install skeleton

Itxaka:


Itxaka:
@davidcassany ^^ :D

Itxaka:
I think this could be merged already as its mostly the skeleton plus a couple of utils for the installer and test, so this allows us to work in parallel filling the gaps, and it doesnt grow to a monstrosity

davidcassany (approved):

pkg/utils/common.go
```diff
@@ -0,0 +1,66 @@
...
+
+var (
+	Client HTTPClient
+)
+
+func init() {
+	Client = &http.Client{}
+}
+
+func GetUrl(url string, destination string) error {
```
mudler: suggestion: maybe better to take a `client` interface here. For example, the default above doesn't set a default timeout which would make call never timeout and they could hang indefinetly in case of network issues

pkg/utils/common.go
```diff
@@ -0,0 +1,66 @@
...
+	var source io.Reader
+	var err error
+
+	switch {
+	case strings.HasPrefix(url, "http"), strings.HasPrefix(url, "ftp"), strings.HasPrefix(url, "tftp"):
```
mudler, line 32: maybe we can just rely on something like https://github.com/cavaliergopher/grab ?

pkg/action/install.go
```diff
@@ -0,0 +1,111 @@
...
+	"fmt"
+	"github.com/rancher-sandbox/elemental-cli/pkg/utils"
+	"github.com/spf13/viper"
+	"github.com/zloylos/grsync"
+	"os"
+	"strings"
+	"time"
+)
+
+type InstallAction struct {
```
mudler, line 30: I'm ok in general as this is a first write up, but I'd start to thinker on how to approach it in general. 

As we will have a lot of commands, we might want to start to think on how we want to structure the core.

.. that might be even a simple loop that execute a bunch of Run ops by respecting an interface, or either a "core" struct `Elemental` that does the operations and take for e.g. configs in a functional interface style, or.. :)

This has also other implications, for e.g. logging, how do we plan to handle it?

cmd/install.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(installCmd)
+	installCmd.Flags().StringP("configDir", "e", "/etc/elemental/", "dir where the elemental config resides")
+	installCmd.Flags().StringP("dockerImage", "d", "", "Install a specified container image")
+	installCmd.Flags().StringP("cloudInit", "c", "", "Cloud-init config file")
```
mudler: not sure, it feels a bit weird having cli flags in camel case, we always stick to kebab-case in cos-toolkit.. why suddenly change?

Itxaka: well, mainly so they can be overriding via env vars automatically without having to mangle our way around. This way, if you set `ELEMENTAL_CLOUDINIT=whatever` it will override the `cloudInit` flag thanks to the `viper.AutomaticEnv()`

Otherwise, we need to manually load those env vars into the proper flags and we would have a disconnect between flag-name <-> var name

pkg/utils/common.go
Itxaka: not sure if for a couple of lines, we should bring a whole package (not updated in a year!)

I mean if our download/get stuff grow, Im ok, but for the current use?

pkg/utils/common.go
Itxaka: good point!

pkg/action/install.go
Itxaka: Well, something like an Action interface that has X methods? (setup, run, post??)

Sounds good to me. Take into mind that I have not worked on big go programs from scratch, so I dont have any idea of the best approaches in things like that, so keep those suggestions incoming.

For example, I just noticed that we should not set much of this stuff in the action struct, but probably pass a config object which contains all the decent data, so others can re utilize this lib to implement their own stuff easily.

I even just saw that Im using viper OUTSIDE the cmd!! that its terrible and should be banned!

cmd/install.go
mudler: If that's the main reason I'd rather keep kebab-case if possible, it is more 'classical'.. It feels a bit bad to force CLI UX just for a technical reason, in `luet` the above I think was handled here: https://github.com/mudler/luet/blob/79e98af604828a1ac9755006664b644718c3337a/cmd/util/config.go#L81

pkg/action/install.go
mudler: > For example, I just noticed that we should not set much of this stuff in the action struct, but probably pass a config object which contains all the decent data, so others can re utilize this lib to implement their own stuff easily.

There are several approaches that can be taken there, and a lot examples on the net, for example, if we look at how `kubernetes` api code is structured, most of the client public functions in the packages takes an Option, while similar methods can share the same Options too. 

For example, a type `SearchOptions{ Order string, Limit string }` can apply to `podlist.Search(pod, namespace, opts)` and `deployment.Search(name, namespace, opts)`, but as you can see it depends on what functionality they should bring (search in this case). 

Kubernetes code is mostly following Domain Driven Design principles, so we could look at the bigs and do the same and go into that direction. 

I'm _personally_ lately mixing a bit of functional with DDD without being too much pedantic, it really depends on how much can grow the codebase but seems a good balance and a really flexible approach (constructors are really tight and no need to touch them once you write them up from the first time!)

Anyway, don't want to throw too much into the mix, as we should go baby steps always!

pkg/action/install.go
```diff
@@ -0,0 +1,111 @@
...
+				state.Speed,
+			)
+			<- time.After(time.Second)
+		}
+	}()
+	if err := task.Run(); err != nil {
+		return err
+	}
+	// 2 - if we got a cloud config file get it and store in the target
+	if viper.GetString("cloudInit") != "" {
```
mudler: Now I see why the camel case :) 

It is more common to pass by an actual structure like a config or an opts instead of calling viper directly. Tying them together makes the code very fragile, better let viper unmarshal that data into a custom struct, as it supports it.

 In luet we are using it here: https://github.com/mudler/luet/blob/ba0551caab45d334b0694f5778b6f6c77cbc0ec6/cmd/util/config.go#L95 to unmarshal the configuration into a config struct. The config struct should just map in the annotations the relative fields (see the Config struct in luet as an example)

cmd/install.go
Itxaka: sounds good to me, I didnt knew it was that easy!

Itxaka: wait, Im not sure its that easy. The thing is env vars dont accept `-` so it would be substituted by something else. But that would still not match the actual keys directly becuase we would be reading `whatever_whatever` and trying to map it to `whatever-whatever`

unless that is exactly what that does....I need to test that then :)

pkg/action/install.go
Itxaka: done!

cmd/install.go
Itxaka: done!

pkg/action/install.go
Itxaka: kind of fixed with the config struct passed aorund instead of using viper to access those values.

pkg/utils/common.go
Itxaka: done!

pkg/action/install.go
```diff
@@ -18,22 +18,20 @@ package action
 
 import (
 	"fmt"
+	"github.com/rancher-sandbox/elemental-cli/pkg/types/v1/config"
```
mudler: :100:

pkg/utils/config.go
```diff
@@ -10,39 +11,36 @@ import (
...
 	viper.AddConfigPath(configDir)
 	viper.SetConfigType("yaml")
 	viper.SetConfigName("manifest.yaml")
 
 	// If a config file is found, read it in.
-	err := viper.ReadInConfig()
-	if err != nil {
-		return err
-	}
+	_ = viper.ReadInConfig()
```
mudler: I guess the error skip here is a leftover unless is a best-effort action? anyway this is equivalent to:

```golang
viper.ReadInConfig()
```

pkg/utils/config.go
```diff
@@ -10,39 +11,36 @@ import (
...
-	viper.AddConfigPath(cfg)
+	viper.AddConfigPath(configDir)
 	viper.SetConfigType("yaml")
 	viper.SetConfigName("config.yaml")
 	// If a config file is found, read it in.
-	err := viper.ReadInConfig()
-	if err != nil {
-		return err
-	}
+	_ = viper.ReadInConfig()
```
mudler: ditto

pkg/utils/config.go
```diff
@@ -10,39 +11,36 @@ import (
...
 
 	// Set the prefix for vars so we get only the ones starting with ELEMENTAL
 	viper.SetEnvPrefix("ELEMENTAL")
 
 	viper.AutomaticEnv() // read in environment variables that match
-	return nil
+	// unmarshal all the vars into the config object
+	_ = viper.Unmarshal(cfg)
```
mudler: ditto

pkg/utils/config.go
Itxaka: umm, yeah I was supposed to do something with this error, mainly log it somewhere as we dont want to fail so early during loading config, we want to continue, but also we want to store those warnings somewhere, but as there is no logging facility yet....

let me remove those.

mudler (approved):
Looks good here!

pkg/utils/config.go
```diff
@@ -1,66 +1,46 @@
...
+	viper.AddConfigPath(configDir)
 	viper.SetConfigType("yaml")
 	viper.SetConfigName("manifest.yaml")
 
 	// If a config file is found, read it in.
-	err := viper.ReadInConfig()
-	if err != nil {
-		return err
-	}
+	viper.ReadInConfig()
```
mudler, line 21: Last thought here. we are mixing pkg/utils with cmd/utils. usually Viper helpers are contained to a `cmd/` folder in the project repo, as they are not really tight to the underlying code logic, and also we will have more viper helpers for sure :)

Itxaka: good point, maybe I should move the logic into the cmd/config folder so they are part of the cmd stuff and have nothing to do with our pkg

Itxaka: done!

PR #3, opened 2021-12-03
Work in progress partitioner module

davidcassany:
Adds a partitioner module, it mostly wraps `parted` and  `mkfs`
utilities to provide an API to add/modify partitions and format them
into a disk.


davidcassany:
I am doing this PR in case anyone wants/needs to make use of it before next Thursday, this is still under heavy development. I have not even compiled it (I am not that far though 😉 ).

This basically provides a `Disk` structure with the following relevant methods:

* NewPartitionTable
* AddPartition
* FormatPartition
* ExpandLastPartition

This covers most if not all installer needs in terms of partitions settings.

Note the implementation is heavily based in yip's layout plugin. The layout plugin from yip wraps around `sgdisk` utility, here I switched to `parted` to also support msdos partition tables. The API is almost the same as in yips plugin, my idea is that, in the future once this is settled, we can actually use this as a library in yip, so we concentrate the logic in elemental client and also extend layout plugin to `msdos` partition table types.

At first I thought about absolutely copy paste `sgdisk` implementation from layout plugin and extend it to support `msdos`, but I realized there is no real counter part of `sgdisk` for `msdos` partition tables (`sfdisk` is a pain to wrap).

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/onsi/gomega v1.17.0
+	github.com/sirupsen/logrus v1.8.1 // indirect
```
Itxaka, line 7: So we setting on logrus? +1 for me, I didnt add any logging at all due to not knowing what to use :D

pkg/partitioner/disk.go
```diff
@@ -0,0 +1,374 @@
...
+	"errors"
+	"fmt"
+	log "github.com/sirupsen/logrus"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+)
+
+type Runner interface {
```
Itxaka: you can just import the RealRunner from the v1 package here. no need to redefine the interface

tests/mocks/runner_mock.go
```diff
@@ -22,3 +25,37 @@ func (r TestRunner) Run(command string, args ...string) ([]byte, error) {
 	out, err := cmd.CombinedOutput()
 	return out, err
 }
+
+type TestRunnerV2 struct {
```
Itxaka, line 29: Cant we expand the other runner instead? or does that break the other tests?

Itxaka: Im very much like this one more!

pkg/partitioner/disk.go
```diff
@@ -0,0 +1,374 @@
...
+	dev.parts = partitions
+	dev.label = label
+	return nil
+}
+
+// Size is expressed in MiB here
+func (dev *Disk) CheckDiskFreeSpaceMiB(minSpace uint) bool {
+	freeS, err := dev.GetFreeSpace()
+	if err != nil {
+		log.Warningf("Could not calculate disk free space")
```
mudler, line 82: for the logging maybe we can also accept a logging interface and use that across the code, so we can go back at this later to have more contextualized logs

mudler (approved):

mudler:
> Note the implementation is heavily based in yip's layout plugin. The layout plugin from yip wraps around `sgdisk` utility, here I switched to `parted` to also support msdos partition tables. The API is almost the same as in yips plugin, my idea is that, in the future once this is settled, we can actually use this as a library in yip, so we concentrate the logic in elemental client and also extend layout plugin to `msdos` partition table types.

Yes definetly, we can just reuse this instead of the layout plugin when instantiating the cloud-init lib

go.mod
davidcassany: Oh, I accidentally used that... I thought you were also already using that, so I just used it. Probably I did not check it properly and I took it from `yip` not from the skeleton you did. My rebase had plenty of conflicts, I guess I overlooked that.

tests/mocks/runner_mock.go
davidcassany: I did not change the other as I am not sure I understand it 😅 . My intention is also trying to get all the flexibility and functionality we need in a single mock. But first I wanted to have what I need, second round trying to merge both into a single interface implementation.

pkg/partitioner/disk.go
davidcassany: Yes makes sense, I should be capable to import the Runner interface from v1, so there is a single definition.

tests/mocks/runner_mock.go
Itxaka: Feel free to use that v2 and then we can adapt the other code, this seems a much better implementation!

Itxaka (approved):

PR #4, opened 2021-12-03
Add a few more utility functions, move mock into its own file

Itxaka:


pkg/utils/common_mock.go
```diff
@@ -0,0 +1,26 @@
+package utils
+
+import (
+	"io"
+	"net/http"
+)
+
+var MockClientCalls []string
```
mudler: better place mock unit and fixtures separated from the code, maybe under `test/` at top level?

pkg/utils/common.go
```diff
@@ -52,3 +62,51 @@ func GetUrl(client v1.HTTPClient, url string, destination string) error {
 
 	return nil
 }
+
+func selinuxRelabel(target string, fs afero.Fs) error {
```
mudler: somehow feels wrong to have SELinux relabel as a common helper, maybe under an hypotetical `file.go`?

pkg/utils/common.go
```diff
@@ -52,3 +62,51 @@ func GetUrl(client v1.HTTPClient, url string, destination string) error {
...
+
+	return err
+}
+
+func commandExists(command string) bool {
+	_, err := exec.LookPath(command)
+	return err == nil
+}
+
+func setupStyle(config *v1.RunConfig, fs afero.Fs) {
```
mudler: from the signature most likely `setupStyle` should go in v1.RunConfig instead.

pkg/utils/common.go
Itxaka: yeah, utils is usually where I dump all the stuff that I have no idea what to do with :D

Will move it to a file.go!

pkg/utils/common.go
Itxaka: yes! good catch!

pkg/utils/common_mock.go
Itxaka: Good one, I just looked around different go projects and they all seem to keep the mock around with the *_test.go files, which was strange to me. But go testing having the files around is also weird to me :D

Will move it to tests! I wonder if it will be worst for importing it, but it should not matter I think.

Itxaka:
@mudler @davidcassany @dragonchaser ready to review!

pkg/types/v1/runner.go
```diff
@@ -0,0 +1,14 @@
+package v1
+
+import "os/exec"
+
+type Runner interface {
```
davidcassany, line 5: Good, Once this is merged I will adapt my PR to make use of this Runner interface, I was also providing a similar one.

pkg/types/v1/config_test.go
```diff
@@ -0,0 +1,28 @@
+package v1
```
mudler:
```suggestion
package v1_test
```

pkg/utils/common_test.go
```diff
@@ -19,70 +19,48 @@ package utils
 import (
```
mudler, line 19: Similarly here, if this is a testing file, it should be `utils_test`

tests/mocks/runner_mock.go
```diff
@@ -0,0 +1,24 @@
...
+	"os/exec"
+)
+
+type TestRunner struct{}
+
+func (r TestRunner) Run(command string, args ...string) ([]byte, error) {
+	var cs []string
+	// If the command is trying to get the cmdline call the TestHelperBootedFrom test
+	// Maybe a switch statement would be better here??
+	if command == "cat" && len(args) > 0 && args[0] == "/proc/cmdline" {
```
davidcassany, line 14: I am not sure I understand this mock... this is actually running the command, isn't it? IMHO would be nice to extend the `TestRunner` struct to track inputs and include outputs in some way and then do not execute anything within the `Run` command. Anyway this is something that can be done later. I am going to need such thing for the `partitioner` to code proper unit tests.

mudler:
There seems to be a mixture of testing functionalities exported as part of the package, which I don't think is wanted here. That would pull the testing part when including the packages

pkg/types/v1/config_test.go
```diff
@@ -0,0 +1,28 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/spf13/afero"
+	"testing"
+)
+
+func TestSetupStyleDefault(t *testing.T) {
+	RegisterTestingT(t)
+	fs := afero.NewMemMapFs()
+	c := RunConfig{}
+	c.setupStyle(fs)
```
mudler: That won't work anymore I think as now it's in another package :)

pkg/utils/common_test.go
Itxaka: :+1:

pkg/types/v1/config_test.go
Itxaka: fixed!

davidcassany (approved):
Looks good to me, I did not dive into the details of the config implementation. I like having versioned types within the code structure.

tests/mocks/runner_mock.go
Itxaka: This is...interesting to say the least.

Its not really running the command but calling go with the `-test.run=TestHelperBootedFrom` parameter which will run the test function named there! So this (if matches) will actually run the `TestHelperBootedFrom` test! 

That is why there is the extra test on the common_test.go:

```
// TestHelperBootedFrom will be called by the TestRunner when running the BootedFrom func as it
// Matches the command + args and return the proper output we want for testing
func TestHelperBootedFrom(*testing.T) {
	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
		return
	}
	defer os.Exit(0)
	fmt.Println("I_EXPECT_THIS_LABEL_TO_EXIST") // Mock that label
}
```

So this will output that line as output, and that output will be returned. Its quite strange but seems to be the way of mocking any exec.CMD on golang, for what I understood without resorting to use full mocks/stubs. Probably @mudler can explain it better than me.

mudler (approved):

mudler:
looks good here!

PR #6, opened 2021-12-10
Move docopy out of the installer

Itxaka:
It doesnt neccesary has to be attached to it and can be reused by other
components

### DIFF
--- a/pkg/utils/copy.go
+++ b/pkg/utils/copy.go
@@ -1,0 +1,68 @@
+package utils
+
+import (
+	"fmt"
+	v1 "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/zloylos/grsync"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+)
+
+func DoCopy(config *v1.RunConfig) error {
+	fmt.Printf("Copying cOS..\n")
+	// Make sure the values have a / at the end.
+	var source, target string
+	if strings.HasSuffix(config.Source, "/") == false {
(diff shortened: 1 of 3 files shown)


pkg/utils/copy.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	v1 "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/zloylos/grsync"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+)
+
+func DoCopy(config *v1.RunConfig) error {
```
mudler, line 13: the `DoCopy` name is pretty generic, but we are copying cOS here..

Just thinking out loud, maybe we can have a cOS struct so we can:

```
cos.Copy(dst)
cos.SetActive(..)
cos.SetPassive(..)
cos.SetRecovery(..)
``

mudler (approved):

pkg/utils/copy.go
Itxaka: yes, I hate the name, but Cant come up with a decent naming for this.

CopyCos? I mean, in the original this is only done on the install, so there is no reusability, but there is also nothing that should tie this to the installer.

Maybe the cos struct is the way to go...

Itxaka: I think we can merge this, and see where it lead us, but Im thinking that yes, the cos struct may be the better way of linking this somewhere.

mudler: maybe we can just live with the cos utilities inside `cos.go` or a comment on the top so we don't get confused

PR #7, opened 2021-12-10
Rework runconfig to use options on init

Itxaka:
On sync with chroot for example, hopefully with everything going forward so we only have to set it on struct init


mudler (approved):
Love it!

pkg/types/v1/config.go
```diff
@@ -26,6 +28,28 @@ const (
 	BOOT = "boot"
 )
 
+type RunConfigOptions func(a *RunConfig) error
+
+func WithFs(fs afero.Fs) func(r *RunConfig) error {
```
mudler, line 33: Maybe we should start to think to document (with comments) the options? :thinking:

Itxaka: good point, godoc or plain old docs?

mudler: godoc please :) ! :rocket:

PR #9, opened 2021-12-12
Add Grub struct and install methods

Itxaka:
Requires https://github.com/rancher-sandbox/elemental-cli/pull/8

### DIFF
--- a/pkg/utils/grub.go
+++ b/pkg/utils/grub.go
@@ -9,16 +9,16 @@ import (
 )
 
 type Grub struct {
-	disk string
+	disk   string
 	config v1.RunConfig
-	fs afero.Fs
+	fs     afero.Fs
 	runner v1.Runner
 }
 
-func NewGrub(config v1.RunConfig, opts...GrubOptions) *Grub{
+func NewGrub(config v1.RunConfig, opts ...GrubOptions) *Grub {
 	g := &Grub{
 		config: config,
-		fs: afero.NewOsFs(),
+		fs:     afero.NewOsFs(),
 		runner: &v1.RealRunner{},
 	}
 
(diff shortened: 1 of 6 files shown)


Itxaka:
now that https://github.com/rancher-sandbox/elemental-cli/pull/7 is in, this needs some modifications to set the default values on config new instead of the grub struct directly :D

pkg/utils/grub.go
```diff
@@ -0,0 +1,122 @@
...
+	v1 "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/spf13/afero"
+	"runtime"
+	"strings"
+)
+
+type Grub struct {
+	disk string
+	config v1.RunConfig
+	fs afero.Fs
```
davidcassany: Interesting, I guess I should have used that in disk.go to manipulate temporary files and run proper tests for it... 👍🏽

davidcassany (approved):

pkg/utils/grub.go
```diff
@@ -0,0 +1,122 @@
...
+	runner v1.Runner
+}
+
+func NewGrub(config v1.RunConfig, opts...GrubOptions) *Grub{
+	g := &Grub{
+		config: config,
+		fs: afero.NewOsFs(),
```
davidcassany: If we allow to pass it as a parameter then we will gain the possibility of passing a FS in memory for tests, isn't it? So then it will be easier to run unit tests with a fake FS.

Itxaka: you can, you just need to use it like so:

```
	grub := NewGrub(
		config,
		WithRunnerGrub(&runner),
		WithFSGrub(fs),
	)
```

As shown on the tests.

This is just setting the default value if nothing is passed, then we iterate over any options passed as optional args (in `opts`) and set the proper fs is `WithFSGrub` is passed.

Blame @mudler for showing me this :P

davidcassany: Ohhhh I see, I am applying the same into the partitioner then :)

PR #12, opened 2021-12-13
Main install

Itxaka:


pkg/types/v1/config.go
```diff
@@ -64,29 +64,41 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
...
+	NoFormat     string `yaml:"no-format,omitempty" mapstructure:"no-format"`
+	ActiveLabel  string `yaml:"ACTIVE_LABEL,omitempty" mapstructure:"ACTIVE_LABEL"`
+	PassiveLabel string `yaml:"PASSIVE_LABEL,omitempty" mapstructure:"PASSIVE_LABEL"`
+	Force        bool   `yaml:"force,omitempty" mapstructure:"force"`
+	PartTable    string
+	BootFlag     string
+	StateDir     string
+	GrubConf     string
+	Logger       Logger
+	Fs           afero.Fs
```
Itxaka, line 124: Seems much more useful to have this exported so consumers of the config can just use that Fs like they use the logger.

I would even suggest having the syscall and runner in here as well, so everything has one source of those interfaces, then tests can still override it with the `WithFS/WithRunner/WithXXX` methods.

thougths @mudler ? Any wrongdoings into having those in the config?

davidcassany: To me it makes sense having access to config elements, however I am not sure if form a nit-picky perspective wouldn't it be better having a getter method instead of exposing it, so we enforce it can't be changed at any time by anyone.

Anyway I am fine with this approach, I am not sure being super strict here pays off.

pkg/partitioner/disk.go
```diff
@@ -48,6 +48,11 @@ func (dev Disk) GetLabel() string {
 	return dev.label
 }
 
+func (dev Disk) IsValid() bool {
+	_, err := os.Stat(dev.device)
```
davidcassany: I'd suggest to name it `Exists` rather than `IsValid`. IsValid may give the impression it does some checks over the partition table or something similar.

Also we should use `afero.Fs` here as this is/will be introduced in #13

davidcassany (approved):
Looks good, just  a couple of comments

pkg/types/v1/config.go
Itxaka: One of the good things of this approach is that once you set a config, then it can be passed to everything, including grub and chroot, which is what Im reworking currently, so no more passing fs all the time around, just pass a config and use that config to get the fs/runner/logger/whatever.

It really makes the code much cleaner and simpler!

Having getters may be cleaner indeed!

Itxaka:
Re-requesting reviews as this has quite changed.

In fact, moving it out of draft cause its growing and I dont want to end up with a 2000 lines change :D

cmd/install.go
```diff
@@ -34,7 +35,13 @@ var installCmd = &cobra.Command{
 		logger := logrus.New()
 		logger.SetOutput(os.Stdout)
 
-		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), logger)
+		path, err := exec.LookPath("mount")
```
davidcassany, line 38: Nit: I am wondering if it would make sense having a LookPath method as part of our Runner interface (or a new one). `exec` package for testing adds quite some complexity and having the availability to pass those close to OS functions wrapped in some interface is really useful for unit testing. Little functions like `LookPath` could can become a nightmare to run proper tests.

In this case it is not a big deal, as we can assume any host running the suit has the `mount`, but I am thinking towards other utilities. For instance now I am thinking it would be good to have some utility existence validation in `partitioner` module for proper error management (`lsblk`, `parted`, `partprobe`, `blkid`...), but if I add `exec.LookPath` then  any host running the tests needs to have these utilities in PATH (probably they are not on non privileged users or they are not installed).

pkg/types/v1/config.go
```diff
@@ -57,36 +81,55 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
...
 	// Set defaults if empty
 	if r.GrubConf == "" {
 		r.GrubConf = "/etc/cos/grub.cfg"
 	}
 	if r.StateDir == "" {
 		r.StateDir = "/run/initramfs/cos-state"
 	}
+
+	if r.ActiveLabel == "" {
+		r.ActiveLabel = "COS_ACTIVE"
```
davidcassany: I doubt we should hard code this defaults here. I'd rather prefer having the defaults written all in a single place, I am not sure how to do it now, but I'd like having some sort of structure for defaults. Would it make sense having some sort of static DefaultRunConfig which gets copied in `NewRunConfig` method as a boiler plate where all `RunConfigOptions` are then applied on top. So we have apply `RunConfigOptions` over a fully qualified `RunConfig` prefilled with all defaults.

pkg/cos/cos.go
```diff
@@ -0,0 +1,130 @@
+package cos
```
davidcassany: Nit: Is `cos` the right name for it? I don't want to start again speculations about naming, but probably we can find some package name that does not relate on _product_ but relates to functionality. Said that, I am fine with `cos` as a package name as we are copying the container image with a full OS here.

cmd/install.go
Itxaka: yes, agreed, this should be part of something that we can control

pkg/types/v1/config.go
Itxaka: We could move this into a constants package and have the constants for the whole thing?

pkg/cos/cos.go
Itxaka: Im bad with names, this was suggested by @mudler some time ago and I just went with it. I mean, its related to the install of cOS so it kind of makes sense as its closely related to cos specific things...but it could be called utils...but we already have a utils package for different utils...

I hate naming stuff.

pkg/types/v1/config.go
Itxaka: done!

pkg/cos/cos.go
davidcassany: I also hate it... I just thought that probably `cos` references in code might be something to avoid as it looks like this could be renamed and become legacy at mid term. Probably another option is to call it `elemental` (or some abbreviation of it) so it refers to core specific functions of the `elemental-cli`, after all this utility is called `elemental-cli`.

cmd/install.go
davidcassany: to me this could be part of the Runner interface, after all this will mostly be used to verify tools are available.

pkg/cos/cos.go
Itxaka: done!

cmd/install.go
Itxaka: Yes, and that would also make it possible to test properly this stuff as we can control whatever we return the proper tools or what to do if some binaries are missing from the system!

Will probably do it on a different path thought, when merging the runners maybe

Itxaka:
No point into making this PR bigger that what it is, now that the tasks are separated

pkg/constants/constants.go
```diff
@@ -0,0 +1,10 @@
+package constants
```
davidcassany, line 1: Not for now, just to keep in mind, but I the longer run having `constants` and `defaults` as different packages could make sense. Currently there are only few values so it doesn't make much of a difference for now.

Itxaka: yep, good point, although we'll cross that bridge when we(if we?) reach it :)

PR #13, opened 2021-12-13
Follow similar code practices as in utilities package

davidcassany:

### DIFF
--- a/pkg/partitioner/disk.go
+++ b/pkg/partitioner/disk.go
@@ -4,8 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
-	log "github.com/sirupsen/logrus"
-	"os"
+	"github.com/spf13/afero"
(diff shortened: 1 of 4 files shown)


davidcassany:
I added a `NewLogger` function that returns a plain sirupsen/logrus instance. This way we can initialize a logger from the our own utilities by default without depending on anything beyond our own types.

@Itxaka @mudler do you think it makes sense merging the partitioner into the `utils` package? I have mixed feeling with this idea, from one side having everything in one place makes it easier to follow the equivalent coding practices and with a single import you get everything. From the other side if the client utilities grow a lot I'd like not to have everything thrown in a huge package. What you think? Do you have any better argumentation towards one options? mines are pretty weak 😅

Itxaka:
I think it doesnt, as its meant to be used in different packages and would probably grow in complexity over the time, much more than the rest of the utils.

In fact, if some of the utils grow too much, I would be in favor of splitting into its own packages

Itxaka (approved):

PR #15, opened 2021-12-14
More testing

Itxaka:


Itxaka:
isnt codecov supposed to post in the PR?

pkg/types/v1/logger.go
```diff
@@ -24,15 +24,8 @@ type Logger interface {
...
-func NewLogger(opts ...LoggerOptions) Logger {
-	logger := log.New()
-
-	for _, opt := range opts {
-		if err := opt(logger); err != nil {
-			return nil
-		}
-	}
-	return logger
+func NewLogger() Logger {
```
davidcassany, line 27: I guess this is only to increase the test coverage right? Fine with me, we can add it again once we need it. I guess at some point we will need to be able to specify a log file and stuff like that, but there is no need to code it now.

tests/mocks/runner_mock.go
```diff
@@ -8,9 +8,14 @@ import (
 	"strings"
 )
 
-type FakeRunner struct{}
+type FakeRunner struct {
+	ErrorOnCommand bool
```
davidcassany, line 12: Nice! this is something we can add to the other interface I did. Once I finish with the layout creation in installer (I hope tomorrow there will be a PR ready to review) I'll probably spent some time trying to merge those two fake Runner  interfaces keeping all functionality.

davidcassany (approved):

pkg/types/v1/logger.go
Itxaka: Yes indeed, afaik the options were not used anywhere so no point in having them if they are not used, and it's a piece of code that we have all around so it should be easy to reproduce if needed.

PR #16, opened 2021-12-14
Cloud init

### LINKED ISSUE #942: `elemental cloud-init`
- [x] reuse `yip` as a library
- [x] Cut a new elemental-cli version (https://github.com/rancher-sandbox/elemental-cli/releases/tag/v0.0.2)
- [ ] Integrate `elemental cloud-init` in place of `yip` inside `cOS-toolkit` #976

mudler:
It required few changes on `yip` side (https://github.com/mudler/yip/commit/09900a173000749db08ef4504ed0f8a5ca211d25) to inject a logger, but now everything should be good for being used as-is.

```
❯ cat /home/mudler/_git/yip/example.yaml | ./elemental-cli cloud-init -s test -
INFO[0000] Running stage: test                          
INFO[0000] Executing stages:
   test:
     - files:
        - path: /tmp/foo
          content: |
                    test
          permissions: 0777
          owner: 1001
          group: 100
       commands:
        - echo "test" 
INFO[0000] Applying '' for stage 'test'. Total stages: 1 
INFO[0000] Processing stage step ''. ( commands: 1, files: 1, ... ) 
INFO[0000] Command output: test                         
INFO[0000] Stage 'test'. Defined stages: 1. Errors: false 
INFO[0000] Done executing stage 'test'    
```
Fixes https://github.com/rancher-sandbox/cOS-toolkit/issues/942.


mudler:
:shrug: for coverage, yip is having it's own test suite, no sense to copy tests here :thinking:

Itxaka (commented):
just like that? You make it seem simple af :)

Itxaka:
oh what, so you get codecov only if the branch is in the same repo?

mudler:
> oh what, so you get codecov only if the branch is in the same repo?

mmm it shouldn't I think..

Itxaka:
I told you I hate codecov, for a reason. It doesnt like me, I dont like it back ¬_¬

mudler:
so I've tried to have a look but tests for this would make sense in an integration test. IF we want to run integration tests as golang code, then that would mean to turn our private funcs into global, which is something, well..

Anyhow, maybe we can think in long term for an integration test suite or in golang, or rather in bash for simple acceptance tests, what do you think? but that hypothetically should run on quite complex scenarios (like the installer should run maybe in a cos container?, ...)

Itxaka:
Agreed. But doing integration test in here seems a bit far away in the future, doesn't it?

Not to mention that this will by tested as part of the cos-toolkit full integration tests so we should be covered by that, maybe expand a bit in there?

Or is the plan to drop some of the installer tests from cos and run them here instead? Or both?

I think running them here would alleviate some of our ci issues on cos, as we would have to treat this package as we do other packages, like a dependency and we expect it to be tested(i.e. jq or yq)

But on the other side, it would be crazy not to test the full installer under our environment....


I'm not sure to be honest, we should probably have a small team tall regarding the best planning for the tests going forward. Currently on cosign and mtree plugins for example are unit tested in their own repos and integration tested on  cos-toolkit.

pkg/types/v1/cloud-init-runner.go
```diff
@@ -0,0 +1,33 @@
+package v1
+
+import (
+	"github.com/mudler/yip/pkg/executor"
+	"github.com/mudler/yip/pkg/plugins"
+)
+
+// CloudInitRunner returns a default cloud init executor with the Elemental plugin set.
+// It accepts a logger which is used inside the runner.
+func CloudInitRunner(l Logger) executor.Executor {
```
Itxaka, line 10: Maybe we can add a simple test to make sure that this return the executor we want with the proper plugins?

I know I know, its kind of dumb to test that but it both increases coverage for something that should not really have coverage and makes sure the "upstream" `WithPlugins` is working

Itxaka: Otherwise, we can just add this file to codecov ignore files stanza?

cmd/cloud-init.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/mudler/yip/pkg/schema"
+	v1 "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/twpayne/go-vfs"
+)
+
+// cloudInit represents the cloud-init command
+var cloudInit = &cobra.Command{
```
Itxaka, line 33: I would say that we should add tests here to make sure we dont break it further down the road (I have done it, break cmd commands with unrelated changes) but I know that testing `Command` is a PITA so feel free to ignore this :D

Itxaka (dismissed):

pkg/types/v1/cloud-init-runner.go
mudler: > Maybe we can add a simple test to make sure that this return the executor we want with the proper plugins?
> 
> I know I know, its kind of dumb to test that but it both increases coverage for something that should not really have coverage and makes sure the "upstream" `WithPlugins` is working

heh.. I thought at that but it felt really stretched. Upstream have a test suite for all of this, so it seems we just double tests here.

On the other hand I do share the coverage feeling, but well..

Another Idea is, I can set the default plugin stack as a var, and check in the test if the cloudrunner by default use that. So we don't accidentally add plugins to the function, but that's just moving things around..

mudler: besides, the executor plugins are private, so no way to check that

Itxaka: nah, I dont think it makes sense to go that far, if it doesnt make sense lets not do it :+1:

pkg/types/v1/cloud-init-runner_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	. "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/twpayne/go-vfs/vfst"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// unit test stolen from yip
+var _ = Describe("Executor", func() {
```
Itxaka: My comments disappeared, not sure why, rewriting it again, sorry if its doubled somehow!

Shouldnt we maintain the same tests struct as in the other tests? This vastly differs from the rest of the testing in the codebase. Not saying that its better or worse, just saying that we should have a similar codebase when we can.

pkg/types/v1/cloud-init-runner_test.go
```diff
@@ -0,0 +1,73 @@
...
+	Context("Loading entities via yaml", func() {
+		logger := logrus.New()
+		logger.SetOutput(ioutil.Discard)
+
+		runner := CloudInitRunner(logger)
+		testConsole := console.NewStandardConsole(console.WithLogger(logger))
+
+		It("Run cloud-init files", func() {
+
+			fs2, cleanup2, err := vfst.NewTestFS(map[string]interface{}{})
```
Itxaka, line 29: Cant we use afero here like we do in the rest of the tests?

pkg/types/v1/cloud-init-runner_test.go
mudler: True.. but have to admit I'm really used to ginkgo way of defining tests... why not using the same cOS way (which is ginkgo defaults..) and be consistent here too?

pkg/types/v1/cloud-init-runner_test.go
mudler: I'll have a look, not sure

pkg/types/v1/cloud-init-runner_test.go
Itxaka: well...pure practicality to be honest, Im not looking forward to redo the whole test stuff into ginkgo format :rofl: 

We should also move the test and stuff into the test directory then...

cmd/cloud-init_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"bytes"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("cloud-init", func() {
+	Context("execution", func() {
+		When("invoked with inline yaml", func() {
+			It("executes command correctly", func() {
```
Itxaka: nice!

pkg/types/v1/cloud-init-runner_test.go
mudler: the ginkgo format has several advantages tho... besides readability :) BeforeEach, AfterEach and AfterSuite.. everything well tight to avoid spaghetti in tests ;)

mudler: > We should also move the test and stuff into the test directory then...

Yep, that sound more the canonical way of doing it

Itxaka: yes, I was missing that dearly while writing tests

pkg/types/v1/cloud-init-runner_test.go
mudler: Sadly we can't:

```
cannot use afero.NewOsFs() (value of type afero.Fs) as vfs.FS value in argument to runner.Run: wrong type for method Create (have func(name string) (github.com/spf13/afero.File, error), want func(name string) (*os.File, error))
```

Afero is not strictly following the fs implementation it seems

mudler: (and now that reminds me why wasn't using afero after all)

mudler: heck, maybe I can workaround this, but won't be nice

Itxaka: cant you just use a newMemFs and then create files in there? We do that around in here https://github.com/rancher-sandbox/elemental-cli/blob/main/pkg/utils/grub_test.go#L47

mudler: mmm, really not possible it seems, https://github.com/spf13/afero/issues/234

Itxaka: Ah so Im guessing that the yip runner uses the FD method somewhere in its code? Maybe that is what Im missing

Itxaka: blargh, does that mean that we should move everything to `vfst` ?......

This PR is small but its turning the repo upside down :rofl:

cmd/cloud-init.go
mudler: done!

pkg/types/v1/cloud-init-runner_test.go
mudler: > Ah so Im guessing that the yip runner uses the FD method somewhere in its code? Maybe that is what Im missing

nono, but yip uses vfs as provides just the same interface as os.FS, that way it's more easy to port around. The issue is that Afero returns a FS that is not matching the general interface, and I tried to do a wrapper.. but Afero doesn't return the file Fd, so can't create an `os.File` out of it

mudler: E.g. 

```golang

type AferoWrapper struct {
	afero.Fs
}

func (w *AferoWrapper) Create(name string) (*os.File, error) {
	f, _ := w.Fs.Create(name)
        // Note, Fd() doesn't exists. That would be the ideal case
	 return os.NewFile(f.Fd(),name),nil
}

```

but as afero returns afero.File, and there seems no way to get the Fd of it.. (unless I'm blind!)

Itxaka: Ahh I see, so then its the same problem I had with the ChDir which also is not on afero.

Maybe it makes sense to move to vfst after all....

Itxaka (approved):

pkg/types/v1/cloud-init-runner_test.go
mudler: > Maybe it makes sense to move to vfst after all....

Yeah.. I remember stumbling upon this, that's why I ended up moving away from it :/

PR #18, opened 2021-12-15
Add license-checker and license headers to all missing files

Itxaka:
Reused the script from hypper


mudler (dismissed):
Nais!

.github/license_check.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
```
mudler, line 2: Maybe this should have the license header too..?

Itxaka: mmmh

This checks only go files. Can we add license to a script like this which was not created from scratch and has no license?

Itxaka: done

mudler: > mmmh
> 
> This checks only go files. Can we add license to a script like this which was not created from scratch and has no license?

ehm, in theory nope, but depends if it was taken from a project which is under a license, the license otherwise cover all the files in a project.

davidcassany (approved):
LGTM

PR #19, opened 2021-12-15
Recovery

Itxaka:


Itxaka:
no tests yet.

This requires the partition stuff to be done, as it should be the partitioner the one to set the value of the `config.Recovery` at least according to the old script. Which makes sense as the partition stuff should check if the noformat is in there and set the proper partitions for the different systems or create them from scratch.

FYI @davidcassany

Itxaka:
definitely codecov is not rigth.

This is what go reports locally:

```
$ go test ./... -covermode=atomic                                                                  
ok  	github.com/rancher-sandbox/elemental-cli/cmd	0.008s	coverage: 72.6% of statements
ok  	github.com/rancher-sandbox/elemental-cli/cmd/config	0.015s	coverage: 100.0% of statements
ok  	github.com/rancher-sandbox/elemental-cli/pkg/elemental	0.036s	coverage: 83.6% of statements
ok  	github.com/rancher-sandbox/elemental-cli/pkg/partitioner	10.014s	coverage: 80.1% of statements
ok  	github.com/rancher-sandbox/elemental-cli/pkg/types/v1	0.815s	coverage: 93.0% of statements
ok  	github.com/rancher-sandbox/elemental-cli/pkg/utils	0.065s	coverage: 84.2% of statements
```

As you can see the elemental part is 83% coverage while codecov reports at 68,25% ????

Itxaka:
rebased

pkg/elemental/elemental.go
```diff
@@ -257,3 +255,166 @@ func (c Elemental) CheckNoFormat() error {
...
+	imgCosSource := fmt.Sprintf("%s/cOS/%s", c.config.StateDir, cnst.ActiveImgFile)
+	imgCosTarget := fmt.Sprintf("%s/cOS/%s", recoveryDir, cnst.RecoveryImgFile)
+
+	err = c.config.Fs.MkdirAll(recoveryDir, 0644)
+	if err != nil {
+		return err
+	}
+	var mountOptions []string
+	// TODO: c.config.Recovery is empty here!!! It should be set by the partitioner after formatting the disk, or with no format, getting the recovery partiton!!!!
+	err = c.config.Mounter.Mount(recoveryDir, c.config.Recovery, "auto", mountOptions)
```
davidcassany: Would make sense to add the device inside the Partition struct from config. So you could do something like `c.config.RecoveryPart.Device` and assume this is filled in previous stages (by partitioning or by checking al partitions are already there).

A second option would be to assume a mapping by labels, this is already the case in some other parts of cOS-toolkit. So for assume that `/dev/disk/by-label/COS_RECOVERY` is the the device and `COS_RECOVERY` is the label accessible by `c.config.RecoveryPart.Label`.

A third option would be to create some utility to get the device from a label, so one could run something like `utils.GetDevice(c.config.RecoveryPart.Label)`

@Itxaka I have no strong optionion on what would be better. Probably having some utilities for this makes sense as they can be reused in other parts of the code that are no necessarily related to partitioning/formatting. Also we can explore using stuff like @mudler suggested in #22 for this kind of utilities.

Itxaka: utility methods would be gret

Something like `GetDeviceByLabel` would probably be used aroud, and there is alreayd something similar in the `IsBootedFrom` so...

Itxaka: done!

Itxaka:
@mudler @davidcassany I think its ready for review

pkg/action/install.go
```diff
@@ -42,15 +42,18 @@ func (i InstallAction) Run() error {
...
 	disk := part.NewDisk(
 		i.Config.Target,
 		part.WithRunner(i.Config.Runner),
 		part.WithFS(i.Config.Fs),
 		part.WithLogger(i.Config.Logger),
 	)
 	// get_iso: _COS_INSTALL_ISO_URL -> download -> mount
-	// cos.GetIso() ?
-
+	err = newElemental.GetIso()
```
davidcassany, line 53: Nit: Little detail, feel free to ignore it. I'd rather include the `if c.config.Iso != ""` check outside the `GetIso` method, so we explicitly know that the ISO is only fetched when this options is set by simply reading the `install` action.

pkg/elemental/elemental.go
```diff
@@ -250,3 +248,183 @@ func (c Elemental) CheckNoFormat() error {
...
+	imgCosTarget := fmt.Sprintf("%s/cOS/%s", recoveryDir, cnst.RecoveryImgFile)
+
+	err = c.config.Fs.MkdirAll(recoveryDir, 0644)
+	if err != nil {
+		return err
+	}
+	var mountOptions []string
+	// Get CURRENT recovery device
+	// This can be an existing one (--no-format flag) or a new one done by the partitioner
+	recovery, err := c.GetDeviceByLabel(cnst.RecoveryLabel)
```
davidcassany: No, you must read it from the RunConfig! Otherwise it would not be configurable. Here a good example why make sense having constants and defaults in separate packages 😄 

All device labels are configurable

pkg/elemental/elemental.go
```diff
@@ -250,3 +248,183 @@ func (c Elemental) CheckNoFormat() error {
...
+	if err != nil {
+		return err
+	}
+	err = c.config.Fs.MkdirAll(recoveryDirCos, 0644)
+	if err != nil {
+		return err
+	}
+	if exists, _ := afero.Exists(c.config.Fs, isoMntCosSquashSource); exists {
+		c.config.Logger.Infof("Copying squashfs..")
+		sourceSquash, err := afero.ReadFile(c.config.Fs, isoMntCosSquashSource)
```
davidcassany: Isn't this reading the whole squashfs in memory? If so it is likely to kill embedded devices or small VMs.

What about something similar to (adapted from an online example):
```
// Open original file
    original, err := c.config.Fs.Open("original.txt")
    if err != nil {
        log.Fatal(err)
    }
    defer original.Close()

    // Create new file
    new, err := c.config.Fs.Create("new.txt")
    if err != nil {
        log.Fatal(err)
    }
    defer new.Close()

    //This will copy
    bytesWritten, err := io.Copy(new, original)
    if err != nil {
        log.Fatal(err)
    }
```
I'd say File entity of afero is implementing `io.Reader` and `io.Writer` if I am understanding docs correctly, I have not tested it though.

If this is correct it easily enables us to make use of the afero mocks to fake copy files. Probably makes also sense to have an utility for copying files as part of `pkg/utils/common.go`. An utility that takes source, destination and FS abstraction as parameters.

davidcassany (changes requested):
Looks good to me, just a couple of concerns:

* do not use constants for device labels, read them from RunConfig
* I think copying files could be implemented differently and probably turn it into a common utility.

pkg/elemental/elemental.go
Itxaka: good point!

pkg/elemental/elemental.go
Itxaka: umm, good point, I didnt consider that. Would need to check what it does to be sure but probably it does that.

Maybe we can direct afero.writefile to an io.Reader directly so it consumes that.

Not sure why I did it like this, I thikn its the 2nd iteretion in which I changed it...It makes no sense to open it like this...

davidcassany:
> As you can see the elemental part is 83% coverage while codecov reports at 68,25% ????

I also struggle to understand codecov report.... I got the impression that this values are related only to the new lines added within the PR. I got some 100% coverage in a PR and I think it was because all new lines of certain package where tested, despite the package had no 100% coverage.

pkg/action/install.go
Itxaka: mmh, Im torn on that

Because its a good thing to have all the steps in the install uncluttered instead of having ifs everywhere, and that means that someone calling `GetIso` cannot screw itself by calling it with an empty iso on the config.

On the other side, its more explicit to do it on the installAction, but we expose ourselves to the function being called without an iso and breaking on runtime, unless we add guards somehow in the method itself....

Maybe @mudler can shed some ligth on what is preferred in the golang community? defense coding/runtime panics?

Maybe we can even do it on both sides....to make sure there is no runtime panics.

pkg/elemental/elemental.go
Itxaka: done

pkg/types/v1/config.go
```diff
@@ -226,6 +226,10 @@ type Partition struct {
 	FS     string
 }
 
+func (r RunConfig) GetRecoveryLabel() string {
+	return r.recoveryLabel
```
davidcassany: You should trust `RecoveryPart.Label` instead. This is why I did not expose `recoveryLabel`

pkg/types/v1/config.go
```diff
@@ -158,46 +169,53 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
...
-	PassiveLabel string `yaml:"PASSIVE_LABEL,omitempty" mapstructure:"PASSIVE_LABEL"`
-	SystemLabel  string `yaml:"SYSTEM_LABEL,omitempty" mapstructure:"SYSTEM_LABEL"`
-	Force        bool   `yaml:"force,omitempty" mapstructure:"force"`
-
-	CloudInitRunner CloudInitRunner
-	PartTable       string
-	BootFlag        string
-	StateDir        string
-	GrubConf        string
+	systemLabel     string `yaml:"SYSTEM_LABEL,omitempty" mapstructure:"SYSTEM_LABEL"`
```
Itxaka, line 187: FYI @davidcassany I changed this to unexported, as far as I can see nothing is touching this (only the cmd while unmarshalling I think) so it should be ok to have it unexported.

pkg/elemental/elemental.go
```diff
@@ -364,7 +364,7 @@ func (c *Elemental) CopyRecovery() error {
 	var mountOptions []string
 	// Get CURRENT recovery device
 	// This can be an existing one (--no-format flag) or a new one done by the partitioner
-	recovery, err := c.GetDeviceByLabel(cnst.RecoveryLabel)
+	recovery, err := c.GetDeviceByLabel(c.config.GetRecoveryLabel())
```
davidcassany: No need for a getter method the label is already accessible from the Partition struct. You can use `c.config.RecoveryPart.Label`. Note that `ActiveLabel`, `PassiveLabel` or `SystemLabel` are different because they are not the label for a partition, just a filesystem image.

pkg/types/v1/config.go
davidcassany: Isn't the recovery making use of it? I do not remember exactlyt he system label use now. Anyway fine with me, the good thing of not exporting this data is that it will definitely notice it if we are missing it.

pkg/action/install.go
davidcassany: Well these are all valid points, a matter of taste I'd say. Anyway, no need to spend much time on that. As said, feel free to ignore it, all fine.

pkg/elemental/elemental.go
```diff
@@ -250,3 +249,187 @@ func (c Elemental) CheckNoFormat() error {
...
+}
+
+// CopyRecovery will
+// Check if we are booting from squash -> false? return
+// true? -> :
+// mkdir -p RECOVERYDIR
+// mount RECOVERY into RECOVERYDIR
+// mkdir -p  RECOVERYDIR/cOS
+// if squash -> cp -a RECOVERYSQUASHFS to RECOVERYDIR/cOS/recovery.squashfs
+// if not -> cp -a STATEDIR/cOS/active.img to RECOVERYDIR/cOS/recovery.img
```
davidcassany, line 342: I find this a weird and unexpected behavior, but I guess this was already the case in `cos.sh`right?
Does it mean that the recovery image can only be different of the active image in case we are using an squash based recovery image?

Looks like a logic coupled with the installer ISO setup. I am wondering if it wouldn't make more sense to make an installer method that dumps the required images into a disk and that's it. I mean, what about having this logic of choosing the images and digesting the setup elsewhere and simply relay on `RunConfig` specific parameters to dump Active, Passive and Recovery to a disk.

So, `CopyRecovery` would simply copy a file into a certain path of `COS_RECOVERY` device. All the runtime logic to choose the images depending on flags and on host status could be done on early stage and store relevant results in `RunConfig`.

By doing that it will be damn easy later in time to adapt the installer to some config file that could specify custom images (even remote from some registry) of Active Passive and Recovery. That's why I believe having state data stored in a common structure like RunConfig is a win.

So in RunConfig we already have the `Partition` struct, so why not having the `Image` struct for Recovery, Passive and Active images? For each we could have defined (filesystem, label, size, source...), so having that in `RunConfig` could allow as to code a an internal method that essentially creates the image, formats it and then copies source to it given a `Image` instance regardless of being Active, Passive or Recovery.

@Itxaka @mudler what do you think? Sorry to keep making suggestions or changes, but I believe this is the right time to do it.

Itxaka: Yes it's very weird. I can't recall now, would need to recheck the cos.sh script, but I think this is because this function is supposed to be called from the recovery as well to recover, so it has different paths to reuse code.

pkg/types/v1/config.go
Itxaka: blergh, you are rigth

pkg/elemental/elemental.go
Itxaka: yes, so this seems to be the path to take depending on if we have a `recovery.squashfs` or not in the current system
https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/installer/cos.sh#L318

Otherwise, if we dont have the recovery we just copy the active partition into the recovery. Im not sure why...

@mudler ?? ^^^ any idea?

Itxaka: The problem I think is that in the old scripts there is several places where paths can be changed by different functions and that was taken advantage of to modify paths and reuse code, i.e. ISOMNT is used for a live system AND if we pass a --iso to install from. so that kind fo tangles things around in a more structured code, as we dont have the possibility of just overwriting stuff from the root namespace and call it a day :/

I would say to move forward with this until we have something that can install the system like the old script and then restructure it once we have the full picture of the whole workflow, i.e. what @davidcassany says about a generic copyX that dumps the image sounds about right, but we dont know what the passive copy methods is gonna look yet as its not implemented, so that could also change global vars and so on.

IMO, a rough installer that can at least install and cover the ground that the old installer did would be the first step before reorganizing everything and refining it.

davidcassany: > I would say to move forward with this until we have something that can install the system like the old script and then restructure it once we have the full picture of the whole workflow, i.e. what @davidcassany says about a generic copyX that dumps the image sounds about right, but we dont know what the passive copy methods is gonna look yet as its not implemented, so that could also change global vars and so on.
> 
> IMO, a rough installer that can at least install and cover the ground that the old installer did would be the first step before reorganizing everything and refining it.

Fine with me if we come back to it as soon as we got the rough installer up and ready. Then, I will move back to the TODO column the copy passive partition (I just had started to read the code and try organize what was needed in my head) and focus on the missing parts between creating all the partitions and copying active partitions (`CopyCos`). Note we are missing the filesystem image creation step.

davidcassany (approved):

PR #20, opened 2021-12-15
Add partition and formatting methods to elemental package

davidcassany:

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -43,11 +43,19 @@ func NewElemental(config *v1.RunConfig) *Elemental {
 func (c *Elemental) PartitionAndFormatDevice(disk *part.Disk) error {
 	c.config.Logger.Infof("Partitioning device...")
 
-	stateFlags := []string{}
-	var efiNum, oemNum, stateNum, recoveryNum, persistentNum int
+	err := c.createPTableAndFirmwarePartitions(disk)
+	if err != nil {
+		return err
+	}
+
+	return c.createDataPartitions(disk)
+}
+
+func (c *Elemental) createPTableAndFirmwarePartitions(disk *part.Disk) error {
 	errCMsg := "Failed creating %s partition"
 	errFMsg := "Failed formatting partition: %s"
 
+	c.config.Logger.Debugf("Creating partition table...")
 	out, err := disk.NewPartitionTable(c.config.PartTable)
 	if err != nil {
 		c.config.Logger.Errorf("Failed creating new partition table: %s", out)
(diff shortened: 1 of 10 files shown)


davidcassany:
Gonna improve config tests, but the current code is already ready to review.

pkg/elemental/elemental.go
```diff
@@ -19,12 +19,17 @@ package elemental
...
+	part "github.com/rancher-sandbox/elemental-cli/pkg/partitioner"
 	v1 "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
 	"github.com/rancher-sandbox/elemental-cli/pkg/utils"
 	"github.com/zloylos/grsync"
 	"net/http"
 	"os"
 	"strings"
+
+	"github.com/mudler/yip/pkg/console"
+	"github.com/twpayne/go-vfs"
```
davidcassany, line 32: I do have concerns about this, I think it would be better to create a wrapper structure for yip. Yip dependencies will spread all over the elemental code with the current approach. I don't think this is a problem for `cmd/cloud-init.go` since the dependency there is clear, having to include yip's `console` and `vfs` everywhere we wanna call yip doesn't look right to me.

pkg/types/v1/config.go
```diff
@@ -101,32 +102,97 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
...
 }
 
 // RunConfig is the struct that represents the full configuration needed for install, upgrade, reset, rebrand.
 // Basically everything needed to know for all operations in a running system, not related to builds
 type RunConfig struct {
+	//Interanlly used to compute RunConfig state
+	recoveryLabel   string `yaml:"RECOVERY_LABEL,omitempty" mapstructure:"RECOVERY_LABEL"`
+	persistentLabel string `yaml:"PERSISTENT_LABEL,omitempty" mapstructure:"PERSISTENT_LABEL"`
+	stateLabel      string `yaml:"STATE_LABEL,omitempty" mapstructure:"STATE_LABEL"`
+	oEMLabel        string `yaml:"OEM_LABEL,omitempty" mapstructure:"OEM_LABEL"`
```
davidcassany, line 159: These are not exposed since they should not be consumed at any later point in time beyond initialization here. They are used to fulfill partition structure.

I have been thinking it would be nice to make config little smarter for early validation to enforce consistency as early as possible and also to avoid having to validate dynamic values at the time we need them (eg. `partition-layout` file exists). The idea is that would be good we can RunConfig structure at any point during the execution without having to check or double check the data it contains. So methods like `SetupStyle` could be executed within the `NewRunConfig` or wherever we initiate RunConfig.

davidcassany (commented):
While coding this part I had a couple ideas I'd like to address before the code grows too much.

* Completely wrap `yip` so there is no need to import yip dependencies everywhere we need it
* Make config structure smarter, so we run early checks on command line flags and config parameters (aka, check target path exists and basic checks like that).

@mudler @Itxaka do you have any thought about that?

pkg/elemental/elemental.go
mudler: Just a quick thought, maybe we can expand later the `CloudRunner` a bit to return a wrapper.. not sure

mudler (approved):

mudler:
> While coding this part I had a couple ideas I'd like to address before the code grows too much.
> 
>     * Completely wrap `yip` so there is no need to import yip dependencies everywhere we need it

Makes sense, on the other hand I fear we start to abstract too much and make complex constructors, but on the `yip` case are well defined, so I think that's a good valid point.
> 
>     * Make config structure smarter, so we run early checks on command line flags and config parameters (aka, check target path exists and basic checks like that).

What about a Validate() we run _somewhere_ (e.g. constructor, or..) ?

davidcassany:
> > While coding this part I had a couple ideas I'd like to address before the code grows too much.
> > ```
> > * Completely wrap `yip` so there is no need to import yip dependencies everywhere we need it
> > ```
> 
> Makes sense, on the other hand I fear we start to abstract too much and make complex constructors, but on the `yip` case are well defined, so I think that's a good valid point.

Yes, also I found that this is good for testing purposes. Otherwise it turns to be painful to test yip execution, as it uses a different console and filesystem wrapper so our current mocks are useless there. Also yip inners should be transparent to us. I found that a great way of passing stuff is through RunConfig, I add a CloudInitRunner there so I have access to it almost everywhere and I can set it to a `yip` runner by default but inject a mock on tests. So we are not really passing it as a parameter everywhere only setting it by default at RunConfig initialization and swapping it to something else in tests.

> 
> > ```
> > * Make config structure smarter, so we run early checks on command line flags and config parameters (aka, check target path exists and basic checks like that).
> > ```
> 
> What about a Validate() we run _somewhere_ (e.g. constructor, or..) ?

Yes, I was thinking about something like that. Problem is I don't fully understand `viper` and `config`, yet 😅  I have not dived much into that part yet, how stuff gets applied to RunConfig from command line or a config file is still foggy to me.

davidcassany:
Merging, follow up PRs will come to improve config coverage and slightly refactor CloudInitRunner through an interface. I want to get that merged to unblock other install action parts and keep covering install action features set.

PR #21, opened 2021-12-16
Make a CloudInitRunner interface

davidcassany:

### DIFF
--- a/pkg/types/v1/cloud-init-runner.go
+++ b/pkg/types/v1/cloud-init-runner.go
@@ -76,7 +76,7 @@ func (ci YipCloudInitRunner) Run(stage string, args ...string) error {
 	return ci.exec.Run(stage, ci.fs, ci.console, args...)
 }
 
-func (ci YipCloudInitRunner) SetModifier(m schema.Modifier) {
+func (ci *YipCloudInitRunner) SetModifier(m schema.Modifier) {
 	ci.exec.Modifier(m)
 }
 
--- a/pkg/types/v1/cloud-init-runner.go
+++ b/pkg/types/v1/cloud-init-runner.go
@@ -17,14 +17,27 @@ limitations under the License.
 package v1
 
 import (
+	"github.com/mudler/yip/pkg/console"
 	"github.com/mudler/yip/pkg/executor"
 	"github.com/mudler/yip/pkg/plugins"
+	"github.com/mudler/yip/pkg/schema"
+	"github.com/twpayne/go-vfs"
 )
 
-// CloudInitRunner returns a default cloud init executor with the Elemental plugin set.
+type CloudInitRunner interface {
+	Run(string, ...string) error
+}
+
+type YipCloudInitRunner struct {
+	exec    executor.Executor
+	fs      vfs.FS
+	console plugins.Console
+}
+
+// CloudInitRunner returns a default yip cloud init executor with the Elemental plugin set.
 // It accepts a logger which is used inside the runner.
-func CloudInitRunner(l Logger) executor.Executor {
-	return executor.NewExecutor(
+func NewYipCloudInitRunner(l Logger) YipCloudInitRunner {
+	exec := executor.NewExecutor(
 		executor.WithConditionals(
 			plugins.NodeConditional,
 			plugins.IfConditional,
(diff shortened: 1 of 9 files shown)


pkg/types/v1/config.go
```diff
@@ -87,6 +94,10 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
 		r.Mounter = mount.New(constants.MountBinary)
 	}
 
+	if r.CIRunner == nil {
+		r.CIRunner = NewYipCloudInitRunner(r.Logger)
```
mudler: You can move this up avoiding the `if` branch entirely. they are actually defaults that can be override otherwise

davidcassany: Oh yes, you are right

pkg/types/v1/config.go
```diff
@@ -69,12 +68,21 @@ func WithRunner(runner Runner) func(r *RunConfig) error {
 	}
 }
 
+func WithCIRunner(ci CloudInitRunner) func(r *RunConfig) error {
```
Itxaka: Nit: Call it `WithCloudInitRunner` as my first idea when seeing this was that this was for CI, not cloud Init :)

pkg/types/v1/config.go
```diff
@@ -172,6 +184,7 @@ type RunConfig struct {
 	SystemLabel  string `yaml:"SYSTEM_LABEL,omitempty" mapstructure:"SYSTEM_LABEL"`
 	Force        bool   `yaml:"force,omitempty" mapstructure:"force"`
 
+	CIRunner       CloudInitRunner
```
Itxaka: do we need this exported?

pkg/types/v1/cloud-init-runner.go
```diff
@@ -53,4 +66,21 @@ func CloudInitRunner(l Logger) executor.Executor {
 			plugins.Layout,
 		),
 	)
+	return &YipCloudInitRunner{
+		exec: exec, fs: vfs.OSFS,
+		console: console.NewStandardConsole(console.WithLogger(l)),
```
mudler, line 71: just as a side node as it was already brought up as a point, `yip` here takes an interface: https://github.com/mudler/yip/blob/1d415391cc37e353facdbc7e5beb22024ca818ba/pkg/plugins/common.go#L24. 

So we can let our console return a specialized wrapper, or either create a func that returns a wrapper that satisfies the interface, in that way we would get rid of the standard console from `yip`.

pkg/types/v1/config.go
mudler: same here ;)

davidcassany: Probably I'll call everything CloudInitRunner without initials anywhere, at least it will be explicit and I don't think we will have to type this variables that often. I tend to go mad trying to find short names, usually with little success 😅

pkg/types/v1/config.go
davidcassany: Yes, I am using it in the elemental.go, see https://github.com/rancher-sandbox/elemental-cli/blob/29e458d89c907660a07e43057ede51b89f14df8b/pkg/elemental/elemental.go#L55

We could use a getter instead of accessing the variable directly, but I don't see much gain these are additional lines of code. 
Another alternative could be to add the CloudInitRunner as a parameter of the constructor of elemental, but I prefer having it stored in RunConfig so we can reuse the same instance (with the same configs) in all places. Also this prevents having to keep passing interfaces to all methods in code, we store them in RunConfig and this struct should be available in any elemental specific code (aka, elemental package, actions, etc.)

pkg/types/v1/cloud-init-runner.go
davidcassany: So you mean creating method in our console that returns an implementation of the yip's interface? I am not sure I understand it fully. This would let us get rid of the `github.com/mudler/yip/pkg/console` import and provide a mechanism to inject our own console in yip. So basically the gain is having the availability to run yip using a console we fully control, hence we could eventually mock/fake it in test.

Is there any other benefit or detail I am missing? For tests this could be helpful indeed, even I doubt we will reach the level getting into verify internal yip's calls, this should be a duty of yip's tests, not necessarily elemental tests.

mudler: was just a note only if we want to have full control on the console and how it is executing commands, and if we want to avoid more deps... I don't think we really need it as `yip` by itself have its own test suites, but as you suggest, during mocking might be helpful later on.. anyhow for that to happen or our console have to satisfy the one from `yip`, or either have a wrapper that 'emulates' and wraps what `yip` expects there

davidcassany: I am not sure how we can make our console to implement both interfaces, both are declaring the `Run` method and as long as I am aware golang does not support overloading.

Is is possible to code a single implementation for both?

```
type Runner interface {                                                         
        Run(string, ...string) ([]byte, error)                                      
}

type Console interface {
	Run(string, ...func(*exec.Cmd)) (string, error)
}
```

I don't think this is possible to code an entity implementing both. If possible I'd do it.

Anyway, gonna merge this PR and if you have an idea of how to solve it without having to create a new specific console type we can code it later. I'd like to avoid having to create a type like `YipConsole` or `CloudInitConsole`, for that I'd just reuse the one form yip and mock at yip wrapper level for tests.

mudler (approved):

PR #27, opened 2021-12-21
Add mounting routines for the installer action

### LINKED ISSUE #1005: Partitions mount and unmount

davidcassany:
Fixes rancher-sandbox/cOS-toolkit#1005

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -48,7 +48,7 @@ var installCmd = &cobra.Command{
 		}
 		// Should probably load whatever env vars we want to overload here and merge them into the viper configs
 		// Note that vars with ELEMENTAL in front and that match entries in the config (only one level deep) are overwritten automatically
-		cfg.Device = args[0]
+		cfg.Target = args[0]
 
 		cfg.Logger.Infof("Install called")
 		// Dont call it yet, not ready
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -74,26 +74,54 @@ func (i InstallAction) Run() error {
 	if err != nil {
 		return err
 	}
+
+	err = newElemental.MountPartitions()
+	if err != nil {
+	}
+	defer func() {
+		if tmpErr := newElemental.UnmountPartitions(); tmpErr != nil && err == nil {
+			err = tmpErr
+		}
+	}()
+
+	// create active file system image
+	err = newElemental.CreateFileSystemImage(i.Config.ActiveImage)
+	if err != nil {
+		return err
+	}
+
+	//mount file system image
+	loop, err := newElemental.MountImage(i.Config.ActiveImage, i.Config.Target)
+	if err != nil {
+		return err
+	}
+
 	// install Active
 	err = newElemental.CopyCos()
 	if err != nil {
+		newElemental.UnmountImage(i.Config.Target, loop)
 		return err
 	}
 	// Copy cloud-init if any
 	err = newElemental.CopyCloudConfig()
 	if err != nil {
+		newElemental.UnmountImage(i.Config.Target, loop)
 		return err
 	}
 	// install grub
 	grub := utils.NewGrub(i.Config)
 	err = grub.Install()
 	if err != nil {
+		newElemental.UnmountImage(i.Config.Target, loop)
 		return err
 	}
 	// Relabel SELinux
 	_ = newElemental.SelinuxRelabel(false)
 	// Unmount everything
-	// cos.CleanupMounts()
+	err = newElemental.UnmountImage(i.Config.Target, loop)
+	if err != nil {
+		return err
+	}
 	// install Recovery
 	// cos.CopyRecovery()
 	err = newElemental.CopyRecovery()
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -19,6 +19,7 @@ package action
 import (
 	"errors"
 	"fmt"
+	cnst "github.com/rancher-sandbox/elemental-cli/pkg/constants"
 	"github.com/rancher-sandbox/elemental-cli/pkg/elemental"
 	part "github.com/rancher-sandbox/elemental-cli/pkg/partitioner"
 	"github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
@@ -77,6 +78,7 @@ func (i InstallAction) Run() error {
 
 	err = newElemental.MountPartitions()
 	if err != nil {
+		return err
 	}
 	defer func() {
 		if tmpErr := newElemental.UnmountPartitions(); tmpErr != nil && err == nil {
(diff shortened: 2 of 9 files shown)


davidcassany:
Draft as tests are missing, but I'd appreciate reviews on the code

pkg/constants/constants.go
```diff
@@ -47,7 +46,9 @@ const (
 	BiosSize           = uint(1)
 	PartStage          = "partitioning"
 	RecoveryDirSquash  = "/run/initramfs/live"
-	RecoveryDir        = "/run/initramfs/cos-state"
+	RecoveryDir        = "/run/initramfs/cos-recovery"
+	StateDir           = "/run/initramfs/cos-state"
+	OEMDir             = "/run/initramfs/cos-oem"
```
davidcassany: I would change those paths, as it has nothing to do with initramfs... I bet this is just legacy form `cos.sh` where probably the methods mounting were having environment variables instead of parameters to define target.

These values are the where partitions get mounted at install time. So I'd say something like `/run/cos/recovery` is more suited and less error prone.

pkg/elemental/elemental.go
```diff
@@ -351,28 +457,12 @@ func (c *Elemental) CopyRecovery() error {
 	if !c.BootedFromSquash() {
 		return nil
 	}
-	recoveryDir := c.GetRecoveryDir()
```
davidcassany, line 354: Note GetRecoveryDir used to depend on `c.BootedFromSquash`, which in turn is just validated on the previous command, so in this method we can safely relay on `cnst.RecoveryDir` and assume the recovery partition is already mounted at that point.

pkg/action/install.go
```diff
@@ -74,26 +74,54 @@ func (i InstallAction) Run() error {
 	if err != nil {
 		return err
 	}
+
+	err = newElemental.MountPartitions()
+	if err != nil {
+	}
+	defer func() {
```
Itxaka, line 83: wow, I dont understand this. does defer run once the current function is closing? Will an error inside the defer return an error from this function or is that lost?

pkg/action/install.go
```diff
@@ -74,26 +74,54 @@ func (i InstallAction) Run() error {
...
+		if tmpErr := newElemental.UnmountPartitions(); tmpErr != nil && err == nil {
+			err = tmpErr
+		}
+	}()
+
+	// create active file system image
+	err = newElemental.CreateFileSystemImage(i.Config.ActiveImage)
```
Itxaka, line 90: love it

pkg/elemental/elemental.go
```diff
@@ -145,6 +145,122 @@ func (c *Elemental) createDataPartitions(disk *part.Disk) error {
...
+	if err != nil {
+		c.config.Mounter.Unmount(cnst.StateDir)
+		c.config.Mounter.Unmount(cnst.RecoveryDir)
+		return err
+	}
+	return nil
+}
+
+// UnmountPartitions unmounts recovery, state and oem partitions.
+func (c Elemental) UnmountPartitions() error {
```
Itxaka, line 170: super-mega-NIT: should we call this method `Close` ? like seems in the golang world that the cleanup is always done on the close method, so people keep calling `defer whatever.close` ?

pkg/elemental/elemental.go
```diff
@@ -145,6 +145,122 @@ func (c *Elemental) createDataPartitions(disk *part.Disk) error {
...
+	var err error
+	errMsg := "Failed to unmount %s"
+	failure := false
+
+	// If there is an early error we still try to unmount other partitions
+	err = c.config.Mounter.Unmount(cnst.RecoveryDir)
```
Itxaka: cant we iterate over this instead? 

```
for _, mnt := range []string{cnst.RecoveryDir, cnst.StateDir, cnst.OEMDir} {
  err = c.config.Mounter.Unmount(mnt)
  if err != nil {
    c.config.Logger.Errorf(errMsg, mnt)
    failure = true
  }
}
```

pkg/elemental/elemental.go
```diff
@@ -145,6 +145,122 @@ func (c *Elemental) createDataPartitions(disk *part.Disk) error {
...
+		c.config.Logger.Errorf(errMsg, cnst.StateDir)
+		failure = true
+	}
+	err = c.config.Mounter.Unmount(cnst.OEMDir)
+	if err != nil {
+		c.config.Logger.Errorf(errMsg, cnst.OEMDir)
+		failure = true
+	}
+	if failure {
+		return errors.New("Failed unmounting some partition, see error log")
```
Itxaka: cant we use the errorMsg+ mnt here instead by setting the failure to a message? `failure= fmt.Sprintf(errorMsg, mnt)` and then return `errors.New(failure)`

pkg/elemental/elemental.go
```diff
@@ -145,6 +145,122 @@ func (c *Elemental) createDataPartitions(disk *part.Disk) error {
...
+	if err != nil {
+		return err
+	}
+	_, err = c.config.Runner.Run("losetup", "-d", device)
+	return err
+}
+
+// CreateFileSystemImage creates the image file for config.target
+func (c Elemental) CreateFileSystemImage(img v1.Image) error {
+	if exists, _ := afero.Exists(c.config.Fs, img.File); exists {
```
Itxaka: I think we can skip that, according to docs Create will truncate the file if it exists.

```
// Create creates or truncates the named file. If the file already exists,
// it is truncated.
```

Itxaka (commented):
ooking good to me!

pkg/elemental/elemental.go
davidcassany: It looks like indeed.

pkg/action/install.go
davidcassany: This is to not silent the errors that may appear on `umount` and also not to overlap any umount error with previous errors.
So something as simple as 

```golang
defer newElemental.UnmountPartitions()
```
silents any unmounting error and I want to error out if unmounting fails. So, instead, we just call a function that checks the return value of unmounting and if it failed and there wasn't any previous error stored in `err` it replaces `err` with the unmounting error. This assumes the function will return `err`.

Itxaka: nice, will stole this

pkg/elemental/elemental.go
davidcassany: I'd prefer to keep the current naming and if later on we see fit for a generic `Close` (umount partitions, if any, and any other required logic) method in elemental we could consider it and defer it almost after the creation of the `Elemental` instance.

davidcassany:
This is finally ready for review. I rebased and added tests for the new methods included in this PR.

cmd/install.go
```diff
@@ -48,7 +48,7 @@ var installCmd = &cobra.Command{
 		}
 		// Should probably load whatever env vars we want to overload here and merge them into the viper configs
 		// Note that vars with ELEMENTAL in front and that match entries in the config (only one level deep) are overwritten automatically
-		cfg.Device = args[0]
+		cfg.Target = args[0]
```
Itxaka, line 51: is Device even used anymore?

Itxaka: ah I see its removed down in the config, nice!

Itxaka (approved):
Looking good!

PR #29, opened 2021-12-22
Adding copy passive image to installer

### LINKED ISSUE #962: Passive partition setup

davidcassany:
Fixes rancher-sandbox/cOS-toolkit#962

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -17,7 +17,11 @@ limitations under the License.
 package utils
 
 import (
+	"errors"
+	"fmt"
 	"github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/spf13/afero"
+	"io"
 	"os/exec"
 	"strings"
 )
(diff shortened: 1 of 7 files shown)


pkg/utils/common.go
```diff
@@ -41,3 +45,41 @@ func FindLabel(runner v1.Runner, label string) (string, error) {
...
+		if err == nil {
+			err = sourceFile.Close()
+		}
+	}()
+
+	targetFile, err := fs.Create(target)
+	if err != nil {
+		return err
+	}
+	defer func() {
```
mudler, line 77: I'm not sure if this is really required at the end. If the copy fails it would error out anyway

davidcassany: This is to ensure we capture the error on `targetFile.Close()`, this is important as this runs the `sync` operation. So if `target.Close()` fails we most likely lost data and have a corrupted file.
I assume we do not need to close the file if it already failed earlier, since I assume we will error out, but yes this is arguable. We could always close and ignore any `Close()` error if there was already a previous error. I did not code it this way to keep logic simple.

mudler (approved):
looks good!

PR #31, opened 2021-12-24
Installer code polishing

davidcassany:
This PR is related to rancher-sandbox/cOS-toolkit#1020


davidcassany:
With this PR I have been able to run a successful installation from an ISO to a disk in non EFI firmware.

mudler (approved):
Looks good here

PR #32, opened 2021-12-28
Fix and add missing installer features

### LINKED ISSUE #1020: Test elemental client installation
Within #905 a relatively big chunk of code has been coded without having the chance to run real tests on development environments. Having rancher-sandbox/elemental-cli#30 merged now should be possible to experiment with basic installations using elemental client.

This card is to encompass all the code polishing that these initial tests will raise.

davidcassany:
This PR includes several fixes and features:

* Fixes and polishes the installation for GPT and EFI use cases
* Adds installation hooks

This PR fixes rancher-sandbox/cOS-toolkit#1020 and it relates to rancher-sandbox/cOS-toolkit#1019

Note: it can be easily reviewed on per commit basis, all commits are self contained and passing all tests.


pkg/elemental/elemental.go
```diff
@@ -144,38 +150,54 @@ func (c *Elemental) createDataPartitions(disk *part.Disk) error {
...
+// MountPartitions mounts state, recovery, oem, persistent and efi partitions.
+// Note umounts must be handled by caller logic.
 func (c Elemental) MountPartitions() error {
-	err := c.mountDeviceByLabel(c.config.StatePart.Label, cnst.StateDir, "rw")
-	if err != nil {
-		return err
+	c.config.Logger.Infof("Mounting disk partitions")
+	var err error
+	parts := []v1.Partition{
+		c.config.StatePart,
```
mudler, line 159: a bit unrelated to the PR, more a note for the future, but this starts to make me think we might need a specific section for partitions in our config?

pkg/elemental/elemental.go
```diff
@@ -221,6 +253,13 @@ func (c Elemental) MountImage(img *v1.Image) error {
 }
 
 func (c Elemental) UnmountImage(img *v1.Image) error {
+	// Using IsLikelyNotMountPoint seams to be safe as we are not checking
+	// for bind mounts here
+	if notMnt, _ := c.config.Mounter.IsLikelyNotMountPoint(img.MountPoint); notMnt == true {
```
mudler, line 258: that's more a style note, I'm just thinking that maybe the mounter should be passed by instead of consuming it from the config. Maybe we can pass by a set of general options directly to the methods

tests/mocks/mounter_mock.go
```diff
@@ -64,6 +64,17 @@ func (e ErrorMounter) List() ([]mount.MountPoint, error) {
 	return e.FakeMounter.List()
 }
 
+func (e ErrorMounter) IsLikelyNotMountPoint(file string) (bool, error) {
+	e.prepare()
+	mnts, _ := e.List()
+	for _, mnt := range mnts {
+		if file == mnt.Path {
```
mudler, line 71: I wonder..shall this check be more flexible, e.g. taking into account that file can include parts of mnt.Path? For example `/tmp/` is mounted, and if I pass `/tmp/foo.txt` to the function, shall it return true?

I say this because the function signature suggests that the is capable to tell if the file is inside a mount path or not, but then the function checks if it's a mounted path, so I'm not sure what's the original wanted behavior here

mudler (approved):
Just minor nitpicks, it looks good! also great that we move tests to the ginkgo format :100:

pkg/elemental/elemental.go
davidcassany: Yes, I fully agree. My next little refactor idea in that regard was having a map of `v1.Partition` instead of specific instances, so in config it could be expressed as `Partitions map[string]v1.Partition` then I could simply iterate over keys.

pkg/elemental/elemental.go
davidcassany: I have no strong opinion here, but I agree that `RunConfig` is probably including too much. One not matured idea I had is that probably we should split user options from command setup. So digest user options (flags, config, files, env variables) and produce a command specific config. So imagine in install action we just pass to `Elemental` (or to Elemental methods) a set of generic options (logger, mounter, etc.) and a precomputed `InstallConfig` struct including all the details for this specific installation.

Just as an example of what concerns me, `RunConfig` struct includes `ForceGpt` and `PartTable`, from an API perspective I just need `PartTable`. I don't want to process `--force-gpt` flag in `partitioner`, obviously I can ignore `ForceGpt` it because I know the code, but if this is ever used somewhere else as an API by looking at the struct definition something like `PartTable = "msdos"` and `ForceGpt = true` is a possible situation.

tests/mocks/mounter_mock.go
davidcassany: I was also surprised by this naming convention. This method signature is copied from the k8s implementation. To me it feels like `/tmp/foo.txt` is not the expected use for this function. However I have to admit I have not tested it with the real implementation.

We just use it for directories, hence this mock is consistent with the use we are making of it. In any case this is just a mock suited to our needs, the  `FakeMounter` was not even implementing any logic for this method, so I assume trying to code a flexible and generic implementation here could be a never ending story.

PR #34, opened 2021-12-31
Add support for `--docker-image`

davidcassany:


pkg/types/v1/luet.go
```diff
@@ -0,0 +1,70 @@
...
+type Luet struct {
+	log Logger
+}
+
+func NewLuet(log Logger, plugins ...string) *Luet {
+	c := &types.LuetConfig{}
+	c.Init()
+	c.Solver.Options = solver.Options{Type: solver.SingleCoreSimple, Concurrency: c.General.Concurrency}
+	// TODO Noted this is a dangerous context setup `util.DefaultContext.Clean()` deletes the current directory!!
+	util.DefaultContext = context.NewContext(
```
davidcassany: @mudler I marked this as draft mostly because I bet this luet initialization should be done differently. I noted some strange behavior with this default context initialization. I guess this all works because the `unpack` command is pretty much disconnected from all these defaults settings.

Do you have any hint about how this should be done?

mudler: It looks good, although calling c.Init isn't really required unless you want all the setup to happen likely in the CLI. 
For unpack indeed which is mostly disconnected most of the default would just work, so I guess this is for preparing later code for it, right?

Keep in mind that case what is "mandatory" is the system configuration:

```
ctx := luetContext.NewContext(
		luetContext.WithConfig(
			&types.LuetConfig{
				System: types.LuetSystemConfig{
					Rootfs:        "/somewhere/",
					TmpDirBase:    "/some/tmp",
					PkgsCachePath: "/some/cache"
					DatabasePath:  "/some/db",
				},
			},
		),
	)
```

davidcassany: Well I started adding `luet install` to the interface too, but realized later that this was not used during the `elemental install` and I also realized I had a non really functional context in tests. Ok so I will try what you suggested. Do you have any special idea about what those system values should be? Rootfs is the target folder for install command? Would it make sense to set it to the installation target?

mudler: The `system` block refers to the system which is running the install/upgrade/uninstall commands.

I'd say to let the rootfs maybe open when creating a new instance, as might be re-used in different contexts, for example while building raw images - for the rest we could stick to some internal defaults, the Tmpdir base is the directory used to store temporary files, while PkgsCache to hold the package cache and Database to hold the system database information. Note all of that must be provided as absolute path as they not need to be residing necessarly inside the installation rootfs

davidcassany: For now I left it with all defaults as this is sufficient for the unpack command. We can improve/extend configuration customization when further luet operations are needed in elemental.

Itxaka (approved):
Nice! This looks way too easy... :D

davidcassany:
@Itxaka  I rebased and added a new commit to configure cosign and mtree luet plugins

Itxaka (approved):

PR #35, opened 2021-12-31
Refactor partitions list

davidcassany:


davidcassany:
This refactor is based on a review https://github.com/rancher-sandbox/elemental-cli/pull/32#discussion_r776958537

Code coverage is reduced because many code lines are dropped in this refactor, hence the covered vs non covered ratio decreases, despite the tests are testing exactly the same logic as they used to.

@mudler finally I did not use a map because I needed the partitions be in a list of a known order, this way it is possible to loop them  from first to last when partitioning the disk.

mudler (approved):
Looks good here! nice :100: !

PR #36, opened 2022-01-10
Add run-stage command

Itxaka:
Substitutes the use of the cos-setup script

### DIFF
--- a/cmd/run-stage.go
+++ b/cmd/run-stage.go
@@ -1,0 +1,147 @@
+/*
+Copyright © 2021 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"github.com/mudler/yip/pkg/schema"
+	"github.com/rancher-sandbox/elemental-cli/cmd/config"
+	"github.com/spf13/afero"
+	"github.com/spf13/viper"
+	"k8s.io/mount-utils"
+	"os"
+	"os/exec"
+	"strings"
+
+	v1 "github.com/rancher-sandbox/elemental-cli/pkg/types/v1"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// cloudInit represents the cloud-init command
+var runStage = &cobra.Command{
+	Use:   "run-stage STAGE",
+	Short: "elemental run-stage",
+	Args:  cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var cmdLineYipUri string
+		var FinalCloudInitPaths []string
+		CloudInitPaths := []string{"/system/oem", "/oem/", "/usr/local/cloud-config/"}
+
+		logger := logrus.New()
+		logger.SetOutput(os.Stdout)
+
+		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), logger, &mount.FakeMounter{})
+
+		if err != nil {
+			cfg.Logger.Errorf("Error reading config: %s\n", err)
+		}
+
+		// Check if we have extra cloud init
+		if cfg.CloudInitPaths != "" {
+			cfg.Logger.Debugf("Adding extra paths: %s", cfg.CloudInitPaths)
+			extraCloudInitPathsSplit := strings.Split(cfg.CloudInitPaths, " ")
+			CloudInitPaths = append(CloudInitPaths, extraCloudInitPathsSplit...)
+		}
+
+		// Cleanup paths. Check if they exist and add them to the final list to avoid failures on non-existing paths
+		for _, path := range CloudInitPaths {
+			exists, err := afero.Exists(afero.NewOsFs(), path)
+			if exists && err == nil {
+				FinalCloudInitPaths = append(FinalCloudInitPaths, path)
+			} else {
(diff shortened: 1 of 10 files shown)


cmd/run-stage.go
```diff
@@ -0,0 +1,147 @@
...
+
+// cloudInit represents the cloud-init command
+var runStage = &cobra.Command{
+	Use:   "run-stage STAGE",
+	Short: "elemental run-stage",
+	Args:  cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var cmdLineYipUri string
+		var FinalCloudInitPaths []string
+		CloudInitPaths := []string{"/system/oem", "/oem/", "/usr/local/cloud-config/"}
```
davidcassany: nit: isn't this slice a candidate for a constant? It could be an static method in constants returning this slice.

davidcassany (approved):
Looks good to me! :+1:

davidcassany:
A feature that I miss from `cos-setup` script is the capability of overwriting already existing stages. I mean, instead of running the stage found in each directory, just run the first match. I am thinking that there might be cases in which we have some stage predefined (`initramfs` stage in `/system/oem`) that we simply want to swap by a custom stage (`initramfs` stage in `/oem`). 

With the current setup both are executed, first the one from `/system/oem` and later the one from `/oem`. I wish there could be a way to enable to only run the custom one. This could be done by looping the paths in reverse order and skip the rest as soon as there is a match.

@Itxaka @mudler do you think it makes sense? This could be useful for low level branding at install time (e.g. partition labels and similar stuff).

mudler:
> A feature that I miss from `cos-setup` script is the capability of overwriting already existing stages. I mean, instead of running the stage found in each directory, just run the first match. 

That would break extendability, as you can have more cloud-inits coming from multiple sources attempting to run in a defined stage.

> I am thinking that there might be cases in which we have some stage predefined (`initramfs` stage in `/system/oem`) that we simply want to swap by a custom stage (`initramfs` stage in `/oem`).
> 

I'm not sure here if the intention is to stop users to mess with system configuration from cloud init, but in that case it is enough to advertise pure cloud init support. By default users using standard cloud-inits syntax would just be mapped as boot.  If they are using our internal stages setup, then it means there is a reason why ( e.g they need to hook into initramfs to embed some module, etc )

> With the current setup both are executed, first the one from `/system/oem` and later the one from `/oem`. I wish there could be a way to enable to only run the custom one. This could be done by looping the paths in reverse order and skip the rest as soon as there is a match.

It is not only about the path, but we have actually more cloud init files hooking at the same stage. They could be even in the same path.

For low level branding during install, we do already have several hooks that let you modify the behavior of the system pre/after install, and also by chrooting , https://rancher-sandbox.github.io/cos-toolkit-docs/docs/customizing/stages/#after-install-chroot wouldn't that cover the case?

cmd/run-stage.go
mudler: Actually this is not static at all. at some point needs to be part of the config,  it can be tweaked in /etc/cos/config

Itxaka: Do we really want to make this configurable? Like this are the default paths, they should not change unless we change them no? I already have a check in place to see if the dirs exists and skip them if not to avoid any issues, not sure if making this configurable its a good idea, we already have the config value to add extra paths to this.

Moving it to constants make sense though.

mudler: no, derivative can specify custom additional cloud-init paths. https://rancher-sandbox.github.io/cos-toolkit-docs/docs/customizing/general_configuration/ see CLOUD_INIT_PATHS

I'm good with constants, but we need still to provide a way to add on top

Itxaka: yes, and that is already in a few lines down, and according to the original script those paths are appended to the default ones.

Was that the original intention? Or was it meant to substitute the default paths? Maybe the cos-setup script was wrongly written?

Itxaka: https://github.com/rancher-sandbox/elemental-cli/pull/36/files#diff-14cf243a616847f5f48ba21fb76d6aaf7e02530cba7cec3050ec6a4efe71cad6R55

And in the original:
 https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cos-setup/cos-setup#L33

mudler: > yes, and that is already in a few lines down, and according to the original script those paths are appended to the default ones.
> 
> Was that the original intention? Or was it meant to substitute the default paths? Maybe the cos-setup script was wrongly written?

didn't noticed that! we are good then :+1:

pkg/types/v1/config.go
```diff
@@ -166,6 +166,7 @@ type RunConfig struct {
 	Force           bool   `yaml:"force,omitempty" mapstructure:"force"`
 	Strict          bool   `yaml:"strict,omitempty" mapstructure:"strict"`
 	Iso             string `yaml:"iso,omitempty" mapstructure:"iso"`
+	CloudInitPaths  string `yaml:"CLOUD_INIT_PATHS,omitempty" mapstructure:"CLOUD_INIT_PATHS"`
```
mudler, line 186: this is a bit weird (the uppercase on yaml), any specific reason?

Itxaka: backwards compatibility I guess? Following up on the rest of the values above of respecting the existing keys

davidcassany (approved):

davidcassany:
@Itxaka now I realized something. I am not sure having the logic of the run-stage directly within the `cmd/run-stage.go` is the best choice. Probably making it part of an utility is better and just call such utility in `cmd/run-stage.go`. I am saying so because now I realized that we should also use this run-stage logic here:

https://github.com/rancher-sandbox/elemental-cli/blob/4dd273dda8705f8a305340c4f9cfeda9b98f1d45/pkg/action/install.go#L38-L57

Probably this can be done as a follow up PR though, as it requires to consider the chroot case. In any case we will need it as soon as we attempt to remove the `cos-setup` script from cOS-toolkit.

Itxaka:
Ah good point, I thought this wasn't gonna use and was restricted to the command only.

Will move it into a different places no problem there

Itxaka:
Some issues with dotnotation on cmdline and with what errors should be returned, as yip usually runs everything and then returns a multierror at the end, not sure if we should propagate that or do our own thing.

PR #38, opened 2022-01-11
Add global flag for debug level

### LINKED ISSUE #37: Enable configuration of loglevel
Currently there is no way of configuring the loglevel which is kind of bad, we have logging all around with debug level but there is no way of accessing it if something starts to fail.

we should have it as part of the runConfig so its loaded from files/env var and reconfigures the logger passed to the Runconfig on instance creation/struct creation.

Itxaka:
Fixes #37 

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -82,5 +82,10 @@ func ReadConfigRun(configDir string, logger v1.Logger, mounter mount.Interface) 
 
 	// unmarshal all the vars into the config object
 	viper.Unmarshal(cfg)
+
+	// Set debug level
+	if cfg.Debug {
+		cfg.Logger.SetLevel(v1.DebugLevel())
+	}
 	return cfg, nil
 }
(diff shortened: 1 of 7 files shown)


davidcassany (approved):
LGTM

Itxaka:
Gonna add a couple of tests maybe to bring those numbers up :/

Maybe we should just ignore the cmd code from being counted against code coverage, as its kind of absurd to test flags and so on...

davidcassany:
> Maybe we should just ignore the cmd code from being counted against code coverage, as its kind of absurd to test flags and so on...

Yes, it would be nice to be able to customize what is counted and what not in code coverage. I have not investigated it in any way, probably it is possible to add decorators to omit code lines or configure the omission of certain files... I agree sometimes it feels silly trying to test certain execution paths or code portions.

Itxaka:
Unfortunately there is nothing in the standard library to ignore lines...nor there are plans to add anything...

PR #41, opened 2022-01-12
Ensure the downloaded ISO is unmounted

davidcassany:

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -355,34 +355,31 @@ func (c Elemental) BootedFromSquash() bool {
 // and mount the iso file as loop,
 // and modify the IsoMnt var to point to the newly mounted dir
 func (c *Elemental) GetIso() error {
-	if c.config.Iso != "" {
-		tmpDir, err := afero.TempDir(c.config.Fs, "", "elemental")
-		if err != nil {
-			return err
-		}
-		tmpFile := filepath.Join(tmpDir, "cOs.iso")
-		err = c.GetUrl(c.config.Iso, tmpFile)
-		if err != nil {
-			defer c.config.Fs.RemoveAll(tmpDir)
-			return err
-		}
-		tmpIsoMount, err := afero.TempDir(c.config.Fs, "", "elemental-iso-mounted-")
-		if err != nil {
-			defer c.config.Fs.RemoveAll(tmpDir)
-			return err
-		}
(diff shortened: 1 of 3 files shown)


pkg/elemental/elemental.go
```diff
@@ -355,34 +355,31 @@ func (c Elemental) BootedFromSquash() bool {
...
+	}
+	tmpFile := filepath.Join(tmpDir, "cOs.iso")
+	err = c.GetUrl(c.config.Iso, tmpFile)
+	if err != nil {
+		defer c.config.Fs.RemoveAll(tmpDir)
+		return err
+	}
+	tmpIsoMount, err := afero.TempDir(c.config.Fs, "", "elemental-iso-mounted-")
+	if err != nil {
+		defer c.config.Fs.RemoveAll(tmpDir)
```
Itxaka: I wonder if there is some kind of try/catch/finally in go to deal with cleanups like this, as it seems really inefficient and prone to errors :/

pkg/elemental/elemental_test.go
```diff
@@ -528,10 +528,6 @@ var _ = Describe("Elemental", func() {
 		})
 	})
 	Context("GetIso", func() {
-		It("Does nothing if iso is not set", func() {
```
Itxaka, line 531: I guess we should move this test into the installer then

Itxaka (approved):
LGTM

pkg/elemental/elemental.go
davidcassany: well it exists through `defer` and this is kind of a design choice of golang, no exceptions but you can explicitly return errors which you have to explicitly handle if you want to, so no magic. In this particular case we could consider ignoring this sort of errors as it will just fail to mount if the directories are not there. Anyway I prefer having this explicit error management. 

However this makes me realize that we are not cleaning this files and directories in a successful execution... 🤔 we are just unmounting the mountpoint. Gonna think about how to properly do it.

pkg/elemental/elemental_test.go
davidcassany: Yes, but there is no installer test yet 😝

Itxaka (approved):

davidcassany:
@Itxaka I finally slightly modified the logic of GetIso to return the temporary directory so we can easily defer its cleanup on the caller logic.

PR #44, opened 2022-01-14
Add install action tests

davidcassany:


pkg/action/action_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"os"
+	"testing"
+)
+
+const printOutput = `BYT;
+/dev/loop0:50593792s:loopback:512:512:gpt:Loopback device:;`
+const partTmpl = `
+%d:%ss:%ss:2048s:ext4::type=83;`
+
+func TestElementalSuite(t *testing.T) {
```
mudler, line 37: usually this goes into a separate file `suite_test.go` as it is enough once for all the _test files in the same folder.. or I am maybe missing something?

mudler (approved):
Looks good here!

pkg/action/action_test.go
Itxaka: AFAIK this is not needed if there is only one test file right? If it starts to grow then we can move it to its own file I guess?

pkg/action/install.go
```diff
@@ -35,25 +35,29 @@ func NewInstallAction(config *v1.RunConfig) *InstallAction {
...
+				err = tmpErr
+			}
+		}()
 	}
-	i.Config.Logger.Debugf("%s output: %s", hook, string(out))
-	if err != nil && i.Config.Strict {
-		return err
+	i.Config.Logger.Infof("Running %s hook", hook)
+	err = utils.RunStage(hook, i.Config)
+	if !i.Config.Strict {
```
Itxaka, line 57: ah niceeee

pkg/elemental/elemental_test.go
```diff
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"fmt"
 	. "github.com/onsi/ginkgo"
-	//"github.com/onsi/ginkgo/config"
```
Itxaka, line 23: can we remove this?

Itxaka (approved):

pkg/elemental/elemental_test.go
davidcassany: This is already a removal :)

pkg/action/action_test.go
davidcassany: well I expected to have all actions being tested in this file, note this is `action_test.go` (it could also be named `suite_test`) testing `install.go`. I did it this way as I was thinking on a test suite per package and I thought that it was likely that there could be single `_test.go `. Anyway, I am fine with any approach or criteria.

PR #45, opened 2022-01-17
Set default grub entry in installer

### LINKED ISSUE #1064: Add logic for grub rebrand (GRUB_ENTRY_NAME env variable)

davidcassany:
Fixes rancher-sandbox/cOS-toolkit#1064

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -188,7 +188,7 @@ func (i InstallAction) Run() (err error) {
 	}
 
 	// installation rebrand (only grub for now)
-	err = newElemental.SetDefaultGrubEntry()
+	err = newElemental.Rebrand()
 	if err != nil {
 		return err
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -186,9 +186,12 @@ func (i InstallAction) Run() (err error) {
 	if err != nil {
 		return err
 	}
-	// TODO Rebrand
-	// cos.Rebrand()
-	// ????
+
+	// installation rebrand (only grub for now)
+	err = newElemental.SetDefaultGrubEntry()
+	if err != nil {
+		return err
+	}
 	// profit!
 	// TODO poweroff or reboot or nothing
 	return err
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -485,8 +485,14 @@ func (c Elemental) SetDefaultGrubEntry() error {
 		return errors.New("State partition not found. Cannot set grub env file")
 	}
 	grub := utils.NewGrub(c.config)
-	return grub.SetEnvFile(
+	return grub.SetPersistentVariables(
 		filepath.Join(part.MountPoint, cnst.GrubOEMEnv),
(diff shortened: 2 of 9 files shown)


pkg/types/v1/config.go
```diff
@@ -224,6 +231,15 @@ type Image struct {
 	LoopDevice string
 }
 
+func (pl PartitionList) GetByPLabel(label string) *Partition {
```
Itxaka, line 234: :+1:

pkg/utils/grub.go
```diff
@@ -126,3 +126,15 @@ func (g Grub) Install() error {
 	g.config.Logger.Infof("Grub install to device %s complete", g.config.Target)
 	return nil
 }
+
+// Sets the given key value pairs into as grub variables into the given file
+func (g Grub) SetEnvFile(grubEnvFile string, vars map[string]string) error {
```
Itxaka: maybe a more clear name? At first glance this tells me that it sets the env file for grub, not that its writing k:v to the given envfile...

pkg/action/install.go
```diff
@@ -186,9 +186,12 @@ func (i InstallAction) Run() (err error) {
 	if err != nil {
 		return err
 	}
-	// TODO Rebrand
-	// cos.Rebrand()
-	// ????
+
+	// installation rebrand (only grub for now)
+	err = newElemental.SetDefaultGrubEntry()
```
Itxaka: should we have a more generic method `Rebrand` that does this inside instead so we can grow it organically?

Itxaka (approved):
all nits, looks good to me

pkg/action/install.go
davidcassany: I thought about it, sure we can. I did not since it would have been an Elemental method including this single call. But yes, probably makes sense. Gonna add it.

pkg/utils/grub.go
davidcassany: good point, what about `SetPersistentVariables`? I am also opened to any idea

pkg/elemental/elemental.go
```diff
@@ -485,8 +485,14 @@ func (c Elemental) SetDefaultGrubEntry() error {
...
-	return grub.SetEnvFile(
+	return grub.SetPersistentVariables(
 		filepath.Join(part.MountPoint, cnst.GrubOEMEnv),
 		map[string]string{"default_menu_entry": c.config.GrubDefEntry},
 	)
 }
+
+// Runs rebranding procedure. Note this assumes all required partitions and
+// images to be mounted in advance.
+func (c Elemental) Rebrand() error {
```
Itxaka, line 496: nice

pkg/utils/grub.go
```diff
@@ -128,7 +128,7 @@ func (g Grub) Install() error {
 }
 
 // Sets the given key value pairs into as grub variables into the given file
-func (g Grub) SetEnvFile(grubEnvFile string, vars map[string]string) error {
+func (g Grub) SetPersistentVariables(grubEnvFile string, vars map[string]string) error {
```
Itxaka, line 131: nice

Itxaka (approved):

PR #46, opened 2022-01-18
Ignore errors from run-stage by default and use strict mode

Itxaka:
As cos-setup didnt raise any errors, we have to mimic the same behaviour
as some services depend on a clean exit to continue booting the system.

Also log those errors and introduce a check for strict mode in which the
errors are returned and elemental exits with an error in that case


mudler (approved):
Looks good here!

PR #48, opened 2022-01-20
Restore CloudInitRunner modifier after runStage

davidcassany:
This commit ensure the yips loader modifier is reset to nil after
running utils.RunStage.

In addition this commit also hides the usage message on errors
during the installation.

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -52,6 +52,7 @@ var installCmd = &cobra.Command{
 		cfg.Target = args[0]
 
 		cfg.DigestSetup()
+		cmd.SilenceUsage = true
 
 		cfg.Logger.Infof("Install called")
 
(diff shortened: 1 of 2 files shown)


davidcassany:
These two details were driving me mad 😓

cmd/install.go
```diff
@@ -52,6 +52,7 @@ var installCmd = &cobra.Command{
 		cfg.Target = args[0]
 
 		cfg.DigestSetup()
+		cmd.SilenceUsage = true
```
Itxaka, line 55: Isn't better to use silence errors?

Itxaka (approved):

cmd/install.go
davidcassany: I am not sure which is the difference. After DigestSetup() if there was something wrong with the flags (aka use of incompatible flags) the error should be raised there. Even for now we are no doing any consistency validation on flags. So my understanding is that from that stage flags are already validated hence the usage message on errors has no sense.

PR #50, opened 2022-01-21
do not skip paths on run-stage

Itxaka:
As paths can be created during run, we should not skip paths ourselves,
let the errors let the user know if those paths were not there


Itxaka:
Probably the path not being there should not be an error but a warning, as paths can be created on the fly in the yip run. Maybe a change to send upstream to yip...

davidcassany (approved):
Looks good

PR #51, opened 2022-01-21
Fix CopyRecovery

davidcassany:
CopyRecovery should only be effective if we are not booting from
an squashed recovery system.

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -361,7 +361,8 @@ func (c *Elemental) CheckNoFormat() error {
 
 // BootedFromSquash will check if we are booting from squashfs
 func (c Elemental) BootedFromSquash() bool {
-	if utils.BootedFrom(c.config.Runner, cnst.RecoveryLabel) {
+	part := c.config.Partitions.GetByPLabel(cnst.RecoveryPLabel)
+	if part != nil && utils.BootedFrom(c.config.Runner, part.Label) {
 		return true
 	}
 	return false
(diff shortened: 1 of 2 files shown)


pkg/elemental/elemental.go
```diff
@@ -361,7 +361,8 @@ func (c *Elemental) CheckNoFormat() error {
 
 // BootedFromSquash will check if we are booting from squashfs
 func (c Elemental) BootedFromSquash() bool {
-	if utils.BootedFrom(c.config.Runner, cnst.RecoveryLabel) {
+	part := c.config.Partitions.GetByPLabel(cnst.RecoveryPLabel)
+	if part != nil && utils.BootedFrom(c.config.Runner, part.Label) {
```
davidcassany, line 365: Check against the configured label as it could potentially be something else than cnst.RecoveryLabel

Itxaka (approved):

pkg/elemental/elemental.go
```diff
@@ -444,33 +444,35 @@ func (c *Elemental) GetUrl(url string, destination string) error {
 // either is get from the system if NoFormat is enabled (searching for label COS_RECOVERY) or is a newly generated partition
 func (c *Elemental) CopyRecovery() error {
 	var err error
-	if !c.BootedFromSquash() {
+	if c.BootedFromSquash() {
```
davidcassany, line 447: Note in https://github.com/rancher-sandbox/cOS-toolkit/blob/45bc7833fccfb7a7d8990fa7310237c8e4422f7e/packages/installer/cos.sh#L1476-L1478 this logic is inverted.

All this in `cos.sh` is so obscure and confusing, my mind gets fried every time I read this part. Also variable names in cos.sh are misleading so it is easy to get confused.

PR #52, opened 2022-01-24
Adds pull-image

Itxaka:
Basically a copy of luet's util unpack command

### DIFF
--- a/cmd/pull-image.go
+++ b/cmd/pull-image.go
@@ -1,0 +1,96 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"github.com/docker/docker/api/types"
+	"github.com/mudler/luet/pkg/api/core/context"
+	"github.com/rancher-sandbox/elemental/cmd/config"
+	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


cmd/pull-image.go
```diff
@@ -0,0 +1,97 @@
+/*
+Copyright © 2021 SUSE LLC
```
mudler:
```suggestion
Copyright © 2022 SUSE LLC
```

mudler (approved):

cmd/pull-image.go
```diff
@@ -0,0 +1,97 @@
...
+		auth := &types.AuthConfig{
+			Username:      user,
+			Password:      pass,
+			ServerAddress: server,
+			Auth:          authType,
+			IdentityToken: identity,
+			RegistryToken: registryToken,
+		}
+
+		info, err := docker.DownloadAndExtractDockerImage(util.DefaultContext, image, destination, auth, verify)
```
davidcassany: I would use and extend the Luet interface instead, so we have a single point of connection between luet and elemental. 

https://github.com/rancher-sandbox/elemental/blob/30ddad2f81f189297b283037bdb4d6e9613e3a8f/pkg/types/v1/luet.go#L50-L60

We could use a map for extra configs, pass the auth struct as parameter, make it part of the constructor of NewLuet, etc.

In any case I'd try to keep luet wrapped into the interface.

davidcassany (commented):
I am just wondering if we could eventually extend the interface of Luet and make use of it here.

cmd/pull-image.go
Itxaka: Uh, didn't even knew that we had that already, so yeah I will change to use the luet.unpack directly indeed.

Thanks for the pointer!

Itxaka: There, actually is much nicer :D

cmd/root.go
```diff
@@ -30,7 +31,10 @@ var rootCmd = &cobra.Command{
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	cobra.CheckErr(rootCmd.Execute())
+	err := rootCmd.Execute()
```
Itxaka, line 34: Otherwise, this prints the error to stdout, so we better pick the error and exit ourselves

davidcassany (approved):
LGTM

Issue #22, opened 2021-12-16
Consider using ghw for disks and host information


mudler:
Just a spare idea , maybe we can replace the udevadm/partprobe calls with something native like https://github.com/jaypipes/ghw#block-storage

cc @davidcassany

davidcassany:
> Just a spare idea , maybe we can replace the udevadm/partprobe calls with something native like https://github.com/jaypipes/ghw#block-storage
> 
> cc @davidcassany

Fine with me, as long as it works. I'd suggest to start with the current implementation and move to something native if possible. The challenging part is reacting to partition table changes, udevadm turned to be a good option to wait until the kernel reports new devices.

davidcassany:
I am thinking about if it would make sense to use it to validate partitions when using the `--no-format` installer option. Gonna check if this is an straight forward exercise, if so it can be a good place to start trying this library. Removing `udevadm` from partitioner is scary as race conditions like what harvester reported a month ago could pop up again and they were not really popping up consistently in our tests. I'd love to drop `partprobe` and `udevadm`, but I am just wondering how to do it in safe way so we do not have to turn back to them again 😅 We need to find some playground to experiment with alternatives.

Itxaka:
ghw doesnt seem to be wht we need. It only reads partitions that are mounted, so we cannot use it to find unmounted partitions.

mudler:
I've used it like so to list all available disks (even not mounted ones): 

```golang
	disks := []string{}
	block, err := ghw.Block()
	if err == nil {
		for _, disk := range block.Disks {
			disks = append(disks, fmt.Sprintf("/dev/%s", disk.Name))
		}
	}
```

each Disk has  `Partitions` , but I didn't used it so can't tell if it does what we need

Itxaka:
it reads from the /proc/mounts to get the info on partitions, https://github.com/jaypipes/ghw/blob/main/pkg/block/block_linux.go#L388 so it wont work for our use case. Which I understand, its the simplest and more straightforward way of getting info on partitions, and the point of the lib is to get info that is available now, not what it could do in the future, so it makes sense they only report that.

mudler:
> it reads from the /proc/mounts to get the info on partitions, https://github.com/jaypipes/ghw/blob/main/pkg/block/block_linux.go#L388 so it wont work for our use case. Which I understand, its the simplest and more straightforward way of getting info on partitions, and the point of the lib is to get info that is available now, not what it could do in the future, so it makes sense they only report that.

[...]

PR #57, opened 2022-01-26
Add retries logic to get devices by label

davidcassany:

### DIFF
--- a/pkg/elemental/elemental_test.go
+++ b/pkg/elemental/elemental_test.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/config"
 	. "github.com/onsi/gomega"
 	cnst "github.com/rancher-sandbox/elemental/pkg/constants"
 	"github.com/rancher-sandbox/elemental/pkg/elemental"
@@ -39,7 +40,7 @@ const partTmpl = `
 
 func TestElementalSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
-	//config.DefaultReporterConfig.SlowSpecThreshold = 10
+	config.DefaultReporterConfig.SlowSpecThreshold = 11
 	RunSpecs(t, "Elemental test suite")
 }
 
@@ -98,7 +99,7 @@ var _ = Describe("Elemental", func() {
 
 		It("Fails if oem partition is not found ", func() {
 			runner.SideEffect = func(cmd string, args ...string) ([]byte, error) {
-				if len(args) >= 2 && args[1] == fmt.Sprintf("LABEL=%s", cnst.OEMLabel) {
+				if len(args) >= 2 && args[1] == cnst.OEMLabel {
 					return []byte{}, nil
 				}
 				return []byte("/some/device"), nil
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"os/exec"
 	"strings"
+	"time"
 )
 
 func CommandExists(command string) bool {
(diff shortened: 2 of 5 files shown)


davidcassany:
Hopefully this will fix some issues I see on the cOS-toolkit CI when using the elemental installer together with the yip layout plugin. The problem is that I could not find a way to asses it locally, CI installer tests succeed on my local env and mostly fails always for EFI in CI and sometimes for Bios. Since the failure is always on fetching devices by label I added a retry logic within the common utility, the partitioner already has equivalent logic.

pkg/utils/common.go
```diff
@@ -38,25 +39,19 @@ func BootedFrom(runner v1.Runner, label string) bool {
...
-	if err != nil {
-		return "", err
-	}
-	if strings.TrimSpace(string(out)) == "" {
-		return "", errors.New("no device found")
+// GetDeviceByLabel will try to return the device that matches the given label.
+// deviceTries value sets the number of attempts to find the device, it
+// waits a second between attempts.
+func GetDeviceByLabel(runner v1.Runner, label string, deviceTries int) (string, error) {
+	for tries := 0; tries < deviceTries; tries++ {
```
Itxaka: NIT: is there any lib that does this already? We may need to use it in several places so it would be good to have a lib that does the retry for us and covers cases that we may miss.

pkg/utils/common.go
```diff
@@ -38,25 +39,19 @@ func BootedFrom(runner v1.Runner, label string) bool {
...
+		runner.Run("udevadm", "settle")
+		out, _ := runner.Run("blkid", "--label", label)
+		if strings.TrimSpace(string(out)) != "" {
```
mudler: maybe shall we check the error here? just thinking as if it does exit for some failure and returns some output we might just parse it accidentally

pkg/elemental/elemental_test.go
```diff
@@ -39,7 +40,7 @@ const partTmpl = `
 
 func TestElementalSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
-	//config.DefaultReporterConfig.SlowSpecThreshold = 10
+	config.DefaultReporterConfig.SlowSpecThreshold = 11
```
Itxaka, line 43: why do we need this? Isnt this only to report X as a slow test?

mudler:
I'm wondering, maybe for a future note, if we should parse in some struct the output of `lsblk -J` (as it is JSON, and easy to parse) and poll that instead, overall it's much easier than retrieving the single data from blkid

pkg/elemental/elemental_test.go
davidcassany: I just added it to prevent reporting there is a slow test. Not really needed but I prefer not having this report specially if the test is designed to take around 10 seconds.

Itxaka:
we probably would benefit from having a lib that deal with disks native to go, including the blkid stuff into structs and partitioning and such. Time to fork ghw and add the missing stuff? :P

mudler (approved):
Besides small nits it looks good!

Itxaka (approved):

pkg/utils/common.go
davidcassany: Yes all this sucks... I could not find a lib for that (I did not investigate much to be honest). There is the lib exposed in #22 but in this particular case it won't work according the docs. The lib is not designed to monitor dynamic changes, only to extract static data from `/sys`, the same `blkid` or `lsblk` do. So the problem is that after running yip partitioning stage there is no guarantee that the kernel already parsed all changes.

davidcassany:
> I'm wondering, maybe for a future note, if we should parse in some struct the output of `lsblk -J` (as it is JSON, and easy to parse) and poll that instead, overall it's much easier than retrieving the single data from blkid

In fact, I started a refactor to only use `lsblk` instead and only inside the `partitioner` to focus all this crap in a single place, however the impact of the changes were getting bigger that I thought they would be at glance and went back the straight forward implementation.

davidcassany:
I agree with you both that would be nice to find a better solution for that.

* One option, for this very particular case, would be to add some more logic in yip layout plugin to make use kernel detected all new devices before leaving as a sanity check.
* Wrapping all device parsers in a single place (being it our own package or by using a third party library) would be also a better approach, this implies a little refactor and probably does not prevent us from having some ugly trial and error logic.

PR #59, opened 2022-01-27
Add upgrade command

Itxaka:

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -1,0 +1,94 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"errors"
+	"github.com/rancher-sandbox/elemental/cmd/config"
+	"github.com/rancher-sandbox/elemental/pkg/action"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"k8s.io/mount-utils"
+	"os/exec"
+)
+
+// upgradeCmd represents the upgrade command
+var upgradeCmd = &cobra.Command{
+	Use:   "upgrade",
+	Short: "upgrade the system",
+	Args:  cobra.ExactArgs(0),
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		path, err := exec.LookPath("mount")
+		if err != nil {
+			return err
+		}
+		mounter := mount.New(path)
+
+		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), mounter)
+
+		if err != nil {
+			cfg.Logger.Errorf("Error reading config: %s\n", err)
+		}
+
+		// Init setup
+		err = cfg.DigestSetup()
+		if err != nil {
+			cfg.Logger.Errorf("Error digesting setup: %s\n", err)
+		}
+
+		// docker-image and directory are mutually exclusive. Can't have your cake and eat it too.
+		if viper.GetString("docker-image") != "" && viper.GetString("directory") != "" {
+			msg := "flags docker-image and directory are mutually exclusive, please only set one of them"
+			return errors.New(msg)
+		}
+
+		if cfg.DockerImg != "" || cfg.DirectoryUpgrade != "" {
+			// Force channel upgrades to be false, because as its loaded from the config files,
+			// it will probably always be set to true due to it being the default value
+			cfg.ChannelUpgrades = false
+		}
+		// Set this after parsing of the flags, so it fails on parsing and prints usage properly
+		cmd.SilenceUsage = true
+		cmd.SilenceErrors = true // Do not propagate errors down the line, we control them
+
+		upgrade := action.NewUpgradeAction(cfg)
+		err = upgrade.Run()
+		if err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(upgradeCmd)
+	upgradeCmd.Flags().String("docker-image", "", "Install a specified container image")
+	upgradeCmd.Flags().String("directory", "", "Use directory as source to install from")
+	upgradeCmd.Flags().Bool("recovery", false, "Upgrade the recovery")
+	upgradeCmd.Flags().Bool("no-verify", false, "Disable mtree verification")
+	upgradeCmd.Flags().Bool("no-cosign", false, "Disable cosign verification")
+	upgradeCmd.Flags().Bool("strict", false, "Fail on any errors")
+	upgradeCmd.Flags().BoolP("reboot", "", false, "Reboot the system after install")
+	upgradeCmd.Flags().BoolP("poweroff", "", false, "Shutdown the system after install")
+	// We bind the --directory into the DirectoryUpgrade value directly to have a more explicit var in the config
+	viper.BindPFlag("DirectoryUpgrade", upgradeCmd.Flags().Lookup("directory"))
+	// We bind the --recovery flag into RecoveryUpgrade value to have a more explicit var in the config
+	viper.BindPFlag("RecoveryUpgrade", upgradeCmd.Flags().Lookup("recovery"))
+	// bind the rest of the flags into their direct values as they are mapped 1to1
+	viper.BindPFlags(upgradeCmd.Flags())
+}
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -31,6 +31,14 @@ var upgradeCmd = &cobra.Command{
 	Use:   "upgrade",
 	Short: "upgrade the system",
 	Args:  cobra.ExactArgs(0),
+	PreRun: func(cmd *cobra.Command, args []string) {
+		// We bind the --directory into the DirectoryUpgrade value directly to have a more explicit var in the config
+		viper.BindPFlag("DirectoryUpgrade", cmd.Flags().Lookup("directory"))
+		// We bind the --recovery flag into RecoveryUpgrade value to have a more explicit var in the config
+		viper.BindPFlag("RecoveryUpgrade", cmd.Flags().Lookup("recovery"))
+		// bind the rest of the flags into their direct values as they are mapped 1to1
+		viper.BindPFlags(cmd.Flags())
+	},
 
 	RunE: func(cmd *cobra.Command, args []string) error {
 		path, err := exec.LookPath("mount")
@@ -85,10 +93,4 @@ func init() {
 	upgradeCmd.Flags().Bool("strict", false, "Fail on any errors")
 	upgradeCmd.Flags().BoolP("reboot", "", false, "Reboot the system after install")
 	upgradeCmd.Flags().BoolP("poweroff", "", false, "Shutdown the system after install")
-	// We bind the --directory into the DirectoryUpgrade value directly to have a more explicit var in the config
-	viper.BindPFlag("DirectoryUpgrade", upgradeCmd.Flags().Lookup("directory"))
-	// We bind the --recovery flag into RecoveryUpgrade value to have a more explicit var in the config
-	viper.BindPFlag("RecoveryUpgrade", upgradeCmd.Flags().Lookup("recovery"))
-	// bind the rest of the flags into their direct values as they are mapped 1to1
-	viper.BindPFlags(upgradeCmd.Flags())
 }
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -1,0 +1,435 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package action
+
+import (
+	"errors"
+	"fmt"
+	"github.com/hashicorp/go-multierror"
+	"github.com/rancher-sandbox/elemental/pkg/constants"
+	"github.com/rancher-sandbox/elemental/pkg/elemental"
+	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
+	"github.com/rancher-sandbox/elemental/pkg/utils"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"k8s.io/mount-utils"
+	"os"
+	"path/filepath"
+)
+
+// UpgradeSource represents a source object to identify where it's coming from easily
+type UpgradeSource struct {
+	source    string
+	isDir     bool
+	isChannel bool
+	isDocker  bool
+}
+
+// UpgradeAction represents the struct that will run the upgrade from start to finish
+type UpgradeAction struct {
+	Config *v1.RunConfig
+}
+
+// Cleanup is the struct that will be passed to the cleanup function at the end of the upgrade or at any error before returning
+type Cleanup struct {
+	Unmount []string
+	Remove  []string
+	Remount []mount.MountPoint
+}
+
+func NewUpgradeAction(config *v1.RunConfig) *UpgradeAction {
+	return &UpgradeAction{Config: config}
+}
+
+func (u *UpgradeAction) Info(s string, args ...interface{}) {
+	u.Config.Logger.Infof(s, args...)
+}
+
+func (u *UpgradeAction) Debug(s string, args ...interface{}) {
+	u.Config.Logger.Debugf(s, args...)
+}
+
+func (u *UpgradeAction) Error(s string, args ...interface{}) {
+	u.Config.Logger.Errorf(s, args...)
+}
+
+func (u *UpgradeAction) upgradeHook(hook string, chroot bool, rebrand bool) (err error) {
+	if chroot { // TODO: This is not really doing anything for hooks as it closes the chroot before running the hooks?
+		newChroot := utils.NewChroot(u.Config.ActiveImage.MountPoint, u.Config)
+		newChroot.SetExtraMounts(map[string]string{
+			"/usr/local": "/usr/local",
+			"/oem":       "/oem",
+		})
+		err = newChroot.Prepare()
+		if err != nil {
+			return err
+		}
+
+		defer func() {
+			if tmpErr := newChroot.Close(); tmpErr != nil && err == nil {
+				err = tmpErr
+			}
+		}()
+
+		if rebrand {
+			grubEnvFile := filepath.Join("/", constants.GrubOEMEnv)
+			vars := map[string]string{"default_menu_entry": u.Config.GrubDefEntry}
+			for key, value := range vars {
+				out, err := newChroot.Run("grub2-editenv", grubEnvFile, "set", fmt.Sprintf("%s=%s", key, value))
+				if err != nil {
+					u.Error("Failed setting grub variables: %v", out)
+					if u.Config.Strict {
+						return err
+					}
+				}
+			}
+		}
+	}
+	u.Info("Running %s hook", hook)
+	// Sorry, but I dont want to see all the yip stuff
+	oldLevel := u.Config.Logger.GetLevel()
+	u.Config.Logger.SetLevel(logrus.ErrorLevel)
+	err = utils.RunStage(hook, u.Config)
+	u.Config.Logger.SetLevel(oldLevel)
+	u.Info("Hook %s run", hook)
+	if u.Config.Strict {
+		return err
+	} else {
+		if err != nil {
+			u.Debug("Strict disabled, setting err to nil: %v", err)
+		}
+		return nil
+	}
+}
+
+func (u *UpgradeAction) Run() error {
+	var statePart v1.Partition
+	var err error
+	var transitionImg string
+	cleanup := Cleanup{Remove: []string{constants.UpgradeTempDir}}
+	upgradeStateDir := constants.UpgradeStateDir
+
+	// if upgrading the recovery we mount the state in a different place as its already mounted RO, we need to remount it
+	if u.Config.RecoveryUpgrade {
+		upgradeStateDir = constants.UpgradeRecoveryDir
+	}
+
+	upgradeTarget, upgradeSource := u.getTargetAndSource()
+	// TODO: support channel upgrades
+	if upgradeSource.isChannel {
+		msg := "channel upgrade not supported yet"
+		u.Error(msg)
+		return errors.New(msg)
+	}
+	u.Config.Logger.Infof("Upgrading %s partition", upgradeTarget)
+
+	err = u.Config.Fs.MkdirAll(constants.UpgradeTempDir, os.ModeDir)
+	if err != nil {
+		u.Error("Error creating target dir %s: %s", constants.UpgradeTempDir, err)
+		return u.cleanup(cleanup, err)
+	}
+
+	cleanup.Remove = append(cleanup.Remove, constants.UpgradeTempDir)
+
+	if u.Config.RecoveryUpgrade {
+		statePart, err = utils.GetFullDeviceByLabel(u.Config.Runner, u.Config.RecoveryLabel, 5)
+		if err != nil {
+			return u.cleanup(cleanup, err)
+		}
+	} else {
+		statePart, err = utils.GetFullDeviceByLabel(u.Config.Runner, u.Config.StateLabel, 5)
+		if err != nil {
+			return u.cleanup(cleanup, err)
+		}
+	}
+
+	u.Info("Mounting state partition %s in %s", statePart.Path, upgradeStateDir)
+	if exists, _ := afero.Exists(u.Config.Fs, upgradeStateDir); !exists {
+		err = u.Config.Fs.MkdirAll(upgradeStateDir, os.ModeDir)
+
+		if err != nil {
+			u.Error("Error creating statedir %s: %s", upgradeStateDir, err)
+			return u.cleanup(cleanup, err)
+		}
+	}
+
+	statePartMountOptions := []string{"remount", "rw"}
+
+	// If we want to upgrade the active but are booting from recovery, the statedir is not mounted, so dont remount
+	if !u.Config.RecoveryUpgrade && utils.BootedFrom(u.Config.Runner, u.Config.RecoveryLabel) {
+		statePartMountOptions = []string{"rw"}
+		cleanup.Unmount = append(cleanup.Unmount, upgradeStateDir)
+	}
+
+	// If we want to upgrade the recovery but are not booting from recovery, the stateDir is not mounted, so dont try to remount
+	if u.Config.RecoveryUpgrade && !utils.BootedFrom(u.Config.Runner, u.Config.RecoveryLabel) {
+		statePartMountOptions = []string{"rw"}
+		cleanup.Unmount = append(cleanup.Unmount, upgradeStateDir)
+	}
+
+	err = u.Config.Mounter.Mount(statePart.Path, upgradeStateDir, statePart.FS, statePartMountOptions)
+	if err != nil {
+		u.Error("Error mounting %s: %s", upgradeStateDir, err)
+		return u.cleanup(cleanup, err)
+	}
+
+	if !utils.BootedFrom(u.Config.Runner, u.Config.RecoveryLabel) {
+		cleanup.Remount = append(cleanup.Remount, mount.MountPoint{Device: statePart.Path, Path: upgradeStateDir, Type: statePart.FS})
+	}
+
+	// Track if recovery.squash file exists which indeicates that the recovery is squash
+	isSquashRecovery, _ := afero.Exists(u.Config.Fs, filepath.Join(upgradeStateDir, "cOS", constants.RecoverySquashFile))
+
+	if isSquashRecovery {
+		u.Debug("Recovery is squash")
+		transitionImg = filepath.Join(upgradeStateDir, "cOS", constants.TransitionSquashFile)
+	} else {
+		transitionImg = filepath.Join(upgradeStateDir, "cOS", constants.TransitionImgFile)
+	}
+
+	u.Debug("Using transition img: %s", transitionImg)
+
+	cleanup.Remove = append(cleanup.Remove, transitionImg)
+
+	// create transition.img
+	img := v1.Image{
+		File:       transitionImg,
+		Size:       u.Config.ImgSize,
+		Label:      u.Config.ActiveLabel,
+		FS:         constants.LinuxImgFs,
+		MountPoint: constants.UpgradeTempDir,
+		RootTree:   upgradeSource.source, // if source is a dir it will copy from here, if it's a docker img it uses Config.DockerImg IN THAT ORDER!
+	}
+
+	// If on recovery, set the label to the RecoveryLabel instead
+	if utils.BootedFrom(u.Config.Runner, u.Config.RecoveryLabel) {
+		img.Label = u.Config.SystemLabel
+	}
+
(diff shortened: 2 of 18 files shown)


cmd/upgrade.go
```diff
@@ -0,0 +1,74 @@
...
+
+		// docker-image and directory are mutually exclusive. Can't have your cake and eat it too.
+		if viper.GetString("docker-image") != "" && viper.GetString("directory") != "" {
+			msg := "flags docker-image and directory are mutually exclusive, please only set one of them"
+			return errors.New(msg)
+		}
+
+		// Set this after parsing of the flags, so it fails on parsing and prints usage properly
+		cmd.SilenceUsage = true
+		cmd.SilenceErrors = true // Do not propagate errors down the line, we control them
```
davidcassany, line 75: I am fine with this, but it is not clear to me what it does. Which errors are we omitting?

pkg/action/upgrade.go
```diff
@@ -0,0 +1,125 @@
...
+func (u *UpgradeAction) Run() error {
+	upgradeTarget, upgradeSource := u.getTargetAndSource()
+	u.Config.Logger.Infof("Upgrading %s partition", upgradeTarget)
+	info := u.Config.Logger.Infof
+
+	// on recovery statedir is /tmp/recovery
+	// on non-recovery statedir is /run/initramfs/cos-state
+
+	// NON-RECOVERY path
+	// prepare statedir, if booting from recovery mount it, else remount state RW
```
davidcassany: Why not simply mount statedir as `rw`? remount if already mounted or simply mount it if not mounted.

I'd say checking how I booted is not the best option as it could be recovery, passive, ISO, etc. I'd say better stick to find state label (mount or remount as needed) and apply the upgraded image in there.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,125 @@
...
+	// prepare target (create img file,grow it to expected size,mkfs ext2, mount as loop on statedir/cos/transition.img
```
davidcassany: JFYI Elemental has already a method to create and format an image file given a size and a filesystem.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,125 @@
...
+	// switch images:
+	//   - check what the current image is ({active,pasive}.img, based on what is mounted from lsblk)
+	//     - if we cant find active/passive it means we could be booting from iso, ask the user? flag?
+	//     in any case it selects the active.img by default if you want to continue with upgrading
+	//     - if we are on active.img, move active to passive and set the label to passive on it
+	//   - Overwrite active.img with transition.img and set the active label on it
+	// Finish
+
+	// RECOVERY path
+	// prepare statedir, create dir, mount recovery partition into statedir
```
davidcassany: why the recovery partition has to be mounted in statedir...? makes no sense, cos.sh has some mountpoint variables absolutely insane and abused.

I'd do the same as I suggested for the Active upgrade, find recovery partition (mount or remount as needed) and apply the recovery image in there. I am not even sure it is required to check if we are in recovery partition at this point in time.

davidcassany (approved):
LGMT :+1:, just some comments regarding the logic to be implemented

cmd/upgrade.go
Itxaka: yeah the only thing this does is prevent cobra from spitting the error itself, so we avoid double error printing (one to log and one to os.stdout)

Its simple to do it ourselves and print errors on our log directly and avoid cobra doing anything.

pkg/action/upgrade.go
Itxaka: yeas, this will be streamlined in the next commits...I hope :D

pkg/action/upgrade.go
Itxaka: niiiiiiice

pkg/action/upgrade.go
Itxaka: I dont understand it either. I think it all has to do with all the vars as you mentione, cos.sh does heavy use of the same vars all around just overwriting them in different places, which makes everything confuse af.

I think there is a small diff in the path if we are on recovery or not, but was not able to understand it properly.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,125 @@
...
+	//   - if squashfs:
+	//     - remove /run/cos/target, create /run/cos/target on statedir/tmp/target
+	//   - if not:
+	//     - prepare target (create img file,grow it to expected size,mkfs ext2, mount as loop on statedir/cos/transition.img
+	// proceed to install from source to target:
+	//   - luet install if source is channel releases
+	//   - rsync if source is directory
+	//   - elemental pull-image if source is docker-image
+	// switch images:
```
Itxaka: I think this is not correct, I screwed up here, cause if we want to upgrade the recovery what does the active/passive images has to do with anything??? I need to have a look at this again

pkg/action/upgrade.go
davidcassany: > I think there is a small diff in the path if we are on recovery or not, but was not able to understand it properly.

Yes I also struggle to identify some of the cos.sh paths. Anyway, in these cases I'd try to make the functionality clear and obvious even if there small behavior changes in some corner cases. I think some behaviors of cos.sh could be simple accidents caused by code reuse in different contexts.

Itxaka: yep yep, the reasoning of putting the cos.sh workflow here is to be clear what is doing and try to make it as simple as possible while reusing most of the paths in a clear way.

davidcassany: I could not resist. because of stuff like this we should be careful copying cos.sh logic:
https://github.com/rancher-sandbox/cOS-toolkit/blob/2db1242a2bdb8907e65d027cc11bfbf802874c62/packages/installer/cos.sh#L1213-L1218

a global variable called `_STATEDIR` is set to a device (wasn't supposed to be a directory?) here while in many other parts of the code is set to the mountpoint of the device... 😅

pkg/types/v1/config.go
```diff
@@ -165,29 +165,34 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
...
+	NoFormat         bool   `yaml:"no-format,omitempty" mapstructure:"no-format"`
+	Force            bool   `yaml:"force,omitempty" mapstructure:"force"`
+	Strict           bool   `yaml:"strict,omitempty" mapstructure:"strict"`
+	Iso              string `yaml:"iso,omitempty" mapstructure:"iso"`
+	DockerImg        string `yaml:"docker-image,omitempty" mapstructure:"docker-image"`
+	Cosign           bool   `yaml:"no-cosign,omitempty" mapstructure:"no-cosign"`
+	NoVerify         bool   `yaml:"no-verify,omitempty" mapstructure:"no-verify"`
+	CloudInitPaths   string `yaml:"CLOUD_INIT_PATHS,omitempty" mapstructure:"CLOUD_INIT_PATHS"`
+	GrubDefEntry     string `yaml:"GRUB_ENTRY_NAME,omitempty" mapstructure:"GRUB_ENTRY_NAME"`
+	DirectoryUpgrade string // configured only via flag, no need to map it to any config
```
davidcassany, line 199: I don't understand that, wouldn't `yaml:"directory,omitempty" mapstructure:"directory"` do the same without having use `viper.BindPFlag` in command configuration?

davidcassany: OH I think I understand now. You don't want to map it into a config file, it can only be accessible as flag, it has to be set at runtime.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,230 @@
...
+	err = u.Config.Fs.MkdirAll(constants.UpgradeTempDir, os.ModeDir)
+	if err != nil {
+		u.Error("Error creating target dir %s: %s", constants.UpgradeTempDir, err)
+		return err
+	}
+	defer u.Config.Fs.RemoveAll(constants.UpgradeTempDir)
+
+	// mount state partition on /run/initramfs/cos-state
+	if u.Config.RecoveryUpgrade {
+		statePart, err = utils.GetFullDeviceByLabel(u.Config.Runner, constants.RecoveryLabel, 5)
```
davidcassany: mmm we have to carefully think about this `constants.RecoveryLabel` usage, note this is configurable, the constant just represents the default. Using RECOVERY_LABEL to something different in `/etc/cos/config` would break it.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		statePart, err = utils.GetFullDeviceByLabel(u.Config.Runner, constants.StateLabel, 5)
```
davidcassany: The same as above. I'd say we could relay on RunConfig.Partitions for this.

cmd/upgrade.go
```diff
@@ -0,0 +1,81 @@
...
+		mounter := mount.New(path)
+
+		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), mounter)
+
+		if err != nil {
+			cfg.Logger.Errorf("Error reading config: %s\n", err)
+		}
+
+		// docker-image and directory are mutually exclusive. Can't have your cake and eat it too.
+		if viper.GetString("docker-image") != "" && viper.GetString("directory") != "" {
```
davidcassany, line 63: Consider it a nit: I'd move this logic to `RunConfig.DigestSetup()` method (or probably makes sense having `DigestInstallSetup` and  `DisgestUpgradeSetup`). Also if you run this DigestSetup method you will get the partitions list populated, it could be useful to check labels and probably other stuff.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return err
+		}
+		// When not on recovery, the partition needs to be remounted, so add that option
+		statePartMountOptions = append(statePartMountOptions, "remount")
+	}
+
+	u.Info("Mounting state partition %s in %s", statePart.Path, constants.UpgradeStateDir)
+	err = u.Config.Fs.MkdirAll(constants.UpgradeStateDir, os.ModeDir)
```
davidcassany: Here if you have the partitions list defined you could eventually use the `Elemental.MountPartition` method and `Elemental.Unmount`, it already creates the directories if missing and the Unmount is also reentrant.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// Final step, move the newly updated img/squash into the proper place
+	if utils.BootedFrom(u.Config.Runner, constants.RecoverySquashFile) {
+		options := constants.GetDefaultSquashfsOptions()
+		err = utils.CreateSquashFS(u.Config.Runner, u.Config.Logger, constants.UpgradeTempDir, transitionImg, options)
+		if err != nil {
+			return err
+		}
+	} else {
+		if utils.BootedFrom(u.Config.Runner, constants.ActiveLabel) {
```
davidcassany: This label is also configurable, so the constant should not be used here.

davidcassany (commented):
Looks good to me. I believe we can reuse a little more from installer though. Also constants of labels should not be used, this is configurable data, you should relay on RunConfig for that.

pkg/action/upgrade.go
Itxaka: umm, arent those values overloaded from the config? This is how it works on cos.sh, it uses `$RECOVERY_LABEL` to do all actions (search for partitions mainly) and others to do the upgrade.

Isnt the /etc/cos/config set on install, so the partition labels wont be changed??

Im very confused right now, because I can see what you mean, but I dont understand if it would even be possible to break it that way.

Itxaka: ah no, this is using constants!! my bad!

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+	var transitionImg string
+	upgradeStateDir := constants.UpgradeStateDir
+
+	// if upgrading the recovery we mount the state in a different place as its already mounted RO, we need to remount it
+	if u.Config.RecoveryUpgrade {
+		upgradeStateDir = constants.UpgradeRecoveryDirSquash
+	}
+
+	if utils.BootedFrom(u.Config.Runner, constants.RecoverySquashFile) {
+		transitionImg = fmt.Sprintf("%s/cOS/%s", upgradeStateDir, constants.TransitionSquashFile)
```
davidcassany: nit: I think we could use the `filepath.Join` for that, it has little sanity checks and probably some portability features we don't care.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+	return err
+}
+
+func (u *UpgradeAction) Run() error {
+	var statePart v1.Partition
+	var err error
+	var transitionImg string
+	upgradeStateDir := constants.UpgradeStateDir
+
+	// if upgrading the recovery we mount the state in a different place as its already mounted RO, we need to remount it
```
davidcassany, line 126: Out of curiosity, have we tried to remount over the same mountpoint?

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+		statePart, err = utils.GetFullDeviceByLabel(u.Config.Runner, u.Config.StateLabel, 5)
+		if err != nil {
+			return err
+		}
+	}
+
+	u.Info("Mounting state partition %s in %s", statePart.Path, upgradeStateDir)
+	// Create the dir only if not on recovery as on recovery the dir exists already
+	if !u.Config.RecoveryUpgrade {
+		err = u.Config.Fs.MkdirAll(upgradeStateDir, os.ModeDir)
```
davidcassany, line 162: Oh, I like this `os.ModeDir`:+1: we should use it in `Elemental`.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/rancher-sandbox/elemental/pkg/elemental"
+	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
+	"github.com/rancher-sandbox/elemental/pkg/utils"
+	"github.com/spf13/afero"
+	"os"
+	"strings"
+)
+
+// UpgradeSource represents a source object to identify where it's coming from easily
+type UpgradeSource struct {
```
davidcassany, line 35: I'd merge this into the v1.Image struct, so this data is stored there. There we could change `RootTree` by a `Source` and include this data. Then the v1.Image struct could be used for nicely to run all operations operations on images (create fs img, copyfile, unpack docker, etc.)

This can be done in a followup PR though.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+type UpgradeAction struct {
+	Config *v1.RunConfig
+}
+
+func NewUpgradeAction(config *v1.RunConfig) *UpgradeAction {
+	return &UpgradeAction{Config: config}
+}
+
+func (u *UpgradeAction) Info(s string, args ...interface{}) {
+	u.Config.Logger.Infof(s, args...)
```
davidcassany, line 59: Nice :+1:

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+	}()
+
+	// create transition.img
+	img := v1.Image{
+		File:       transitionImg,
+		Label:      strings.ToUpper(upgradeTarget),
+		Size:       u.Config.ImgSize,
+		FS:         constants.LinuxImgFs,
+		MountPoint: constants.UpgradeTempDir,
+		RootTree:   upgradeSource.source, // if source is a dir it will copy from here, if it's a docker img it uses Config.DockerImg IN THAT ORDER!
```
davidcassany, line 215: See, we could have a nested struct including `Source` and `Type` instead of `RootTree` here. This would make the logic explicit  and we could easily think of a nice CopyImage method that could be reused in many places.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+	if !u.Config.RecoveryUpgrade {
+		err = ele.CreateFileSystemImage(img)
+		if err != nil {
+			u.Error("Failed to create %s img: %s", transitionImg, err)
+			return err
+		}
+
+		// mount the transition img on targetDir, so we can install the upgraded files into targetDir, and they end up on the img
+		err = ele.MountImage(&img, []string{"rw"}...)
+		defer ele.UnmountImage(&img)
```
davidcassany: note this swallows unmounting errors. I think we should consider those, there could be errors on unmounting and having some explicit failure for that I think it is really good for debugging. This can be done in a follow up though. The install action uses kind of a simple (but verbose) pattern for that.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+		u.Info("Copying %s to %s", transitionImg, destination)
+		err = utils.CopyFile(u.Config.Fs, transitionImg, destination)
+		if err != nil {
+			u.Error("Failed to copy %s to %s: %s", transitionImg, destination, err)
+			return err
+		}
+		u.Info("Finished copying %s to %s", transitionImg, destination)
+	} else {
+		if utils.BootedFrom(u.Config.Runner, u.Config.ActiveLabel) {
+			// backup current active.img to passive.img before overwriting the active.img
```
davidcassany: I don't see the tune2fs call anywhere... ? See CopyRecovery and CopyPassive from `Elemental`. I think you missed to update the filesystem labels.

davidcassany (changes requested):
The only thing that prevents me from giving it an approval is that I can't see anywhere the `tune2fs` calls to rename labels of images. I am missing something or it is a mistake?

All other comments are improvements that can be done in a second round, just not to make this PR even bigger.

pkg/action/upgrade.go
Itxaka: yes, not possible as its "on use"

pkg/action/upgrade.go
Itxaka: Good point, I just steer far away from anything that sounds like a filesystem, so I didnt even consider it :D

pkg/action/upgrade.go
Itxaka: yes, I thougth about expanding but didnt want to cascade things around, i.e. that modification maybe breaking something else. Good to me to integrate it afterwards!

pkg/action/upgrade.go
Itxaka: to be fair, do we really care in the upgrade? I mean, the next step after upgrade should be a reboot in which this will be unmounted automatically no? Plus the transition img will be autoremoved after a reboot. Its more of a "lets leave things as they were" than a "hard cleanup because we dont want to pollute the system". The magic of read only FS :laughing:

pkg/action/upgrade.go
Itxaka: wtf, I totally did

Itxaka: wait, isnt the CreateFileSystemImage doing that automatically??

davidcassany: CreateFileSystemImage creates an image with the `COS_ACTIVE` label. But old active image which will be moved to Passive should change the label from `COS_ACTIVE` to `COS_PASSIVE`.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,285 @@
...
+			source := fmt.Sprintf("%s/cOS/%s", upgradeStateDir, constants.ActiveImgFile)
+			destination := fmt.Sprintf("%s/cOS/%s", upgradeStateDir, constants.PassiveImgFile)
```
davidcassany: Here `constants.PassiveImgFile`, is file that will have a filesystem of `COS_ACTIVE` unless we relabel it. If we don't then active and passive images will use the same FS Label and passive partition will miserably fail to boot, as grub expects `COS_PASSIVE` on `/cOS/passive.img`

Itxaka: yep, found it! Thanks for the pointer, totally forgot about the backup of active to passive!

Itxaka: Actually that also brings another error into mind, as Im setting the img.Label to simply "active" or "recovery" instead of the proper labels!

pkg/action/upgrade.go
```diff
@@ -223,9 +228,16 @@ func (u *UpgradeAction) Run() error {
 				return err
 			}
 			u.Info("Finished copying %s to %s", source, destination)
+			// Label the image to passive!
+			out, err := u.Config.Runner.Run("tune2fs", "-L", u.Config.PassiveLabel, destination)
+			if err != nil {
+				u.Error("Error while labeling the passive image %s: %s", destination, err)
+				u.Debug("Error while labeling the passive image %s, command output: %s", out)
```
davidcassany: A bit unrelated to this, but since a while I am thinking of adding the logger into the runner implementation so we can eventually add a debug message for every single Runner.Run execution. This would produce quite a detailed debug output and we would not need messages like that. The only issue is having to initiate the runner with a logger and the little refactor it implies 🤷🏽‍♂️ 
 
 What you think?

pkg/action/upgrade.go
```diff
@@ -223,9 +228,16 @@ func (u *UpgradeAction) Run() error {
 				return err
 			}
 			u.Info("Finished copying %s to %s", source, destination)
+			// Label the image to passive!
+			out, err := u.Config.Runner.Run("tune2fs", "-L", u.Config.PassiveLabel, destination)
```
davidcassany: yeah! :+1:

pkg/action/upgrade.go
Itxaka: umm, that would be great and avoid this extra stuff, as usually on debug I rather value the output of the commands (and the commands run!)

pkg/action/upgrade.go
```diff
@@ -0,0 +1,297 @@
...
+		Size:       u.Config.ImgSize,
+		Label:      u.Config.ActiveLabel,
+		FS:         constants.LinuxImgFs,
+		MountPoint: constants.UpgradeTempDir,
+		RootTree:   upgradeSource.source, // if source is a dir it will copy from here, if it's a docker img it uses Config.DockerImg IN THAT ORDER!
+	}
+
+	// If on recovery, set the label to the RecoveryLabel instead
+	if utils.BootedFrom(u.Config.Runner, u.Config.RecoveryLabel) {
+		img.Label = u.Config.RecoveryLabel
```
davidcassany: This is not correct... sorry I just saw it now. This label is the recovery partition label, not the recovery image label... yes pretty confusing. It should have been `u.Config.SystemLabel`. See the logic in https://github.com/rancher-sandbox/elemental/blob/42f728f36145bc921ea6ce7d72b345ecce3e5383/pkg/elemental/elemental.go#L475-L491
When the recovery image is not squashfs it is formatted using the `RunConfig.SystemLabel` value.

Itxaka: that is indeed confusing...but yes, I can see that is the way on the cos.sh script as well....wtf

davidcassany (approved):
Looks good to me. Just this issue on the recovery image label for the non-squashfs path.

Itxaka:
Ay ay ay, real life doest not conform to the ideal world of tests...
Need ot rework the geDeviceByLabel due to marshall errors...

cmd/upgrade.go
Itxaka: This sounds good but! I really like having everything related to command X in the cmd file related to that command, like the silenceErrors stuff, that could also be moved somewhere, but for someone having a look at the cmd stuff, having all this in here makes more sense to me. Call it self contained command related stuff, but going to the config feels like an extra step to me and makes it potentially confusing to understand where things related to this command are done.

I dont mind either way, but rather keep it here.

pkg/types/v1/config.go
```diff
@@ -214,10 +224,11 @@ type RunConfig struct {
 type Partition struct {
 	Label      string
 	Size       uint
-	PLabel     string
-	FS         string
-	Flags      []string
+	PLabel     string   `json:"PARTLABEL,omitempty"`
```
davidcassany: Could it be that this makes it configurable from a config file? Note that PLabel is not configurable and uses constant values. Also there is no partition labels on MBR disks... so I'd say you will have a clash there having all the partitions with the same value if you actually feed it from some sort of disk analisys tool.

In fact, in the installer I used `PLabel` as a way to name and find partitions because it is not configurable so it is set in Partitions struct as a constant even if it is ignored for MBR.

That brings me to the point if we actually need to read the disk partitions to upgrade. Can't we simply assume the partitions are present based on the elemental current config? I mean we read the config hence we know the partitions and labels that are expected to be there, I see no need to fetch them from the disk.

Itxaka: > Could it be that this makes it configurable from a config file? Note that PLabel is not configurable and uses constant values. Also there is no partition labels on MBR disks... so I'd say you will have a clash there having all the partitions with the same value if you actually feed it from some sort of disk analisys tool.

mmmh, it should not, this is only a mapping so when you unmarshall a json into a Partition struct, it identifies the fields that correspond to this, i.e. the json key `PARTLABEL` matches the `Plabel` key and the value is unmarshalled into that field in the struct. If struct field names match the json key names, then they are unmarshalled directly into those values.


> That brings me to the point if we actually need to read the disk partitions to upgrade. Can't we simply assume the partitions are present based on the elemental current config? I mean we read the config hence we know the partitions and labels that are expected to be there, I see no need to fetch them from the disk.

To read the FS type I think. I could not get the FS type any other way, and IIRC we cant assume they are using the default FS type no? I mean... Maybe I could use "auto" everywhere?

also the paths of them? Like on a normal HDD you get sda{1....4} but what about an ssd? Or a NVMe drive? My local NVMe has /dev/nvme0n1p{1...4} which is quite different. On qemu/vbox we may get /dev/vda{1...4}, etc...

This way also makes sure that we are not just "trying" to have the defaults but actually getting the real device info from the running system, which makes it more foolproof for the future. If you get the FS/path directly from the system, there is no way that can fail...maybe :P

Itxaka: also the `Plabel` is actually not necessary as far as I can see(currently), so I could drop it. I thought that it would be good to have as much info as possible in the resulting `v1.Partition` in order to maybe reuse that method in the future when we need info from the running system partitions.

davidcassany: Right, I see your point. The original idea is that the partitions should configurable in the future, meaning the partition struct could be eventually initiated from a yaml config file as part of the elemental configuration. So, I am not sure how can we match both. 

Regarding the PLabel if we fetch it from a tool then we need to name the partitions in an additional parameter.  At a time I did not do that as having Label, PLabel and Name I thought it was too much and even confusing. Probably we can consider renaming `PLabel` to `Name` and just use the `Name` as the partition label in partitioner as a convention, so I'd leave PLabel out of the disk analysis for now.

Itxaka: well, this is only for json unmarshall, so has nothing to do with yaml unmarshall. I know, confusing as fuck. Im not sure what golang has against simple unmarshalling from yaml/json...

I will drop the plabel!

Itxaka:
Upgrades correctly.

But system running the upgrade (i.e. live system) gets corrupted, some binaries end up broken, cant restart. Force resetting boots the upgraded system correctly so it is upgrading as expected.

Grub is also not updated somehow.

So something is missing. Im rechecking the whole thing. I think that it may be due to copying over the active.img file with the transition.img, which leads to corruption in the current system? But that is the same as its done with the old installer so not sure why.

No more ideas to be honest. Unless I set a broken path somewhere and its mounting parts of the current FS somehow, I dont understand what is going on. AFAIK the active.img is mounted as a loop on /dev/loop0 which is mounted RO at /
If there is corruption due to copying the transition.img over the active.img...it should also occur with the normal script...

Im really puzzled. @davidcassany @mudler

Itxaka:
well, checking in older versions of the installer, when its was separated, the workflow seems a bit different....

https://github.com/rancher-sandbox/cOS-toolkit/blob/c8d6ad5f7f7f9e3eb9d3e552982e0ce507ef6fef/packages/installer/upgrade.sh#L12
```
# 1. Identify active/passive partition
# 2. Install upgrade in passive partition
# 3. Invert partition labels
```

Itxaka:
cant believe this shit.

So yeah old script was taking advantage of using `mv` as lostup shows, you can freely move the img file mounted and it will update it no problem. But of course that doesnt work if we copy using golang, we just override shit.


```
cos:/run/initramfs/cos-state/cOS # losetup
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE                               DIO LOG-SEC
/dev/loop0         0      0         0  1 /run/initramfs/cos-state/cOS/active.img   0     512
cos:/run/initramfs/cos-state/cOS # mv active.img passive.img 
cos:/run/initramfs/cos-state/cOS # losetup
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE                                DIO LOG-SEC
/dev/loop0         0      0         0  1 /run/initramfs/cos-state/cOS/passive.img   0     512
```

pkg/constants/constants.go
```diff
@@ -68,8 +69,28 @@ const (
...
+	TransitionImgFile      = "transition.img"
+	TransitionSquashFile   = "transition.squashfs"
 )
 
 func GetCloudInitPaths() []string {
 	return []string{"/system/oem", "/oem/", "/usr/local/cloud-config/"}
 }
+
+// GetDefaultSquashfsOptions returns the default options to use when creating a squashfs
+func GetDefaultSquashfsOptions() []string {
```
davidcassany, line 87: Nice!

pkg/types/v1/config.go
```diff
@@ -159,39 +159,49 @@ func NewRunConfig(opts ...RunConfigOptions) *RunConfig {
 	if r.GrubDefEntry == "" {
 		r.GrubDefEntry = cnst.GrubDefEntry
 	}
+
+	if r.ImgSize == 0 {
```
davidcassany, line 163: Today I was thinking if it would make sense to add a config test for default values, it feels quite stupid, but this kind of details are easy to forget. We could also define defaults with viper.SetDefault but I am not sure how to use it and if there is much gain. Ideally, would be nice to code the default as part of the struct `yaml` decorators but I could not find documentation about it 🤷🏽‍♂️

pkg/utils/common.go
```diff
@@ -160,3 +191,20 @@ func CosignVerify(fs afero.Fs, runner v1.Runner, image string, publicKey string,
 	out, err := runner.Run("cosign", args...)
 	return string(out), err
 }
+
+// CreateSquashFS creates a squash file at destination from a source, with options
+// TODO: Check validity of source maybe?
```
davidcassany, line 196: I don't think so, I would assume these kind of methods must get sanitized values. If not it should be a bug somewhere else.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,435 @@
...
+func (u *UpgradeAction) Debug(s string, args ...interface{}) {
+	u.Config.Logger.Debugf(s, args...)
+}
+
+func (u *UpgradeAction) Error(s string, args ...interface{}) {
+	u.Config.Logger.Errorf(s, args...)
+}
+
+func (u *UpgradeAction) upgradeHook(hook string, chroot bool, rebrand bool) (err error) {
+	if chroot { // TODO: This is not really doing anything for hooks as it closes the chroot before running the hooks?
```
davidcassany, line 71: It sets the chroot here and it is closed right before returning upgradeHook method. The `utils.RunStage(hook, u.Config)` call will run in the chroot env if this is set to true.

pkg/action/upgrade.go
```diff
@@ -0,0 +1,435 @@
...
+					u.Error("Failed setting grub variables: %v", out)
+					if u.Config.Strict {
+						return err
+					}
+				}
+			}
+		}
+	}
+	u.Info("Running %s hook", hook)
+	// Sorry, but I dont want to see all the yip stuff
```
davidcassany, line 103: Yes, yip messages tend to be quite annoying here.

davidcassany (approved):
Looks good to me

pkg/action/upgrade.go
Itxaka: sure about that? 'Cause I was testing this and it wasnt running on the chroot as the actual chdir+chroot only happens *inside* the Run function :=)

pkg/types/v1/config.go
Itxaka: Yes, agree. We should probably have at least checks for non-empty values for things that should always be there. I.E. ImgSize is necessary but the running system may not have it, etc...

Take into account that Viper.Setdefaults is only for CMD commands, while this struct can be used from anywhere. So while I would like to also have checks and defaults in there, I think having them in this struct is much more important, as it can be used from other parts of our code not related to CMD

pkg/utils/common.go
Itxaka: ummm, "assume" is the keyword here. If you pass a non existing value you would get a panic in runtime right? Not good. Maybe Im just used to code defensively due to python :D

pkg/action/upgrade.go
Itxaka: yes, I think we should silence all by default and have a global flag in elemental to show cloud-init messages.

pkg/action/upgrade.go
davidcassany: gonna re-check it, I think I did a time but I could be wrong...

PR #60, opened 2022-01-27
Ensure RunConfig parameters are writable by Viper

davidcassany:


davidcassany:
At a time I was surprised that Viper/Cobra worked while making few elements private and trying set them from yaml or params or env vars. Now I am surprised it did not complaint at all if it was not working 😅

Itxaka (approved):

PR #61, opened 2022-01-28
Fixes grub tests flakyness

davidcassany:

### DIFF
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -359,7 +359,7 @@ var _ = Describe("Utils", func() {
 				Expect(grub.SetPersistentVariables(
 					"somefile", map[string]string{"key1": "value1", "key2": "value2"},
 				)).To(BeNil())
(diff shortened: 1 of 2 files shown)


pkg/utils/utils_test.go
```diff
@@ -370,7 +370,7 @@ var _ = Describe("Utils", func() {
 				config.Runner = runner
 				grub := utils.NewGrub(config)
 				Expect(grub.SetPersistentVariables(
-					"somefile", map[string]string{"key1": "value1", "key2": "value2"},
+					"somefile", map[string]string{"key1": "value1"},
```
Itxaka, line 373: cant we have a tests also with both values to check if it works with n>1 values?

Itxaka (approved):

pkg/utils/utils_test.go
davidcassany: This is a failure test, on the first execution will fail. But we can't be sure which will be the first execution if we request many.

Note the previous non failure test includes n>1

PR #64, opened 2022-01-28
Hide yip errors under warning

Itxaka:
Instead of showing them, just hide them under debug and notify that
there were ignored errors and how to see them/fail on them

Related https://github.com/rancher-sandbox/cOS-toolkit/issues/1109

### DIFF
--- a/pkg/utils/runstage.go
+++ b/pkg/utils/runstage.go
@@ -108,8 +108,8 @@ func RunStage(stage string, cfg *v1.RunConfig) error {
 	}
 
 	if errors != nil && !cfg.Strict {
-		cfg.Logger.Info("Some errors found but were ignored:")
-		cfg.Logger.Info(errors)
+		cfg.Logger.Info("Some errors found but were ignored. Enable --strict mode to fail on those or --debug to see them in the log")
+		cfg.Logger.Warn(errors)
 		return nil
 	}
 


pkg/utils/runstage.go
```diff
@@ -108,8 +108,8 @@ func RunStage(stage string, cfg *v1.RunConfig) error {
 	}
 
 	if errors != nil && !cfg.Strict {
-		cfg.Logger.Info("Some errors found but were ignored:")
-		cfg.Logger.Info(errors)
+		cfg.Logger.Info("Some errors found but were ignored. Enable --strict mode to fail on those or --debug to see them in the log")
+		cfg.Logger.Debug(errors)
```
mudler: That would leave genuine failures in the cold, and that becomes tricky to handle as if you want to collect debugs from the various systemd unit is gonna be messy (especially if you think we have some of them in the initrd) unless we have sentinel files for debug, or cmdline args

my 2c here is that we should maybe start to identify errors if we want to treat them differently, or otherwise I would actually move the two above to the Warning level instead, to differentiate from INFO, but not hiding it under the debug shelf. In this way if someone wants to silence those they can level up the log level

davidcassany: Yes I was also thinking on warning messages rather than debug. We can simply hide warning messages but still give the impression something unexpected happened.

Also, what about throwing a warning if the folder is not there and skip it from CloudInitPaths list? For this particular case this could be a good solution. Then for strict run we could eventually error out or simply by pass it to yip. I know this is only for this specific use case, but IMHO it makes sense, since the search paths management belongs to runStage not yip.

Itxaka: > Also, what about throwing a warning if the folder is not there and skip it from CloudInitPaths list? For this particular case this could be a good solution. Then for strict run we could eventually error out or simply by pass it to yip. I know this is only for this specific use case, but IMHO it makes sense, since the search paths management belongs to runStage not yip.

Well...The problem is that the folder can be created *during* the yip run, i.e. a cloud-config file creates the folder, and then yip runs over that. So all the folders existence is kind of dynamic, so its impossible to check beforehand.

Im not sure what we win by setting them to warning instead of debug? The whole problem steams from cos-setup being very lenient with errors due to the dynamic nature of those. If warnings appear, they will give an impression of failure if there is not, like now, and if log level is upped (we dont have that yet BTW) and hidden by default, then its the same as having them on debug??

IMO, substantial changes would be needed in yip to identify the source maybe better that currently is (parsing a string and trying to understand what it is, as the problem stems from a non-existing dir, which yip fails to parse as dir (of course, it doesnt exists) and then it proceeds to try to load it directly, as if it was a dotnotation value. If we could identify a dir VS a real dotnotation string, this problem would be fixed AND we could be more strict about errors.

davidcassany: Debug vs Warning is just a cosmetic matter, I prefer Warning, but this is a minor thing. Yes, I see, I did not consider the option of yip creating and feeding new paths for the this same stage... I consider it insane (as I would expect one stage to impact on other but not self), in any case I agree it is good to leave it at users choice.

So I believe this is  a two stage matter, first we can apply some logic here to move error messages from yip to some other level than INFO (my preference for warning, but debug is also fine for me). Second we need to improve yip's error management. Probably we need to distinguish few error types and handle them differently:

* syntax errors
* non existing path/yaml
* error on plugin execution

Then about parsing do notation probably we need some identifier to pass the appropriate argument to yip instead of passing the whole kernel command line, which, theoretically, could even clash with non related arguments. Then we could treat errors on dot notation parsing a simple syntax errors. Alternatively we can assume a different error management for yamls than for dot notation, however I don't know if this would be easy to code in yip.

davidcassany:
@Itxaka what about if we simply hide it for now as warning and remove "Fixes rancher-sandbox/cOS-toolkit#1109" from the PR so we keep around the card to apply required changes in yip?

davidcassany (approved):

PR #65, opened 2022-01-31
Do not use cosign luet plugin

### LINKED ISSUE #1106: Do not set luet plugins in elemental
This card relates to https://github.com/rancher-sandbox/cOS-toolkit/pull/1074#issuecomment-1024031382 

The idea is to not set luet plugins within luet but simply use them as we need outside the luet scope. This essentially applies to mtree and cosign. This gives us further control on how to call mtree and cosign utilities and provide a better, cleaner and more understandable UI.

This card also encompassed new elemental client flags to properly set and handle `cosign`.

davidcassany:
This commit instead of using luet's cosign plugin it just calls
cosign utility as needed. It facilitates a better use of environmental
variables and cosign specific flags.

Fixes rancher-sandbox/cOS-toolkit#1106

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -70,6 +70,7 @@ func init() {
 	installCmd.Flags().StringP("partition-layout", "p", "", "Partitioning layout file")
 	installCmd.Flags().BoolP("no-verify", "", false, "Disable mtree checksum verification (requires images manifests generated with mtree separately)")
 	installCmd.Flags().BoolP("cosign", "", false, "Enable cosign verification (requires images with signatures)")
+	installCmd.Flags().BoolP("cosign-key", "", false, "Sets the URL of the public key to be used by cosign validation")
 	installCmd.Flags().BoolP("no-format", "", false, "Don’t format disks. It is implied that COS_STATE, COS_RECOVERY, COS_PERSISTENT, COS_OEM are already existing")
 	installCmd.Flags().BoolP("force-efi", "", false, "Forces an EFI installation")
 	installCmd.Flags().BoolP("force-gpt", "", false, "Forces a GPT partition table")
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -70,7 +70,7 @@ func init() {
 	installCmd.Flags().StringP("partition-layout", "p", "", "Partitioning layout file")
 	installCmd.Flags().BoolP("no-verify", "", false, "Disable mtree checksum verification (requires images manifests generated with mtree separately)")
 	installCmd.Flags().BoolP("cosign", "", false, "Enable cosign verification (requires images with signatures)")
-	installCmd.Flags().BoolP("cosign-key", "", false, "Sets the URL of the public key to be used by cosign validation")
+	installCmd.Flags().StringP("cosign-key", "", "", "Sets the URL of the public key to be used by cosign validation")
 	installCmd.Flags().BoolP("no-format", "", false, "Don’t format disks. It is implied that COS_STATE, COS_RECOVERY, COS_PERSISTENT, COS_OEM are already existing")
 	installCmd.Flags().BoolP("force-efi", "", false, "Forces an EFI installation")
 	installCmd.Flags().BoolP("force-gpt", "", false, "Forces a GPT partition table")
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/zloylos/grsync"
 	"io"
+	"os"
 	"os/exec"
 	"strings"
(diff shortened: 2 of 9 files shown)


cmd/install.go
```diff
@@ -70,6 +70,7 @@ func init() {
 	installCmd.Flags().StringP("partition-layout", "p", "", "Partitioning layout file")
 	installCmd.Flags().BoolP("no-verify", "", false, "Disable mtree checksum verification (requires images manifests generated with mtree separately)")
 	installCmd.Flags().BoolP("cosign", "", false, "Enable cosign verification (requires images with signatures)")
+	installCmd.Flags().BoolP("cosign-key", "", false, "Sets the URL of the public key to be used by cosign validation")
```
Itxaka: Bool? Is this correct?

davidcassany:
Note this is only removing the use of luet-cosign plugin. For luet mtree plugin I believe calling it within luet makes complete sense as there is some luet specific logic around luet repositories to fetch metadata and so on.

cmd/install.go
davidcassany: Ouch... 😓 This is certainly not correct

davidcassany:
I have few doubts from UI perspective. The client, with this PR, supports `--cosign` and `--cosign-key` falgs. If only `--cosign` is used it runs a keyless validation. For a validation using a public key it requires `--cosign --cosign-key=<key>`. I am wondering if it would make more sense something like `--cosign=<key>` to use cosign with public key and `--cosign` for keyless validation, so just reuse the same flag.

I think this could be cleaner from a user POV, what you think?
This way we would having to think/code what happens if only `--cosign-key` is provided.

davidcassany:
> I am wondering if it would make more sense something like `--cosign=<key>` to use cosign with public key and `--cosign` for keyless validation, so just reuse the same flag.

Ok this is not possible in Cobra. So I am leaving it as is. For the time being.

cmd/install.go
```diff
@@ -70,6 +70,7 @@ func init() {
 	installCmd.Flags().StringP("partition-layout", "p", "", "Partitioning layout file")
 	installCmd.Flags().BoolP("no-verify", "", false, "Disable mtree checksum verification (requires images manifests generated with mtree separately)")
 	installCmd.Flags().BoolP("cosign", "", false, "Enable cosign verification (requires images with signatures)")
+	installCmd.Flags().StringP("cosign-key", "", "", "Sets the URL of the public key to be used by cosign validation")
```
Itxaka, line 73: I wonder if there is a way to link flags to subflags or something like that so we can group them.

Itxaka: also URL OR file.

pkg/utils/common.go
```diff
@@ -116,3 +117,27 @@ func SyncData(source string, target string, excludes ...string) error {
...
+
+func CosignVerify(fs afero.Fs, runner v1.Runner, image string, publicKey string, debug bool) (string, error) {
+	args := []string{}
+
+	if debug {
+		args = append(args, "-d=true")
+	}
+	if publicKey != "" {
+		args = append(args, "-key", publicKey)
+	}
```
Itxaka, line 132: sorry, we need to set the keyless verification which is only possible via....env var :D https://github.com/sigstore/cosign/blob/main/KEYLESS.md

So ...

```
if publicKey != "" {
	args = append(args, "-key", publicKey)
} else {
	os.Setenv("COSIGN_EXPERIMENTAL", 1)
}
```

Itxaka (changes requested):

Itxaka (commented):
The idea would alos be to integrate the cosign stuff into the "new" config, so we only need to read from there instead of having to pass all the arguments.

Cant be done separately thought.

Itxaka:
> Ok this is not possible in Cobra. So I am leaving it as is. For the time being.

Probably doable by creating our own flag type I guess.

davidcassany:
> > Ok this is not possible in Cobra. So I am leaving it as is. For the time being.
> 
> Probably doable by creating our own flag type I guess.

The only option that I thought about is setting a known default something like `_none_` so if `RunConfig.Cosign` is set to `_none_` we can assume `--cosign` flag is not used. Then to set an empty key something like `elemental install --cosign= /dev/sda` must be used (not the `=` sign). So that I thought about keeping both flags separated as it the code for that is clear and obvious with the only downside of having to use two flags in order to make use of cosign, good thing is that `cosign: true` can be set in a config file, then the user just needs to use `--cosing-key`.

pkg/utils/common.go
davidcassany: Ouch damn it. Another round is coming

Itxaka (approved):

PR #72, opened 2022-02-03
Install config refactor

davidcassany:
With this refactor the install command can be expressed in three simple steps:
    
1. Basic checks on flags and config as part of 'cmd/install.go'
2. Setup install details from a rough RunConfig initiated and run all logic to determine partitions, images, etc. according to the install action specifics.
3. Run installation after installation setup is done

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cmd
 
 import (
+	"errors"
 	"github.com/rancher-sandbox/elemental/cmd/config"
 	"github.com/rancher-sandbox/elemental/pkg/action"
 	"github.com/spf13/cobra"
@@ -41,15 +42,30 @@ var installCmd = &cobra.Command{
 		mounter := mount.New(path)
 
 		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), mounter)
-
 		if err != nil {
 			cfg.Logger.Errorf("Error reading config: %s\n", err)
 		}
+
+		err = errors.New("Invalid options")
+		if viper.GetBool("reboot") && viper.GetBool("poweroff") {
+			cfg.Logger.Errorf("'reboot' and 'poweroff' are mutually exclusive options")
+			return err
+		}
+
+		if viper.GetString("cosign-key") != "" && !viper.GetBool("cosign") {
+			cfg.Logger.Errorf("'cosign-key' requires 'cosing' option to be enabled")
+			return err
+		}
+
+		if viper.GetBool("cosign") && viper.GetString("cosign-key") == "" {
+			cfg.Logger.Warnf("No 'cosign-key' option set, keyless cosign verification is experimental")
+		}
+
 		// Should probably load whatever env vars we want to overload here and merge them into the viper configs
 		// Note that vars with ELEMENTAL in front and that match entries in the config (only one level deep) are overwritten automatically
 		cfg.Target = args[0]
 
-		err = cfg.DigestSetup()
+		err = action.InstallSetup(cfg)
 		if err != nil {
 			return err
 		}
@@ -57,9 +73,7 @@ var installCmd = &cobra.Command{
 
 		cfg.Logger.Infof("Install called")
 
-		// Dont call it yet, not ready
-		install := action.NewInstallAction(cfg)
-		err = install.Run()
+		err = action.InstallRun(cfg)
 		if err != nil {
 			return err
 		}
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -69,7 +69,6 @@ var installCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(installCmd)
-	installCmd.Flags().StringP("config-dir", "e", "/etc/elemental/", "dir where the elemental config resides")
 	installCmd.Flags().StringP("docker-image", "d", "", "Install a specified container image")
 	installCmd.Flags().StringP("cloud-init", "c", "", "Cloud-init config file")
 	installCmd.Flags().StringP("iso", "i", "", "Performs an installation from the ISO url")
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -54,7 +54,7 @@ var upgradeCmd = &cobra.Command{
 		}
 
 		// Init setup
-		err = cfg.DigestSetup()
+		err = action.InstallSetup(cfg)
 		if err != nil {
 			cfg.Logger.Errorf("Error digesting setup: %s\n", err)
 		}
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -54,6 +54,7 @@ var upgradeCmd = &cobra.Command{
 		}
 
 		// Init setup
+		// TBC action.InstallSetup is the exact equivalent of the old DigestSetup
 		err = action.InstallSetup(cfg)
 		if err != nil {
 			cfg.Logger.Errorf("Error digesting setup: %s\n", err)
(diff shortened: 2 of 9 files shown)


pkg/types/v1/config.go
```diff
@@ -243,138 +232,6 @@ func (pl PartitionList) GetByPLabel(label string) *Partition {
...
-
-	if r.Cosign && r.CosignPubKey == "" {
-		r.Logger.Warnf("No 'cosign-key' option set, keyless cosign verification is experimental")
-	}
-
-	return nil
-}
-
-// setupLuet will initialize Luet interface if required
-func (r *RunConfig) setupLuet() {
```
Itxaka, line 368: This blows cause Im using this in the upgrade part to setup a luet for the config. Now I either implement what Im missing from here on the upgrade (and in future stuff) or call the InstallAction directly?

davidcassany:
@Itxaka I am doing this in the scope in of coding the `reset` command, I realized it shares quite some logic with the `install` command however some little differences on configuration are needed. 
At first started moving from `config.DigestSetup` to `config.InstallSetup` but while working on the `ResetSetup` method I realized that would cause cycling dependencies (they import each other) between `v1` package and `utils`. Then I noted the the `InstallSetup` is coupled with the install action so I moved it there.

I added the `Do not merge` label because I expect to rebase on top of the first PR of the upgrade command. Even I believe there is no overlap, but I don't want to add another rock block in the upgrade command path.

Itxaka (changes requested):
This doesn't seem good to me. First is, having the install Action directly means that is one unmovable object. You either run the install or not, like it doesn't allow you to have an object that you can modify, play or change, its just install or nothing. Seems very constrained to me, you can no longer get the InstallAction and inspect it, modify or do anything with it.

Second is the removal of the setup, which is kind of bad, as all the setup done in there can be used in other commands, i.e. Im using it for the upgrade. Should I duplicate the code somewhere else? Why move all that generic code that setups a lot of important data in the config into an exclusive setup for install?

davidcassany:
> Why move all that generic code that setups a lot of important data in the config into an exclusive setup for install?

The problem I found it is that it is not generic. Active image differs and partitions setup differs for the reset command vs install command. Yes I agree that quite some logic of the setup could be reused, but if so I'd suggest to move into a common action method. The problem I found in `reset` command is that to set the partitions and ActiveImage I needed to run `BootedFrom` checks, this is part of `utils` package and this can't be imported in `v1`, it causes a cyclic dependency.

pkg/types/v1/config.go
davidcassany: Well, that's why I want to rebase this PR once the upgrade command is there, to feed the generic stuff in a `pkg/action/common.go` utility.

Itxaka:
blergh, I hate circular dependencies...

davidcassany:
> This doesn't seem good to me. First is, having the install Action directly means that is one unmovable object. You either run the install or not, like it doesn't allow you to have an object that you can modify, play or change, its just install or nothing

Well I did it this way as I saw no gain on having it as an structure since all the details are stored in `RunConfig`, everything happens there and to avoid having to create a new object just to bypass the `RunConfig` to the core `elemental` package. Anyway, I do not have a strong opinion in that detail, I'd treat actions as mere high level scripts that operate on a given config by using the core API. But I am also fine keeping them as objects.

Itxaka:
cant we maybe move the bootedFrom to a different place to avoid circular deps? Like it may be good to have it as part of elemental? I have a feeling we will hit this again in the future otherwise :(

davidcassany:
> cant we maybe move the bootedFrom to a different place to avoid circular deps? Like it may be good to have it as part of elemental? I have a feeling we will hit this again in the future otherwise :(

I don't think this would solve the problem. Elemental also imports `v1`. IMHO the problem is having runtime logic in `v1`. So moving `BootedFrom` kind of logic out of the the `pkg/types/v1`. I though that `./pkg/action` could be a good place for that mostly because this sort of configuration checks and analysis are coupled with the actual executed command.

Another option that comes to my mind is splitting configuration struct in two entities, one `Config` in `v1` mostly to include all flags from files or client UI, and then with this produce a new `RunConfig` entity that is part of another package. This sounds like a deeper refactor though.

What I'd really like to keep is the approach of 3 steps actions: basic flags checks, define configuration and execute it. If we follow this approach we will be capable to reuse quite a lots of code in `install`, `reset` and `ugprade` run method.

cmd/upgrade.go
```diff
@@ -54,7 +54,7 @@ var upgradeCmd = &cobra.Command{
 		}
 
 		// Init setup
-		err = cfg.DigestSetup()
+		err = action.InstallSetup(cfg)
```
davidcassany, line 58: @Itxaka, my idea here is that we can run action specific initialization including checks like from which system are we booting, the DigestSetup was coupled with installation logic. This way we can simply set `ActiveImage` and partitions correct as we need from the very begining. For instance we could initialize partitions we need using the `GetFullDeviceByLabel` and have sanitized data from the very beginning before applying any kind of change.

pkg/types/v1/config.go
```diff
@@ -189,7 +192,7 @@ type RunConfig struct {
 	Strict           bool   `yaml:"strict,omitempty" mapstructure:"strict"`
 	Iso              string `yaml:"iso,omitempty" mapstructure:"iso"`
 	DockerImg        string `yaml:"docker-image,omitempty" mapstructure:"docker-image"`
-	Cosign           bool   `yaml:"no-cosign,omitempty" mapstructure:"no-cosign"`
+	Cosign           bool   `yaml:"cosign,omitempty" mapstructure:"cosign"`
```
davidcassany, line 195: @Itxaka I think we should use `--cosign` flag instead of `--no-cosign` and assume that by default cosign is disabled unless the user includes `--cosign` flag. At least this is how we finally did it for the installer, I think is simpler to understand form a user PoV.

cmd/install.go
```diff
@@ -41,25 +42,38 @@ var installCmd = &cobra.Command{
 		mounter := mount.New(path)
 
 		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), mounter)
-
 		if err != nil {
 			cfg.Logger.Errorf("Error reading config: %s\n", err)
 		}
+
+		err = errors.New("Invalid options")
```
Itxaka, line 49: we should probably extract his and some of the flag setting into a cmd/flags.go file so it can be reused, i.e. this will be the same in upgrade

cmd/upgrade.go
Itxaka: Then why can it be called `ActionSetup`? Maybe accept a `action string` so we can even branch inside the setup?

pkg/action/common.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	defer func() {
+		if tmpErr := chroot.Close(); tmpErr != nil && err == nil {
+			err = tmpErr
+		}
+	}()
+	return actionHook(config, hook)
+}
+
+func setupLuet(config *v1.RunConfig) {
```
Itxaka, line 51: Im not sure setupLuet should be here as its tightly coupled with the config more than the action no? Like we dont need to have an "action" in order to setupLuet on a config object (i.e. pull image?)

pkg/types/v1/config.go
Itxaka: mmmh, I thought it was this way already? Didnt we discuss that?? Any way, good for me disabled by default, enabled by flag :+1:

Itxaka (approved):
some nits, feel free to ignore them

pkg/action/common.go
davidcassany: Yes I don't know... my first idea was to keep all configuration logic into v1.RunConfig, however I faced cycling dependency issues. Thats also why I also thought about making action methods stateless, so they can be used as high level utilities and there is no need to initate any object to run `action.setupLuet(config)` anywhere is just fine to initiate Luet.

I'd be happy to have this logic in an other package, I am just bad at thinking names, I don't want to make a another crap as `PLabel` 😅

cmd/upgrade.go
davidcassany: Well I would create an upgrade specific method to actually set the ActiveImage (as an example) to meaningful values for the Upgrade process. Many RunConfig default values were set only having installation in mind. I think it makes sense loading them from the disk as you attempted first on upgrade, then we can be sure to have a sane partition list and probably there is no need to have all possible partitions (aka BIOS and/or EFI), only the ones that you may need to mount or remount.

I am fine with the idea of having a general ActionSetup passing the action string.

pkg/types/v1/config.go
davidcassany: Yes it was this way, but this was changed on the upgrade command merge. I did not see it during the review...

Itxaka: argh, sorry about that!

PR #73, opened 2022-02-03
Move to ginkgo v2

### LINKED ISSUE #68: Move to ginkgov2
v2 has been released and it brings some cool things like labels.

### Requirements

 - [ ] move to ginkgo v2
 - [ ] make sure test suite still passes

### LINKED ISSUE #69: Split tests into root/non-root
Some tests naturally require root privs to pass but we dont want to run the test suite always under root as it hinders local development and some test may even be destructive if configured wrong.

Once we have done #68 it would be possible to mark those tests with labels to make sure they are not run by default and we cna run a root suite and a non-root suite.

### Requirements
 - [x] Mark tests that require root with root labels
 - [x] Make sure the default `make test` target runs only the non-root suite
 - [x] Have an extra target to run the root suite
 - [x] Have the CI run both suites
 - [x] Merge coverage from both suites so we get full coverage

### LINKED ISSUE #71: Use ginkgo runner for tests
Instead of using go test, we should use the ginkgo runner which looks much nicer and allows us to set more options directly like labels.

### Requirements

 - [x] Use ginkgo runner in the `make test` target

Itxaka:
- Drop deprecated slow treshold by using the reporter. Is now settable
by cli flag only
 - No other changes

Fixes #68 


mudler (approved):

davidcassany (approved):
Nice

PR #74, opened 2022-02-04
Rename 'PLabel' form v1.Partition to 'Name'

davidcassany:


davidcassany:
Lets admit it, having this `v1.Partition.PLabel` and `v1.Partition.Label` attributes was not a good idea of mine. `PLabel` is suposed to stand for the partition label and, at least in installer code, it is used only to feed `parted` on GPT (there is no such a think for MBR).

Using `Name` and assume this is the static name (not configurable in any way) can be useful to identify partitions regardless of any crazy label the user decides to use. This meant to be only used internally.

Itxaka (approved):

PR #76, opened 2022-02-04
Add cache for go modules and build to CI

Itxaka:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,5 +14,13 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: 1.16.x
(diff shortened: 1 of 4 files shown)


davidcassany (approved):
Looks good. 

Is this just to speed up tests or something else I miss?

Itxaka:
Basically should shave about a minute and a half from the tests and also protect us from network issues, i.e. cant download some modules.

takes about 11s to download the cache and extract it but tests take a couple of minutes less. Plus both test runs (roo/nonroot) use the same cache instead of downloading twice.

EDIT: I could not make both tests use the same cache so root tests still downloads stuff :cry: Maybe the next step is vendoring...

PR #78, opened 2022-02-04
Add local flag to unpack local image

Ludea:
Add local flag to unpack local image instead of pulling it


cmd/pull-image.go
```diff
@@ -89,5 +91,6 @@ func init() {
 	pullImage.Flags().String("auth-identity-token", "", "Authentication identity token")
 	pullImage.Flags().String("auth-registry-token", "", "Authentication registry token")
 	pullImage.Flags().Bool("verify", false, "Verify signed images to notary before to pull")
+        pullImage.Flags().Bool("local", false, "Use local imagz")
```
davidcassany: I guess this is a typo

pkg/types/v1/luet.go
```diff
@@ -51,7 +51,7 @@ func NewLuet(log Logger, context *context.Context, auth *dockTypes.AuthConfig, p
 	}
 }
 
-func (l Luet) Unpack(target string, image string) error {
+func (l Luet) Unpack(target string, image string, local bool) error {
```
davidcassany, line 54: Shouldn't `local`be used to run a local unpack of the image?

davidcassany (commented):
@Ludea thanks much for your contribution, this is appreciated!

A couple of comments:
* Is the PR completed? To me it looks like it misses the actual local unpack from Luet API
* I am wondering if it the extra `--local` flag is needed, probably we can do something like try to state the given image as it was a local path and if it fails atempt to pull from a registry.

cmd/pull-image.go
Ludea: Yes

Ludea:
> * Is the PR completed? To me it looks like it misses the actual local unpack from Luet API

Yes, you're right, I miss https://github.com/Ludea/elemental/blob/main/pkg/types/v1/luet.go#L56

> * I am wondering if it the extra `--local` flag is needed, probably we can do something like try to state the given image as it was a local path and if it fails atempt to pull from a registry.

It's an another way to do job. 
There is already an api with luet, with flag local, so we can use it. 
As you want

Ludea:
Implemented Luet api !

Itxaka:
Looks good to me. Its missing a test to see if the local path is run properly thought.

AFAIK this means that ti doesn't go to any remote but tries to get the image from the local docker daemon right? That is really helpful :D

Ludea:
> AFAIK this means that ti doesn't go to any remote but tries to get the image from the local docker daemon right? That is really helpful :D

Exactly 😊

pkg/types/v1/luet_test.go
```diff
@@ -42,7 +44,15 @@ var _ = Describe("Types", Label("luet", "types"), func() {
...
 			image := "quay.io/costoolkit/releases-green:cloud-config-system-0.11-1"
-			Expect(luet.Unpack(target, image)).NotTo(BeNil())
+			Expect(luet.Unpack(target, image, false)).NotTo(BeNil())
+		})
+	})
+        Describe("Luet", func() {
+		It("Unpack local images", Label("unpack"), func() {      
+			image := "quay.io/costoolkit/releases-green:cloud-config-system-0.11-1"
+			ref, err := name.ParseReference(image) 
+                        img, err := remote.Image(ref, remote.WithAuth(staticAuth{auth}))
```
Itxaka: `staticAuth` is not defined anywhere

Itxaka:
```suggestion
			basicAuth := authn.Basic{}
                        img, err := remote.Image(ref, remote.WithAuth(&basicAuth)) 
```

Itxaka: This would require to `import 	"github.com/google/go-containerregistry/pkg/authn"`

Itxaka:
Im wondering if we could reuse the already used `docker` dependency for the test. Like we only need to pull an image into the local docker registry in order to test this, so bringing an extra dep into this seems overboard.

pkg/types/v1/luet_test.go
Itxaka: disregard this, I got it working with existing deps, so no need to import those new dependencies

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/ProtonMail/go-crypto v0.0.0-20220113124808-70ae35bab23f // indirect
 	github.com/docker/docker v20.10.12+incompatible
 	github.com/docker/go-units v0.4.0
+	github.com/google/go-containerregistry v0.7.0
```
Itxaka: this can be dropped as its not needed anymore! Im guessing running `go mod tidy` should auto remove it?

Itxaka (changes requested):
elemental bin was mistakenly added :D

pkg/types/v1/luet_test.go
```diff
@@ -19,6 +19,7 @@ package v1_test
 import (
 	dockTypes "github.com/docker/docker/api/types"
 	context2 "github.com/mudler/luet/pkg/api/core/context"
+	dockClient "github.com/docker/docker/client"
 	. "github.com/onsi/ginkgo/v2"
```
Itxaka, line 24: missing importing `"context"` here

PR #81, opened 2022-02-04
upgrade from channel

### LINKED ISSUE #75: Add upgrade from channel release
Currently the upgrade is missing the default upgrade path which is the channel upgrade. Pieces are in there for it to be picked up but the path on CopyActive is not done.

### Requirements

- [x] CopyActive supports channel-upgrades
- [x] CopyActive is aware of the target source and where it comes from (dockerImg, directory, channel-upgrades)
- [x] Install action is not affected by the changes

Itxaka:


cmd/pull-image.go
```diff
@@ -68,7 +66,7 @@ var pullImage = &cobra.Command{
 			RegistryToken: registryToken,
 		}
 
-		luet := v1.NewLuet(cfg.Logger, context, auth, plugins...)
+		luet := v1.NewLuet(v1.WithLuetLogger(cfg.Logger), v1.WithLuetAuth(auth), v1.WithLuetPlugins(plugins))
```
mudler: :heart:

pkg/types/v1/common_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
+)
+
+var _ = Describe("Types", Label("types", "common"), func() {
+	FDescribe("Common", func() {
+		It("Can create it?? Like what is expected to be test of a simple struct?", func() {
```
mudler, line 27: if the type doesn't have attached any logic to it no need for a test file at all, no? we can add in case, for coverage the struct should be hit by the other code paths..

pkg/types/v1/luet.go
```diff
@@ -35,20 +43,70 @@ type Luet struct {
...
 
-func NewLuet(log Logger, context *context.Context, auth *dockTypes.AuthConfig, plugins ...string) *Luet {
-	bus.Manager.Initialize(context, plugins...)
-	if len(bus.Manager.Plugins) != 0 {
-		log.Infof("Enabled plugins:")
-		for _, p := range bus.Manager.Plugins {
-			log.Infof("* %s (at %s)", p.Name, p.Executable)
+type LuetOptions func(l *Luet) error
+
+func WithLuetPlugins(plugins []string) func(r *Luet) error {
```
mudler: small style nit, if you make this
```suggestion
func WithLuetPlugins(plugins ...string) func(r *Luet) error {
```
makes it easier to the caller, but that's just about taste and consistency..

pkg/types/v1/common_test.go
Itxaka: I dont know with coverage anymore. I saw a simple reformatting that increased coverage...I saw codecov complaining that the struct was not tested which I found impossible...

mudler (approved):
just small nits, but looks good here overall!

pkg/types/v1/luet.go
Itxaka: mmh, thats a good point, but the only time we use it we already got a []string .... but indeed, for manual calling it should be better that way!

pkg/utils/common.go
```diff
@@ -208,3 +209,22 @@ func CreateSquashFS(runner v1.Runner, logger v1.Logger, source string, destinati
 	}
 	return nil
 }
+
+// LoadOsRelease will try to parse the /etc/os-release file and return a map with the kye/values
+func LoadOsRelease(fs afero.Fs) (map[string]string, error) {
+	var osReleaseMap map[string]string
+	var err error
+
+	f, err := fs.Open("/etc/os-release")
```
davidcassany, line 218: I believe that if you pass this path as a parameter (or a relative root as a parameter), you will not need the chroot to rebrand grub and eventually extend `Elemental.SetGrubDefaultGrubEntry` to make use if this utility. So all the grub rebranding can be simplyfied `Elemental.Rebrand()`.

This can be done in a follow up PR

pkg/action/upgrade.go
```diff
@@ -67,53 +57,57 @@ func (u *UpgradeAction) Error(s string, args ...interface{}) {
...
+			map[string]string{
+				"/usr/local": "/usr/local",
+				"/oem":       "/oem",
+			},
+		)
+	}
+	return ActionHook(config, hook)
+}
+
+func rebrandChroot(config *v1.RunConfig) error {
```
davidcassany, line 73: I struggle to understand why a chroot is needed here. If this is only to load the `/etc/os-release` why not pass the root parameter to the loader utility? something like `utils.LoadOsRelease(config.Fs, stateDevice.MountPoint)`.

pkg/action/upgrade.go
```diff
@@ -67,53 +57,57 @@ func (u *UpgradeAction) Error(s string, args ...interface{}) {
...
-			"/usr/local": "/usr/local",
-			"/oem":       "/oem",
-		})
-		err = newChroot.Prepare()
+func upgradeHook(config *v1.RunConfig, hook string, chroot bool) error {
+	if chroot {
+		return ActionChrootHook(
+			config, hook, config.ActiveImage.MountPoint,
+			map[string]string{
+				"/usr/local": "/usr/local",
```
davidcassany, line 65: nit: I'd rather set this to the persistent and oem partitions mountpoint. Note that for RPI we were exploring images without OEM partition. Also I am not sure about the full upgrade matrix we support now, but, for instance, if running from recovery persistent partition is not mounted by default.

davidcassany (commented):
Looks good :+1: 

I just added a comment regarding the grub rebrand, I fail to see the need of a chroot there. It could also be I miss some detail.

pkg/utils/common.go
Itxaka: that is a really good point...

pkg/action/upgrade.go
Itxaka: Yeah mee too. I think I was so feed up with this not working that I mainly forced it to work however it can :laughing:

pkg/action/upgrade.go
Itxaka: Good point, noted, incoming PR.

PR #82, opened 2022-02-04
Fix chroot Prepare and Close methods

davidcassany:
This fixes Chroot.Prepare and Chroot.Close methods so they fully set and unset the chroot environment.

This enables to run chrooted hooks stages without having to actually call the elemental or yip from the chroot.

In addition this PR also include a little change to make configuration methods public.

### DIFF
--- a/pkg/action/common.go
+++ b/pkg/action/common.go
@@ -24,7 +24,9 @@ import (
 	"github.com/rancher-sandbox/elemental/pkg/utils"
 )
 
-func actionHook(config *v1.RunConfig, hook string) error {
+// ActionHook is RunStage wrapper that only adds logic to ignore errors
+// in case v1.RunConfig.Strict is set to false
+func ActionHook(config *v1.RunConfig, hook string) error {
 	config.Logger.Infof("Running %s hook", hook)
 	err := utils.RunStage(hook, config)
 	if !config.Strict {
@@ -33,7 +35,8 @@ func actionHook(config *v1.RunConfig, hook string) error {
 	return err
 }
 
-func actionChrootHook(config *v1.RunConfig, hook string, chrootDir string, bindMounts map[string]string) (err error) {
+// ActionChrootHook executes ActionHook inside a chroot environment
+func ActionChrootHook(config *v1.RunConfig, hook string, chrootDir string, bindMounts map[string]string) (err error) {
 	chroot := utils.NewChroot(chrootDir, config)
 	chroot.SetExtraMounts(bindMounts)
 	err = chroot.Prepare()
@@ -45,10 +48,11 @@ func actionChrootHook(config *v1.RunConfig, hook string, chrootDir string, bindM
 			err = tmpErr
 		}
 	}()
-	return actionHook(config, hook)
+	return ActionHook(config, hook)
 }
 
-func setupLuet(config *v1.RunConfig) {
+// SetupLuet sets the Luet object with the appropriate plugins
+func SetupLuet(config *v1.RunConfig) {
 	if config.DockerImg != "" {
 		plugins := []string{}
 		if !config.NoVerify {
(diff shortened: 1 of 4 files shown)


davidcassany:
@Itxaka you were absolutely right, the chroot hooks were not functional at all. There was no chroot at all 😞

pkg/action/common.go
```diff
@@ -57,3 +61,83 @@ func setupLuet(config *v1.RunConfig) {
 		config.Luet = v1.NewLuet(config.Logger, context.NewContext(), &dockTypes.AuthConfig{}, plugins...)
 	}
 }
+
+// SetPartitionsFromScratch initiates all defaults partitions in order is they
+// would be on a fresh installation. It does not run any kind of block device analysis
+// it only populates partitions from defaults or configurations.
+func SetPartitionsFromScratch(config *v1.RunConfig) {
```
Itxaka, line 68: is this related to this change? Seems...unrelated

pkg/utils/chroot.go
```diff
@@ -32,6 +32,7 @@ type Chroot struct {
 	extraMounts   map[string]string
 	activeMounts  []string
 	config        *v1.RunConfig
+	oldRootF      *os.File
```
Itxaka, line 35: good one

pkg/utils/chroot.go
```diff
@@ -51,9 +52,8 @@ func (c *Chroot) SetExtraMounts(extraMounts map[string]string) {
 	c.extraMounts = extraMounts
 }
 
-// Prepare will mount the defaultMounts as bind mounts, to be ready when we run chroot
-func (c *Chroot) Prepare() error {
-	var err error
+// Prepare will mount all required bind mounts and chroot in to the new path
+func (c *Chroot) Prepare() (err error) {
```
Itxaka, line 56: I have some issues with this. Mainly my expectation with the chroot was not to change the whole environment to a chroot directly as that can lead to weird things if we forget about this, or mistakenly think that `Prepare` is only a setup and something calls it and continues doing stuff, mistakenly doing it on the chrooted environment.

The old prepare was ok because it got everything readied, some bind mounts and then did not chroot until you run something **explicitly** on it.

Could we maybe do all the prepare stuff before hand and then have a wrapper called maybe `InsideChroot` that accepts a function to be run? And that cleans up the chroot env after running the command(s) on the chroot?

I dont even know if its possible, but for me that would be ideal, only run in chroot whatever functions we need, dont have it active otherwise :D

Itxaka (commented):
Just comments

pkg/utils/chroot.go
davidcassany: > I have some issues with this. Mainly my expectation with the chroot was not to change the whole environment to a chroot directly as that can lead to weird things if we forget about this, or mistakenly think that `Prepare` is only a setup and something calls it and continues doing stuff, mistakenly doing it on the chrooted environment.

This is good point and very valid concern...

> Could we maybe do all the prepare stuff before hand and then have a wrapper called maybe InsideChroot that accepts a function to be run? And that cleans up the chroot env after running the command(s) on the chroot?

> I dont even know if its possible, but for me that would be ideal, only run in chroot whatever functions we need, dont have it active otherwise :D

I am not sure this is possible, the ideal would be to pass an execution call as a parameter like we do in deferred calls, I did some quick googling over the weekend and I could not find how to do it. I'll do some more little search and if I don't find anything promising what I'll do is to revert this chroot refactor and add a new chroot method to explicitly run stages, not generic but probably we don't need to run generic code inside a chroot.

What do you think?

Itxaka: Well, there is at least one other place when we need to run a different command, and that is the "rebrand" on upgrade as it needs to run on the newly installed system to update the grub entries and potentially more if there is any cloud-config stuff in there


But yeah, if its not possible lets go with the simplest approach, have a chroot run stage and chroot rebrand.

davidcassany:
I am closing this PR for now. Gonna start form a scratch a different approach.

pkg/action/common.go
davidcassany: yes indeed, splitting to another small PR

pkg/utils/chroot.go
davidcassany: There is an option which is defining a wrapper function to what ever it needs to be called. A PR is coming, I think this will be fine.

PR #84, opened 2022-02-07
Add a Chroot.RunCallback method

davidcassany:
This commit add the RunCallback method on Chroot object.

This also fixes the action.ActionChrootHook.

### DIFF
--- a/pkg/action/common.go
+++ b/pkg/action/common.go
@@ -39,16 +39,10 @@ func ActionHook(config *v1.RunConfig, hook string) error {
 func ActionChrootHook(config *v1.RunConfig, hook string, chrootDir string, bindMounts map[string]string) (err error) {
 	chroot := utils.NewChroot(chrootDir, config)
 	chroot.SetExtraMounts(bindMounts)
-	err = chroot.Prepare()
-	if err != nil {
-		return err
+	callback := func() error {
+		return ActionHook(config, hook)
 	}
-	defer func() {
-		if tmpErr := chroot.Close(); tmpErr != nil && err == nil {
-			err = tmpErr
-		}
-	}()
-	return ActionHook(config, hook)
+	return chroot.RunCallback(callback)
 }
 
 // SetupLuet sets the Luet object with the appropriate plugins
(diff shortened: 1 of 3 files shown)


davidcassany:
@Itxaka at the end I think this is pretty reasonable. You can call any function but you have to wrap the call you want into a function without parameters and returning `error`. I think this is the simplest way of doing it and should be flexible enough.

pkg/utils/chroot.go
```diff
@@ -118,15 +119,72 @@ func (c *Chroot) Close() error {
 	return nil
 }
 
+// RunCallback runs the given callback in a chroot environment
+func (c *Chroot) RunCallback(callback func() error) (err error) {
+	// Store current root
+	oldRootF, err := os.Open(".") // Can't use afero here because doesn't support chdir done below
```
davidcassany: Note I changed that to open the current directory. I think this way we ensure to keep being at the same working directory after running any chroot action.

Itxaka: mmmh. Will that work? Like the working dir of the current system should be "/" always no? If you try to chroot back to /home/cos/" for example it will fail as its missing all the expected dirs for chroot to work.

pkg/action/common.go
```diff
@@ -39,16 +39,10 @@ func ActionHook(config *v1.RunConfig, hook string) error {
 func ActionChrootHook(config *v1.RunConfig, hook string, chrootDir string, bindMounts map[string]string) (err error) {
 	chroot := utils.NewChroot(chrootDir, config)
 	chroot.SetExtraMounts(bindMounts)
-	err = chroot.Prepare()
-	if err != nil {
-		return err
+	callback := func() error {
+		return ActionHook(config, hook)
```
Itxaka, line 43: pretty cool and simple, love it!

pkg/utils/chroot.go
davidcassany: yeah... probably this is a risky change. Gonna revert it

Itxaka (approved):

PR #86, opened 2022-02-07
Add reset command

### LINKED ISSUE #1049: `elemental reset`
port current `cos --reset` code and consume it in cOS-toolkit

davidcassany:
Fixes rancher-sandbox/cos-toolkit#1049

### DIFF
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -1,0 +1,92 @@
+/*
+Copyright © 2021 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"errors"
+	"github.com/rancher-sandbox/elemental/cmd/config"
+	"github.com/rancher-sandbox/elemental/pkg/action"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"k8s.io/mount-utils"
+	"os/exec"
+)
+
+// resetCmd represents the install command
+var resetCmd = &cobra.Command{
+	Use:   "reset",
+	Short: "elemental reset OS",
+	Args:  cobra.ExactArgs(0),
+	PreRun: func(cmd *cobra.Command, args []string) {
+		viper.BindPFlags(cmd.Flags())
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		path, err := exec.LookPath("mount")
+		if err != nil {
+			return err
+		}
+		mounter := mount.New(path)
+
+		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), mounter)
+
+		if err != nil {
+			cfg.Logger.Errorf("Error reading config: %s\n", err)
+		}
+
+		err = errors.New("Invalid options")
+		if viper.GetBool("reboot") && viper.GetBool("poweroff") {
+			cfg.Logger.Errorf("'reboot' and 'poweroff' are mutually exclusive options")
+			return err
+		}
+
+		if viper.GetString("cosign-key") != "" && !viper.GetBool("cosign") {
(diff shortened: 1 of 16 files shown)


cmd/reset.go
```diff
@@ -0,0 +1,93 @@
...
+
+		cmd.SilenceUsage = true
+		err = action.ResetSetup(cfg)
+		if err != nil {
+			return err
+		}
+
+		cfg.Logger.Infof("Reset called")
+
+		err = action.ResetRun(cfg)
```
mudler:
```suggestion
		return action.ResetRun(cfg)
```

pkg/action/reset.go
```diff
@@ -0,0 +1,275 @@
...
+			config.Fs.Remove(config.ActiveImage.File)
+			return err
+		}
+		config.Logger.Infof("Finished copying Active...")
+		err = ele.MountImage(&config.ActiveImage, "rw")
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if tmpErr := ele.UnmountImage(&config.ActiveImage); tmpErr != nil && err == nil {
```
mudler, line 220: ok I see it now it, the defer is just in case of errors, right?

maybe makes sense to define a private func that is shared over so it's more clear and avoid the dups around as we can call it also down below. At the end IF there is an error and this gets hit we just want the error that generated the failure on the first place,  I'd just print a warning here in case the umounts fails.

mudler (approved):
Just small nits, but looks good here!

pkg/action/reset.go
```diff
@@ -0,0 +1,275 @@
...
+	// TODO execute rootTree sanity checks
+	if config.Directory != "" {
+		rootTree = config.Directory
+	} else if config.DockerImg != "" {
+		rootTree = config.DockerImg
+	} else if utils.BootedFrom(config.Runner, cnst.RecoverySquashFile) {
+		rootTree = cnst.IsoBaseTree
+	}
+
+	_, err := config.Fs.Stat(cnst.EfiDevice)
```
Itxaka: we can use `exists, _ := afero.Fs(config.Fs, cnst.EfiDevice)` directly I think and it takes care of checking the error type to see if there was a real error or the error is that the file does not exists.

pkg/action/reset.go
Itxaka: Check what Im doing on the upgrade part, I keep a `cleanup` struct to track things that need to be removed, unmounted or remounted and allows to keep the original error and print new errors.

Its a bit rougth but Im sure we could make it generic for use everywhere and you dont have to track all this stuff manually.

pkg/constants/constants.go
```diff
@@ -76,6 +79,8 @@ const (
 	UpgradeRecoveryDir     = "/run/initramfs/live"
 	TransitionImgFile      = "transition.img"
 	TransitionSquashFile   = "transition.squashfs"
+	// TODO converge this constant with StateDir/RecoveryDir in dracut module from cos-toolkit
+	RunningStateDir = "/run/initramfs/cos-state"
```
Itxaka, line 82: we should merge UpgradeStateDir with this one :)

davidcassany: yes I know, that's why there is a TODO, there constants require some care.

pkg/action/reset.go
davidcassany: I like the idea of making some generic approach for it. However I am not fully convinced with Cleanup struct approach, or at least with the idea of tracking specific very specific things (mountpoints and such), as this won't work reset and install commands, as they make extensive use of Elemental package to do all this stuff. So probably a struct stacking callbacks (and some metadata if needed) with a cleanup method doing err management logic.

I'd like being capable of doing something similar to:

```golang
...
defer cleanup.clean()
...
err := elemental.MountPartitions()
if err != nil {
  return err
}
cleanup.AddCallback(func() error {return elemental.UnmountPartitions()})
...
```

Probably it can be written in a nicer way

Itxaka: yes, that would be pretty nice, cleanup was very specifically done for upgrade so it only suits there :D

Itxaka (approved):

pkg/elemental/elemental.go
```diff
@@ -282,36 +282,48 @@ func (c *Elemental) CopyActive(source v1.InstallUpgradeSource) error {
...
 		}
-	}
-
-	if source.IsChannel {
+	} else if source.IsChannel {
 		err = c.config.Luet.UnpackFromChannel(c.config.ActiveImage.MountPoint, source.Source)
 		if err != nil {
 			return err
 		}
+	} else if source.IsFile {
```
Itxaka, line 312: nice

Itxaka (approved):

PR #87, opened 2022-02-08
Unpack local flag v2

Ludea:
Add local flag to unpack to avoid pulling remote image

### DIFF
--- a/pkg/types/v1/luet.go
+++ b/pkg/types/v1/luet.go
@@ -25,7 +25,7 @@ import (
 )
 
 type LuetInterface interface {
-	Unpack(string, string) error
+	Unpack(string, string, bool) error
 }
 
 type Luet struct {
@@ -51,7 +51,7 @@ func NewLuet(log Logger, context *context.Context, auth *dockTypes.AuthConfig, p
 	}
 }
 
-func (l Luet) Unpack(target string, image string) error {
+func (l Luet) Unpack(target string, image string, local bool) error {
 	l.log.Infof("Unpacking docker image: %s", image)
 	info, err := docker.DownloadAndExtractDockerImage(l.context, image, target, l.auth, l.VerifyImageUnpack)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/types/v1/luet.go
```diff
@@ -60,4 +60,19 @@ func (l Luet) Unpack(target string, image string) error {
 	l.log.Infof("Pulled: %s %s", info.Target.Digest, info.Name)
 	l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
 	return nil
+	if !local {
```
Itxaka, line 56: This should start in line 55, substituting what's in there, otherwise this will never be reached

Itxaka: wait, this is correct...why it was showing different before???

pkg/types/v1/luet_test.go
```diff
@@ -42,7 +46,17 @@ var _ = Describe("Types", Label("luet", "types"), func() {
 	Describe("Luet", func() {
 		It("Fails to unpack without root privileges", Label("unpack"), func() {
 			image := "quay.io/costoolkit/releases-green:cloud-config-system-0.11-1"
-			Expect(luet.Unpack(target, image)).NotTo(BeNil())
+			Expect(luet.Unpack(target, image, false)).NotTo(BeNil())
+			image := "docker.io/library/alpine"
```
Itxaka, line 52: This should be its own test IMO.

```suggestion
		})
		It("Unpacks local image", Label("unpack", "root"), func() {
			image := "docker.io/library/alpine"
```

pkg/types/v1/luet_test.go
```diff
@@ -42,7 +46,19 @@ var _ = Describe("Types", Label("luet", "types"), func() {
 	Describe("Luet", func() {
 		It("Fails to unpack without root privileges", Label("unpack"), func() {
 			image := "quay.io/costoolkit/releases-green:cloud-config-system-0.11-1"
-			Expect(luet.Unpack(target, image)).NotTo(BeNil())
+			Expect(luet.Unpack(target, image, false)).NotTo(BeNil())
+		})
+		It("Check that luet can unpack the local image", Label("unpack"), func() {
```
Itxaka:
```suggestion
		It("Check that luet can unpack the local image", Label("unpack", "root"), func() {
```

pkg/types/v1/luet_test.go
```diff
@@ -39,10 +43,22 @@ var _ = Describe("Types", Label("luet", "types"), func() {
 	AfterEach(func() {
 		Expect(os.RemoveAll(target)).To(BeNil())
 	})
-	Describe("Luet", func() {
+	Describe("Luet", Label("luet", "root"), func() {
```
Itxaka: Not here, as we want the first test `"Fails to unpack without root privileges"` to fail and confirm that we need root privs to execute unpack, and this label here will trickle down into any tests inside this "container", so that test will inherit this label.

Itxaka (approved):
nice!

PR #89, opened 2022-02-08
Changes by running 'make fmt'

davidcassany:


davidcassany:
I thought that the lint check would catch these little leftovers, it seams it doesn't 🤷🏽‍♂️
I guess it does not return an error if some changes are applied.

Itxaka (approved):
Weird, I swear I thought I added this as part of make lint :/

davidcassany:
> Weird, I swear I thought I added this as part of make lint :/

I believe it is just executed, applies changes in the local checkout and it doesn't error out.

Itxaka:
I don't understand codecov. How is this supposed to increase coverage???

PR #90, opened 2022-02-08
Ensure make fmt returns error if some file is changed

davidcassany:


davidcassany:
Requires #89 to pass the lint test

Itxaka (approved):

PR #92, opened 2022-02-09
Adding CleanStack utility

davidcassany:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -65,6 +65,8 @@ func InstallSetup(config *v1.RunConfig) error {
 func InstallRun(config *v1.RunConfig) (err error) {
 	newElemental := elemental.NewElemental(config)
 	installSource := v1.InstallUpgradeSource{}
+	cleanup := utils.NewCleanStack()
+	defer func() { err = cleanup.Cleanup(err) }()
 
 	if config.DockerImg != "" {
 		installSource.Source = config.DockerImg
@@ -91,13 +93,11 @@ func InstallRun(config *v1.RunConfig) (err error) {
 		if err != nil {
 			return err
 		}
-		defer func() {
+		cleanup.Push(func() error {
 			config.Logger.Infof("Unmounting downloaded ISO")
-			if tmpErr := config.Mounter.Unmount(config.IsoMnt); tmpErr != nil && err == nil {
-				err = tmpErr
-			}
 			config.Fs.RemoveAll(tmpDir)
-		}()
+			return config.Mounter.Unmount(config.IsoMnt)
+		})
 	}
 
 	// Check device valid
@@ -125,11 +125,7 @@ func InstallRun(config *v1.RunConfig) (err error) {
 	if err != nil {
 		return err
 	}
-	defer func() {
-		if tmpErr := newElemental.UnmountPartitions(); tmpErr != nil && err == nil {
-			err = tmpErr
-		}
-	}()
+	cleanup.Push(func() error { return newElemental.UnmountPartitions() })
 
 	// create active file system image
 	err = newElemental.CreateFileSystemImage(config.ActiveImage)
@@ -142,11 +138,7 @@ func InstallRun(config *v1.RunConfig) (err error) {
 	if err != nil {
 		return err
 	}
-	defer func() {
-		if tmpErr := newElemental.UnmountImage(&config.ActiveImage); tmpErr != nil && err == nil {
-			err = tmpErr
-		}
-	}()
+	cleanup.Push(func() error { return newElemental.UnmountImage(&config.ActiveImage) })
 
 	// install Active
 	err = newElemental.CopyActive(installSource)
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -191,6 +191,12 @@ func InstallRun(config *v1.RunConfig) (err error) {
 		return err
 	}
 
+	// Do not reboot/poweroff on cleanup errors
+	err = cleanup.Cleanup(err)
+	if err != nil {
+		return err
+	}
+
 	// Reboot, poweroff or nothing
 	if config.Reboot {
 		config.Logger.Infof("Rebooting in 5 seconds")
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -18,7 +18,6 @@ package action
 
 import (
 	"fmt"
-	"github.com/hashicorp/go-multierror"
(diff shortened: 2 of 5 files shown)


davidcassany:
This is a draft PR based on https://github.com/rancher-sandbox/elemental/pull/86#discussion_r802552890 It is an initial approach for discussion. I am not sure it fully suits our needs as is.

@mudler @Itxaka thoughs are more than welcome

pkg/action/install.go
```diff
@@ -125,11 +125,7 @@ func InstallRun(config *v1.RunConfig) (err error) {
 	if err != nil {
 		return err
 	}
-	defer func() {
-		if tmpErr := newElemental.UnmountPartitions(); tmpErr != nil && err == nil {
-			err = tmpErr
-		}
-	}()
+	cleanup.Push(func() error { return newElemental.UnmountPartitions() })
```
davidcassany, line 128: I am not happy with the syntax of having to add the `func() error {...}` but I could not think up of any other alternative.

pkg/action/install.go
```diff
@@ -65,6 +65,8 @@ func InstallSetup(config *v1.RunConfig) error {
 func InstallRun(config *v1.RunConfig) (err error) {
 	newElemental := elemental.NewElemental(config)
```
davidcassany, line 66: I also thought about the idea of passing the cleanstack to elemental, so it can eventually feed the stack internally. For instance the method `Elemental.MountPartitions()` could actually push the `Elemental.UnmountPartitions()` into the clean stack.

From one side I like this idea as it prevents us having to manually feed the cleanup stack, but from the other side it feels this could be too magic from the `pkg/action/install.go` POV.

Itxaka: Agreed, I rather be explicit that magical. If you mount/create anything you should be in charge of unmounting/removing, otherwise it can get chaotic

Itxaka:
I like it! AFAIK it just do things in reverse order right? Very useful!

Not much to comment, it already look much nicer than my cleaner!

davidcassany:
> I like it! AFAIK it just do things in reverse order right? Very useful!

Correct, it is just a LIFO queue of callbacks. You think it is better to include the upgrade command in this PR or leave it for a second round?

Itxaka:
> Correct, it is just a LIFO queue of callbacks. You think it is better to include the upgrade command in this PR or leave it for a second round?

The code is simple enough, I think I can be on this one as well unless  @mudler has some other ideas regarding the CleanStack

pkg/utils/cleanstack.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil
+	}
+	clean.count--
+	return clean.jobs[clean.count]
+}
+
+// Cleanup runs the whole cleanup stack. In case of error it runs all jobs
+// and returns the first error occurrence.
+func (clean *CleanStack) Cleanup(err error) error {
+	var tmpErr error
```
Itxaka: would it be possible to extend this with the multierror library to store all errors that happens and report all of those? That would at least give the user a hint of what is pending in the system, i.e. 2 paths coult not be unmounted, or one path was not unmounted and another dir was not cleaned up?

davidcassany: Yes we can, I did not include the multierror library as this pulls a new third party dependency stack, but yes I guess it doesn't make much sense at this point worrying of adding an extra dependency we already have tons of them 😅 gonna add it.

Itxaka: I think it should be in already, I use it as part of the upgrade :)

pkg/action/upgrade.go
```diff
@@ -277,96 +271,62 @@ func (u *UpgradeAction) Run() error {
...
-		err = u.cleanup(cleanup, err)
-		if err != nil {
-			// If cleanup fails, do not reboot
-			return err
-		} else {
-			u.Info("Rebooting in 5 seconds")
-			return utils.Reboot(u.Config.Runner, 5)
-		}
+		u.Info("Rebooting in 5 seconds")
+		return utils.Reboot(u.Config.Runner, 5)
```
Itxaka, line 297: how does this delay affects the cleanup?  Is the cleanup run due to the `defer` then this? Wont that potentially take more than 5 seconds after the message?

pkg/action/upgrade.go
```diff
@@ -277,96 +271,62 @@ func (u *UpgradeAction) Run() error {
...
-			return originalError
-		}
-		return errs
-	} else {
-		if originalError != nil {
-			return originalError
-		}
-		return nil
+// remount attemps to remount the given mountpoint with the provided options. Does nothing if not mounted
+func (u *UpgradeAction) remount(m mount.MountPoint, opts ...string) error {
```
Itxaka, line 324: This methods are now perfect ot move into utils IMO. For a further PR down the road.

Itxaka (approved):
Looking good

pkg/action/upgrade.go
Itxaka: probably should move the message into the reboot method :D not needed to change in this PR.

davidcassany: The cleanup stack is ampty at that point. We cleared manually just before. The sequence goes as follows:

1. in u.Run utils.Reboot is called
2. in utils.Reboot waits for 5 seconds
3. in utils.Reboot shells out `reboot -f`
4. in u.Run returns
5. deferred methods of u.Run are executed
6. in defer cleanup.Cleaup(err) is executed but does nothing, the stack is empty

I guess a better way to do it would be something like.

```golang
err = cleanup.Cleanup(err)
if err != nil {
    return err
}

if u.Config.Reboot {
    u.Info("Rebooting in 5 seconds")
    defer utils.Reboot(u.Config.Runner, 5)
}

return err
 ```
So running the reboot is last thing done before returning. Also probably makes sense having an `elemental.Turnoff()` that checks `RunConfig.Reboot` and runs `poweroff`, `reboot` or nothing.

So we could have:

```golang
err = cleanup.Cleanup(err)
if err != nil {
    return err
}

return  ele.Turnoff()
```

pkg/action/upgrade.go
davidcassany: Yes I was also thinking so

pkg/action/upgrade.go
Itxaka: oh I see, its called just before indeed. Nice enough for me, no need to modify it further, thanks!

Issue #33, opened 2021-12-31
Refactor RunConfig struct


davidcassany:
as per discussion in https://github.com/rancher-sandbox/elemental-cli/pull/32#discussion_r776967358 there is room to improve how we are processing and organizing run time configuration details. This can be actually our way to read the manifest in the various phases ( e.g. https://github.com/rancher-sandbox/elemental/pull/160 )

Ideas:
* Define a set of generic configs and pass it to methods instead of moving around the whole `RunConfig`
* Defile a config struct isolated from flags and config files (no marshall/unmarshall) for each subcommand (aka `InstallConfig` struct)

See also:
 https://github.com/rancher-sandbox/elemental/blob/f1228393e3f342c07e265850b7f4c06f0c0128d3/pkg/types/v1/config.go#L47

mudler:
I would add to move validation outside from `cmd` and be strict with the config itself, this way it is detached from the cli framework being used, and can be used in different contexts (validate a config file, for example)

davidcassany:
Here a proposal from my side (covers install/upgrade/reset commands):

 ```yaml
install:
  # target is the only mandatory
  target: /dev/sda

  # basic disk configs for formatting
  firmware: efi|bios
  partitionTable: gpt|msdos

  # partitions setup, any missing value is taken from defaults
  partitions:
    oem:
      label:
      size:
      fs:
      flags:
    recovery:
      ...
    state:
      ...
    persistent:
      ...

  # recovery image should be squashed or not (eventual future feature)
  squashfsRecovery: true|false

  # format disk or not
  noFormat: true|false

  # use this iso as installation media (overwrites 'image' and 'recoveryImage' sources with iso contents)
  iso:

  # main OS image
  image:
    label:
    size:
    fs:
    source:
      type: dir|channel|package|image|file
      uri:

  # these are recovery image details, uses the same as 'image' if unset
  recoveryImage:
    label:
    size:
    fs:
    source:
      type: dir|channel|package|image|file
      uri:
    
upgrade:
  # upgrade system to this OS Image
  image:
    label:
    size:
    fs:
    source:
      type: dir|channel|package|image|file
      uri:

  # upgrade recovery image to this setup
  recoveryImage:
    label:
    size:
    fs:
    source:
      type: dir|channel|package|image|file
      uri:
    
reset:
  # reset to this OS image
  image:
    label:
    size:
    fs:
    source:
      type: dir|channel|package|image|file
      uri:

  # format persistent partition or not
[...]

PR #94, opened 2022-02-10
Rework flags to unify some of them

### LINKED ISSUE #79: Move flags validaton to some generic place (`cmd/flags.go`?)
This to avoid having to duplicate flags checks incompatibilities like `--reboot` and `--poweroff` in all subcommands and reuse the logic.

In addition we could align which flags should be set from env_variables, config files and command line. For instance `--reboot` and `--poweroff` do not make much sense from a config file.

See https://github.com/rancher-sandbox/elemental/pull/72#discussion_r799453623

Itxaka:
Fixes #79 

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -1,0 +1,95 @@
+/*
+Copyright © 2021 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"errors"
+	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// addCosignFlags adds flags related to cosign
+func addCosignFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolP("cosign", "", false, "Enable cosign verification (requires images with signatures)")
+	cmd.Flags().StringP("cosign-key", "", "", "Sets the URL of the public key to be used by cosign validation")
+}
+
+// addPowerFlags adds flags related to power
+func addPowerFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolP("reboot", "", false, "Reboot the system after install")
+	cmd.Flags().BoolP("poweroff", "", false, "Shutdown the system after install")
+}
+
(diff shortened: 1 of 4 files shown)


davidcassany (approved):
Nice, looks good!

Just a couple of comments regarding further arrangements on flags.

What about binding `--reboot` and `--poweroff` (and probably others) manually? (I mean no config file)
Also `--directory` from reset and upgrade should be merged, `--directory` makes sense for the installer too (I think this is a miss from the former code, as this was only possible by using env vars so I missed it).

Itxaka:
> What about binding --reboot and --poweroff (and probably others) manually? (I mean no config file)

Ummm I dont think its possible with our current setup. Anything that its on the files loaded that matches the Runconfig key will be unmarshalled there. We could make the keys in Runconfig be different so they dont match I guess? Like we could have Runconfig.PowerOffOnFinishTask so to match that from a config file it should have `poweroffonfinishtask: true` values in the config, but in the end it can be matched.

We could bind them to a local var instead and use that to set the value in the config directly if we dont want to expose it, i.e. :

```
var reboot bool 
upgradeCmd.Flags().BoolVar(&reboot, "--reboot", false, "")

```

But this means generating that `reboot` var in the root of the file so it can be accessed by the init and the Execute method, which is kind of bad. Also not sure if we can bind the flag in the Execute/PreRun methods 

We could also delay the command adding and use a function that does all of this stuff, and doesnt add itself to the root command. Instead on the root command you can that function that returns the cmd to add it to your root command. But that requires basically rewriting the whole cmd dir lol

Let me try if we are able to get the real flag value on Execute and bind it directly to the config value but Im not sure if that is possible at that point in time.

In any case, any reason for that? One of the cool things of the viper stuff its that you could have all your config directly on a config file, so no need for flags at all....not like we are using the config file anywhere lol

davidcassany:
> In any case, any reason for that? One of the cool things of the viper stuff its that you could have all your config directly on a config file, so no need for flags at all....not like we are using the config file anywhere lol

I just thought that is kind a weird setting reboot on a config file, the same applies for the `--directory` flag, feels weird having it on a config file, those things are likely to be runtime and host dependent decisions. Anyway, you are right that this is additional flexibility. So probably we can assume mapping all flags to config values and keep it simple.

Itxaka:
@davidcassany seems to work to override flags in the Execute methods so we can always ignore whatever is coming from the config and override the flag directly only with the flag value.

`cfg.Reboot, err = cmd.Flags().GetBool("reboot")` will override any values from the config or env and set the value passed via the flag.

Itxaka:
We could also have a different struct for flag-only options and not bind that, only bind the config/flag/env options and then merge both together in the ReadConfigRun so we know which options can only be toggled from one or the other

davidcassany:
> We could also have a different struct for flag-only options and not bind that, only bind the config/flag/env options and then merge both together in the ReadConfigRun so we know which options can only be toggled from one or the other

Yes that's also a good idea. I think this requires some more thinking about how to organize environment variables, config files and flags. Probably it could also make sense having a similar approach for env variables...

There's nothing wrong with the current implementation, just that it feels a bit messy. I guess this all related to #33 somehow, I have not a good/clear idea about how to do it though.

PR #95, opened 2022-02-10
Refactor CopyActive method to turn tin CopyImage

davidcassany:
This commit refactors CopyActive method to become CopyImage, where
the image being copied is passed as a parameter. This makes it easy to
reuse it other contexts beyond copying the active image.

In the same scope the refactor includes the image source struct as part
of the v1.Image struct. So all the details of the image are
self contained.

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -190,7 +190,7 @@ func (u *UpgradeAction) Run() (err error) {
 		Label:      u.Config.ActiveLabel,
 		FS:         constants.LinuxImgFs,
 		MountPoint: constants.UpgradeTempDir,
-		RootTree:   upgradeSource.Source, // if source is a dir it will copy from here, if it's a docker img it uses Config.DockerImg IN THAT ORDER!
+		Source:     upgradeSource, // if source is a dir it will copy from here, if it's a docker img it uses Config.DockerImg IN THAT ORDER!
 	}
 
 	// If on recovery, set the label to the RecoveryLabel instead
@@ -202,7 +202,7 @@ func (u *UpgradeAction) Run() (err error) {
 
 	if !isSquashRecovery {
 		// Only on recovery+squash we dont use the img file
-		err = ele.CreateFileSystemImage(img)
+		err = ele.CreateFileSystemImage(&img)
 		if err != nil {
 			u.Error("Failed to create %s img: %s", transitionImg, err)
 			return err
@@ -223,7 +223,7 @@ func (u *UpgradeAction) Run() (err error) {
 	}
 	// Setting the activeImg to our img, tricks CopyActive into doing it anyway even if it's a recovery img
 	u.Config.ActiveImage = img
-	err = ele.CopyActive(upgradeSource)
+	err = ele.CopyImage(&img)
 	if err != nil {
 		u.Error("Error copying active: %s", err)
 		return err
@@ -357,8 +357,8 @@ func (u *UpgradeAction) remount(m mount.MountPoint, opts ...string) error {
 }
 
 // getTargetAndSource finds our the target and source for the upgrade
-func (u *UpgradeAction) getTargetAndSource() (string, v1.InstallUpgradeSource) {
-	upgradeSource := v1.InstallUpgradeSource{Source: constants.UpgradeSource, IsChannel: true}
+func (u *UpgradeAction) getTargetAndSource() (string, v1.ImageSource) {
+	upgradeSource := v1.ImageSource{Source: constants.UpgradeSource, IsChannel: true}
 	upgradeTarget := constants.UpgradeActive
 
 	// if channel_upgrades==true then it picks the default image from /etc/cos-upgrade-image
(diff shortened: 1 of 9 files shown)


pkg/action/upgrade.go
```diff
@@ -196,7 +196,7 @@ func (u *UpgradeAction) Run() (err error) {
 	}
 	// Setting the activeImg to our img, tricks CopyActive into doing it anyway even if it's a recovery img
 	u.Config.ActiveImage = img
```
Itxaka, line 225: I guess this is not needed anymore?

pkg/elemental/elemental.go
```diff
@@ -277,59 +277,97 @@ func (c Elemental) CreateFileSystemImage(img v1.Image) error {
...
+		err = c.UnmountImage(img)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Copies the given filesystem image into the target
+func (c *Elemental) CopyImage(img *v1.Image) error {
```
Itxaka, line 318: I fell like copyImage is a realyl bad description. Like we are really just copying from one point to another using different methods depending on the source type. The only thing related to the image here is  the tune2fs and Im not even sure it belongs here as this is mainly extracting stuff from a to b, this method has no idea that the mountpoint at b is an image file(or not in case of recovery-squashfs)

pkg/elemental/elemental.go
```diff
@@ -277,59 +277,97 @@ func (c Elemental) CreateFileSystemImage(img v1.Image) error {
...
 			return err
 		}
-		err = utils.CopyFile(c.config.Fs, source.Source, c.config.ActiveImage.File)
+		err = utils.CopyFile(c.config.Fs, img.Source.Source, img.File)
 		if err != nil {
 			return err
 		}
-		_, err = c.config.Runner.Run("tune2fs", "-L", c.config.ActiveLabel, c.config.ActiveImage.File)
+		//TODO do not run tune2fs on squashfs images
+		_, err = c.config.Runner.Run("tune2fs", "-L", img.Label, img.File)
```
Itxaka, line 359: `isSquashRecovery, _ := afero.Exists(u.Config.Fs, filepath.Join(upgradeStateDir, "cOS", constants.RecoverySquashFile))`

 :)

Itxaka: oh or even checking the img.name should be enough! `filepath.Join(upgradeStateDir, "cOS", constants.TransitionSquashFile)`

pkg/types/v1/config.go
```diff
@@ -241,12 +241,11 @@ type PartitionList []*Partition
...
-	Label string
-	Size  uint
-	FS    string
-	// Path of the root tree
-	RootTree   string
+	File       string
+	Label      string
+	Size       uint
+	FS         string
+	Source     ImageSource
```
Itxaka, line 248: I dont think this should belong here

Itxaka (approved):
code looks good, not really onboard with the change but it can be changed further down the line I guess

pkg/action/upgrade.go
davidcassany: This is needed for the chroot hook I think.

pkg/elemental/elemental.go
davidcassany: I was thinking to check the filesystem and the label value emptyness. A squashfs images do not have labels.

pkg/types/v1/config.go
davidcassany: How would you define the passive, transition and recovery images then? The whole point of this refactor is to reach the point where we can set all images we need and simply deploy/create them.

My idea is that we can set

```golang
activeImg := &v1.Image {
    File:         "active.img",
    FS:           "ext4",
    Size:         "2048",
    Mountpoint:   "/run/cos/active",
    Label:        "COS_ACTIVE",
    Source:       "docker/file/dir/channel",
}

passiveImg := &v1.Image {
     File:  "passive.img",
     Label: "COS_PASSIVE",
     Source: v1.ImageSource(Source: activeImg.File, IsFile: true),
}

elemental.DeployImage(activeImg)
elemental.DeployImage(passiveImg)
```

I think that if we elabore it properly we can also use it for upgrade setting the transition image properly. 
Similarly with the `RunConfig.Partitions` I do not see it as a disk representation, I see it as the partitions list that is relevant to current execution, on install it means the whole disk and partitions are filled with config data, on reset we load the needed ones from the current disk(s). So I was after to a similar concept for images.

I am opened to other concepts too, I envisioned this since install command coding, but this can be discussed.

pkg/elemental/elemental.go
davidcassany: Yes, I agree CopyImage is not a great description... The squashfs issue is easy to solve, it should essentially check the filesystem (or label different to empty string) and not apply any label in that case. I thought to add it in a follow up step since in this PR `v1.Image` is only used for the ActiveImage currently.

In fact, ideally, from an action PoV my idea is that the `DeployImage` (could also be called differently, CreateImage?) is the one to use. If `v1.Image` is set of filesystem squashfs and source a dir it should do all the required steps, including mksquashfs call.

Itxaka (approved):

Itxaka:
I dont have any better ideas to contribute sorry, I think I get what your are moving forward to.

PR #96, opened 2022-02-10
Create new command 

Ludea:
New command to create derivative.
Currently support only opensuse


mudler:
That's a really nice touch! This was part of https://github.com/rancher-sandbox/cOS-toolkit/issues/108 so it definitely makes sense to have it

cmd/new.go
```diff
@@ -0,0 +1,86 @@
...
+
+		if err != nil {
+			cfg.Logger.Errorf("Error reading config: %s\n", err)
+		}
+
+		flavor := args[0]
+
+		client := &getter.Client{
+			Ctx: context.Background(),
+			Dst: "derivatives/opensuse",
```
mudler:
```suggestion
			Dst: fmt.Sprintf("derivatives/%s",flavor),
```

cmd/new.go
```diff
@@ -0,0 +1,86 @@
...
+			Dir: true,
+			Src: "github.com/rancher-sandbox/cOS-toolkit//examples/standard",
```
mudler:
```suggestion
			Src: "github.com/rancher-sandbox/cOS-toolkit/examples/standard",
```

cmd/new.go
```diff
@@ -0,0 +1,86 @@
...
+			Mode: getter.ClientModeDir,
+			Detectors: []getter.Detector{
+				&getter.GitHubDetector{},
+			},
+		}
+
+		var destination string
```
mudler: I guess we can move this up and generate it for each flavor

cmd/new.go
```diff
@@ -0,0 +1,86 @@
...
+		switch flavor {
+		case "opensuse": destination = "derivatives/opensuse"
+		case "fedora":
+		case "ubuntu":
+		default: cfg.Logger.Errorf("Unsupported flavor")
+			return errors.New("Unsupported flavor")
+		}
+
+		client = &getter.Client{
```
mudler: Looks like a dup here?

cmd/new.go
```diff
@@ -0,0 +1,86 @@
...
+	SilenceUsage:  true, // Do not show usage on error
+	SilenceErrors: true, // Do not propagate errors down the line, we control them
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cfg, err := config.ReadConfigRun(viper.GetString("config-dir"), &mount.FakeMounter{})
+
+		if err != nil {
+			cfg.Logger.Errorf("Error reading config: %s\n", err)
+		}
+
+		flavor := args[0]
```
mudler, line 40: I'd suggest to rather gate here the ones currently supported by the examples

cmd/new.go
Ludea: https://github.com/hashicorp/go-getter#subdirectories
I need the double / to download the subfolder

cmd/new.go
Ludea: Done

cmd/new.go
Ludea: Done

cmd/new.go
Ludea: Will do separate PR to add more examples

mudler (commented):
Looks good here, any thoughts @Itxaka @davidcassany ?

davidcassany (approved):
Looks good to me, nice!

mudler:
@Ludea we need commits to be signed-off, otherwise looks good to merge!

PR #98, opened 2022-02-10
Fix luet not having a logger and changing context

Itxaka:
Problem 2fold, first its that if we changed context, the bus.manager
could get lost as we registered the plugins under one context but then
changed contexts. The second is that we need to set a luet logger now as
we create an empty context. as a side effect of that, it means that we
cna control the log level of luet actions, so we could even silence it
adn only show errors. Currently it sets at the same level as our logger


Itxaka:
blergh

davidcassany (approved):
Looks good to me :+1:

PR #99, opened 2022-02-11
cloud-init: Process users before SSH

mudler:
See: https://github.com/rancher-sandbox/os2/issues/15


Itxaka:
Isn't this a yip issue? I mean, ok as a workaround but it should be fixed at the source?

Itxaka (approved):

mudler:
> Isn't this a yip issue? I mean, ok as a workaround but it should be fixed at the source?

We are the sources here :), we define a custom execution order in the cloud runner

PR #100, opened 2022-02-11
Upgrade: only add mounts for chroot if they have a mountpoint

Itxaka:


pkg/action/upgrade.go
```diff
@@ -122,6 +122,31 @@ func (u *UpgradeAction) Run() (err error) {
 	if !u.Config.RecoveryUpgrade && utils.BootedFrom(u.Config.Runner, u.Config.RecoveryLabel) {
 		statePartMountOptions = []string{"rw"}
 		cleanup.Push(func() error { return u.unmount(upgradeStateDir) })
+		// Also mount oem and persistent as they are not mounted on recovery
+		oemPart, err := utils.GetFullDeviceByLabel(u.Config.Runner, u.Config.OEMLabel, 5)
```
davidcassany, line 126: For the future, but probably this logic could be substituted by  `elemental.MountPartitions`, it mounts all `RunConfig.Partitions` that have a mountpoint defined. I think only the `remount` use case is missing in `MountPartitions`, probably we could store the mount options as part of `v1.Partition` and then have an utility method in `pkg/actions/common.go` that loads state, recovery, oem and persistent partitions from the system if they exist. Then it should be simple to fine tune manually RunConfig.Partitions according the use case so  `elemental.MountPartitions` is doing what is needed.

davidcassany (approved):
Looks good

pkg/action/upgrade.go
Itxaka: problem is that then in the upgrade we should fill the RunConfig.Partitions rigth? Which has not much to do with the upgrade, its only kind of forcing a methods that is meant for X to work for Z. I dunno, I like it being explicit like this, its clear what this is doing and in which case.

PR #101, opened 2022-02-11
Fix upgrade source from channel in recovery

Itxaka:
We were picking the default system/cos but we should be picking the
recovery one and falling back into the default image

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -359,19 +359,22 @@ func (u *UpgradeAction) getTargetAndSource() (string, v1.InstallUpgradeSource) {
 	upgradeSource := v1.InstallUpgradeSource{Source: constants.UpgradeSource, IsChannel: true}
 	upgradeTarget := constants.UpgradeActive
 
-	// if upgrade_recovery==true then it upgrades only the recovery
-	// if upgrade_recovery==false then it upgrades only the active
-	// default is active
-	if u.Config.RecoveryUpgrade {
-		u.Debug("Upgrading recovery")
-		upgradeTarget = constants.UpgradeRecovery
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/action/upgrade.go
```diff
@@ -359,19 +359,22 @@ func (u *UpgradeAction) getTargetAndSource() (string, v1.InstallUpgradeSource) {
...
-	}
-
 	// if channel_upgrades==true then it picks the default image from /etc/cos-upgrade-image
-	// this means, it gets the UPGRADE_IMAGE(default system/cos) from the luet repo configured on the system
+	// this means, it gets the UPGRADE_IMAGE(default system/cos)/RECOVERY_IMAGE from the luet repo configured on the system
 	if u.Config.ChannelUpgrades {
 		u.Debug("Source is channel-upgrades")
-		upgradeSource.Source = u.Config.UpgradeImage // Loaded from /etc/cos-upgrade-image
+
+		if u.Config.RecoveryUpgrade {
```
mudler, line 367: I'd suggest to rewire it simply (need to be adapted a bit):

```golang
upgradeSource.Source = u.Config.UpgradeImage
if u.Config.RecoveryUpgrade {
	u.Debug("Upgrading recovery")
	upgradeTarget = constants.UpgradeRecovery
	if u.Config.RecoveryImage != "" {
		upgradeSource.Source = u.Config.RecoveryImage
	}
}
```

Itxaka: IIRC we cannot assume the existence of u.Config.UpgradeImage at least according to the old cos.sh, so it needs to be a bit more convoluted. Thats why I init the source withe the constants value then overwrite it afterwards

mudler: right, missed it, looks good!

mudler (approved):

PR #104, opened 2022-02-14
Be more smart about upgrade dir

Itxaka:

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -27,6 +27,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


pkg/utils/common.go
```diff
@@ -228,3 +229,20 @@ func LoadEnvFile(fs afero.Fs, file string) (map[string]string, error) {
...
+	dir := os.Getenv("TMPDIR")
+	if dir != "" {
+		return filepath.Join(dir, "elemental-upgrade")
+	}
+	// Check persistent and if its mounted
+	persistent, err := GetFullDeviceByLabel(config.Runner, config.PersistentLabel, 5)
+	if err == nil && persistent.MountPoint != "" {
+		return filepath.Join(persistent.MountPoint, "elemental-upgrade")
+	}
+	return filepath.Join("/", "tmp", "elemental-upgrade")
```
davidcassany, line 247: nit: probably instead of using direclty `/tmp` as a fallback we could use the OS default or whatever uses `afero.TempDir` which will be `/tmp` in most cases.

davidcassany (approved):
Looks good! :+1:

pkg/utils/common.go
Itxaka: afero does something a bit weird with the dir, as it creates and chmods 777 to it for some reason....I did not understand why it does so, so I went with manual procedure.
Thing is os.TempDir() does basically the same as this method, checks the var (TMPDIR) or returns /tmp as fallback so it felt a bit weird calling it here after just doing the TMPDIR check stuff.

PR #107, opened 2022-02-16
Upgrade fixes for paths and mounts

Itxaka:
- Calculate the sqush recovery at the start in a more clear way
 - Calculate state mounts at the start and fill all the paths properly
at once
 - Several instances of not checking if we are in recovery
 - Always copy active to passive if we are updating active no matter
where we are
 - Set proper transitionImg based on where we boot from and sqush status
 - Mount state partition if not mounted before the grub rebrand,
otherwise we fail
 - Mount/unmonut img file based on recovery+squash
 - Create sqush file only on squash+recovery upgrade
 - Error out on SetDefaultGrubEntry if state mountpoint is empty, i.e.
state is not really mounted
 - Adapt tests to changes

### DIFF
--- a/pkg/action/action_test.go
+++ b/pkg/action/action_test.go
@@ -571,6 +571,9 @@ var _ = Describe("Actions", func() {
 				if command == "blkid" && args[0] == "--label" && args[1] == constants.StateLabel {
 					return []byte("/dev/active"), nil
 				}
+				if command == "blkid" && args[0] == "--label" && args[1] == constants.RecoveryLabel {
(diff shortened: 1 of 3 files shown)


Itxaka:
Manually tested this on all paths

 - Active from active
 - Active from recovery
 - Recovery from active
 - Recovery squash from active
 - Recovery from recover
 - Recovery squash from recovery

And it seems to finally work

davidcassany (approved):
Looks good

PR #108, opened 2022-02-16
upgrade: always try to remount state partition

Itxaka:
Rebrand works differently so we cant expect state to alwasy be mounted
or even mounted RW


davidcassany (approved):

Itxaka:
I noticed that I can directly test this branch from cos toolkit so I'm gonna leave the tests running to see if there is a missing path somewhere before merging anything

Issue #58, opened 2022-01-26
Create a device parsing library/package


davidcassany:
Related to #22 and #57. Would be nice considering some package in elemental to handle device discovery. This is mostly all the logic to find device paths by some of its attributes (filesystem label, partition label, partition number, etc.) in dynamic environments (just right after formatting or partition creation).

Itxaka:
Did this get covered by using ghw? or is it something that we are still missing @davidcassany ?

davidcassany:
Yes I think this should have been closed by https://github.com/rancher/elemental-cli/pull/155

PR #111, opened 2022-02-17
Implement yip's Layout plugin using partitioner package

davidcassany:
This commit makes use of the partitioner elemental package to provide a
custom implementation of the yip's layout plugin. This has the benefit
of adding support for MSDOS partition tables.

In order to make that happen few refactors were needed:

* Separate cloud-init-runner interface and implementation in separate
packages.
* Extend v1.Runner interface
* Separate v1.RunConfig initializer to a separate package


davidcassany:
@itxaka @mudler this the refactor I did in the scope of rancher-sandbox/cOS-toolkit#1075. If you remember this was about having cycling dependencies.

This PR moves cloud-init-runner  interface and its implementation to two different packages. The interface stays at `v1` (`pkg/types/v1`) and the implementation is at `cloudinit` (`pkg/cloudinit`, do you think `pkg/types/cloudinit` would be better?). In the same scope I also have to move `v1.NewRunConfig` method to another package (because it requires an implementation of cloud-init which in turn requires `v1`). I moved it to `config` (`pkg/config`, again, probably better `pkg/types/config`?). 

Since `plugins.Console` and `v1.Runner` are different and incompatible I also extended v1.Runner to make it flexible enought to implement `plugins.Console` by using `v1.Runner`. This really opens the door of running sophisticated tests for yip in elemental and also ensures a more consistent behavior. As a result of that I had to modify `v1.Runner` implementations and in that path I also killed the former `v1mock.FakeRunner` and renamed `v1mock.TestRunnerV2` to `v1mock.FakeRunner` this allows us to get rid of many runner redeclarations in tests. 

Another alternative could have been implementing the yip's layout plugin in yip directly (by importing elemental's partitioner), however this would have not prevented us from having to separate cloud-init implementation from `v1`. I believe rewriting yips plugin is something to think about so we have only a single implementation of the layout plugin, but I see no further benefits, in terms of refactoring requires more or less the same.

IMHO we could extend the above changes to other interfaces. In addition, I believe that `pkg/action/common.go` could be merged to the new `config` package, as those are simple methods to feed RunConfig based on action + config file/flags + hostchecks constraints.

Does makes sense to you? Any thoughts?

pkg/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,216 @@
...
+
+type YipCloudInitRunner struct {
+	exec    executor.Executor
+	fs      vfs.FS
+	console plugins.Console
+}
+
+// elementalCloudInitConsole represents a yip's Console implementations using
+// the elemental v1.Runner interface.
+type elementalCloudInitConsole struct {
```
mudler: Small nit, I'd suggest to move this out in it's own file, like `console.go` in this package, I'm not sure how much is untied from the code, but maybe can even be a separate package for cloud init functionalities

pkg/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Useful for testing purposes
+func (ci *YipCloudInitRunner) SetFs(fs vfs.FS) {
+	ci.fs = fs
+}
+
+// layoutPlugin is the elemental's implementation of Layout yip's plugin based
+// on partitioner package
+func layoutPlugin(l logger.Interface, s schema.Stage, fs vfs.FS, console plugins.Console) (err error) {
```
mudler: Nit: same here, this piece of logic seems to deserve its own place, like `plugin_layout.go`, or even a separate package (e.g. `plugin.Layout`)

pkg/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,216 @@
...
+	if s.Layout.Device == nil {
+		return nil
+	}
+
+	var dev *partitioner.Disk
+	elemConsole, _ := console.(*elementalCloudInitConsole)
```
mudler: Nit: I'm ok with this, but note that this breaks easily on subsequent refactors when you are mutating structs, and tests fails horribly and makes debugging tricky

pkg/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,216 @@
...
+	runner := elemConsole.getRunner()
+	log, _ := l.(v1.Logger)
```
mudler: Nit: same as above :arrow_up:

pkg/cloudinit/cloudinit.go
davidcassany: I fully agree on a separate file, however I am not sure if a separate package, this is highly couple to yip and I think all this should be constrained in `cloudinit` package (or eventually rename it to `yip` package).

pkg/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,216 @@
...
+
+		// Set default filesystem
+		if part.FileSystem == "" {
+			part.FileSystem = constants.LinuxFs
+		}
+
+		l.Infof("Creating %s partition", part.FSLabel)
+		partNum, err := dev.AddPartition(part.Size, part.FileSystem, part.PLabel)
+		if err != nil {
+			l.Error("Failed creating partition")
```
mudler: Nit: I'm not sure if we should wrap here instead, we might be duplicating error messages as those would be printed out anyhow

mudler (approved):
Just few small nits, but looks good here!

pkg/cloudinit/cloudinit.go
davidcassany: I know, couldn't think of any other method. Note I am not checking the error since this is all private, so there can't be a `console` that is not of type `elementalCloudInitConsole` in `layoutPlugin` method. At first I added an error check here but noted that it was not possible to reach this error. 

Also this crap is a motivation of having the yip console and the yip plugin implemented within the same package, so all the logic can be private.

But you got a point that having this explicit check could help on refactors down the line. I can add a check for that.

pkg/cloudinit/cloudinit.go
davidcassany: Yes we can wrap, this was mostly a copy & paste from yip's layout, it can be improved.

mudler:
> @Itxaka @mudler this the refactor I did in the scope of [rancher-sandbox/cOS-toolkit#1075](https://github.com/rancher-sandbox/cOS-toolkit/issues/1075). If you remember this was about having cycling dependencies.
> 
> This PR moves cloud-init-runner interface and its implementation to two different packages. The interface stays at `v1` (`pkg/types/v1`) and the implementation is at `cloudinit` (`pkg/cloudinit`, do you think `pkg/types/cloudinit` would be better?). In the same scope I also have to move `v1.NewRunConfig` method to another package (because it requires an implementation of cloud-init which in turn requires `v1`). I moved it to `config` (`pkg/config`, again, probably better `pkg/types/config`?).

I'm in favor of this, ideally, we could have a `pkg/core` for core modules that are leveraging types et al. but makes sense as well having them in the top dir, I don't mind at all as the others are scoped to their own dirs

> 
> Since `plugins.Console` and `v1.Runner` are different and incompatible I also extended v1.Runner to make it flexible enought to implement `plugins.Console` by using `v1.Runner`. This really opens the door of running sophisticated tests for yip in elemental and also ensures a more consistent behavior. As a result of that I had to modify `v1.Runner` implementations and in that path I also killed the former `v1mock.FakeRunner` and renamed `v1mock.TestRunnerV2` to `v1mock.FakeRunner` this allows us to get rid of many runner redeclarations in tests.

:+1: 

> Another alternative could have been implementing the yip's layout plugin in yip directly (by importing elemental's partitioner), however this would have not prevented us from having to separate cloud-init implementation from `v1`. I believe rewriting yips plugin is something to think about so we have only a single implementation of the layout plugin, but I see no further benefits, in terms of refactoring requires more or less the same.
> 

Indeed, I don't see the need at the moment so it looks good here, but we could think about it in the context of https://github.com/rancher-sandbox/elemental/issues/58  as we could have a library also for partitioning that could be shared from both codebases.

mudler (approved):
Looking good! nice! :100:

Itxaka (approved):
If I understand correctly this also means that we can bring our own yip plugins here directly, no need to change them upstream on yip rigth? Like the layout one. That is nice!

Itxaka:
> This PR moves cloud-init-runner interface and its implementation to two different packages. The interface stays at v1 (pkg/types/v1) and the implementation is at cloudinit (pkg/cloudinit, do you think pkg/types/cloudinit would be better?). In the same scope I also have to move v1.NewRunConfig method to another package (because it requires an implementation of cloud-init which in turn requires v1). I moved it to config (pkg/config, again, probably better pkg/types/config?).

Agreed with Ettore here, I dont care where they are as long as they are scoped, it works.

Itxaka:
OMG you are lowering our coverage a whooping 0.06%!!!! now that it was finally green!!!!111!!! PR REJECTED !!"32321!!!! :p

Itxaka:
> IMHO we could extend the above changes to other interfaces. In addition, I believe that pkg/action/common.go could be merged to the new config package, as those are simple methods to feed RunConfig based on action + config file/flags + hostchecks constraints.

Agreed on this, they are coupled and AFAIK they are mostly using constants (and some v1 which should be moved to constants anyhow, v1.BIOS, v1.MSDOS, etc...) so they are only coupled to the config struct anyway and should not pull anything else from other packages.

The hook stuff is still using utils so maybe those ones should be left there?

mudler:
> OMG you are lowering our coverage a whooping 0.06%!!!! now that it was finally green!!!!111!!! PR REJECTED !!"32321!!!! :p

PR #112, opened 2022-02-18
Ensure cloudinit tests can be executed in parallel

davidcassany:


davidcassany:
I noticed some random errors on cloudinit tests, I believe this was caused because they were not safe to be executed in parallel, as all of them were creating and destroying the same statically named tempfile. With change each test creates its own random tempfile so there are not collisions.

Itxaka (approved):
Good catch

PR #114, opened 2022-02-21
Define all needed images in runconfig

davidcassany:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -24,6 +24,7 @@ import (
 	"github.com/rancher-sandbox/elemental/pkg/partitioner"
 	"github.com/rancher-sandbox/elemental/pkg/types/v1"
 	"github.com/rancher-sandbox/elemental/pkg/utils"
+	"github.com/spf13/afero"
 	"path/filepath"
 )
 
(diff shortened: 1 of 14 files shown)


davidcassany:
This is a refactor aligned with rancher-sandbox/cos-toolkit#1133 and the natural follow up of #95. The main idea is to define images (passive, active, recovery, etc.) in a similar way as we do for partitions in runconfig. This provides some further flexibility to consider features such setting each image from a specific independent source (docker, iso, channel, etc.). This handy on #1133 scope to consider full installations not running from an installation media (ISO). Currently the recovery image is always taken from an ISO. With this PR adjusting it is as simple as setting the source type to the recovery image.

pkg/types/v1/common.go
```diff
@@ -18,9 +18,49 @@ package v1
 
 // ImageSource represents the source from where an image is created for easy identification
 type ImageSource struct {
-	Source    string
-	IsDir     bool
-	IsChannel bool
-	IsDocker  bool
-	IsFile    bool
+	source    string
```
davidcassany, line 21: The changes on this object are just done to force each source has a consistent type associated with it. I found while testing that I was reusing the same souce instance reusing type without a proper clean up, leading to situations where two types defined simultaneously. Since this is easy to programmatically force it by setting all attributes private I went for it.

pkg/action/install.go
```diff
@@ -38,36 +39,87 @@ func installHook(config *v1.RunConfig, hook string, chroot bool) error {
...
+		recoveryImg.File = filepath.Join(recoveryDirCos, cnst.RecoveryImgFile)
+		recoveryImg.Source = v1.NewFileSrc(activeImg.File)
+		recoveryImg.FS = cnst.LinuxImgFs
+		recoveryImg.Label = config.SystemLabel
+	}
+
+	// Add images to config
+	config.Images.SetActive(&activeImg)
+	config.Images.SetPassive(&passiveImg)
+	config.Images.SetRecovery(&recoveryImg)
```
davidcassany, line 112: The whole idea of this PR is that any of these images could easily be defined from an independent source (docker, channel, an image file, a directory...).

pkg/action/install.go
```diff
@@ -38,36 +39,87 @@ func installHook(config *v1.RunConfig, hook string, chroot bool) error {
...
+	}
+
+	// Set recovery image
+	partRecovery := config.Partitions.GetByName(cnst.RecoveryPartName)
+	if partState == nil {
+		config.Logger.Errorf("Recovery partition not configured")
+		return errors.New("Error setting Recovery image")
+	}
+
+	// TODO use iso for all images? Formerly it was only for recovery, but
```
mudler, line 88: yes, it looks some kinda of regression to me indeed. At least in the os2 context it is implied that the install involves all the images, not only recovery.

davidcassany: Yes this was so weird...

pkg/elemental/elemental.go
```diff
@@ -414,7 +415,7 @@ func (c *Elemental) SelinuxRelabel(rootDir string, raiseError bool) error {
 func (c *Elemental) CheckNoFormat() error {
 	c.config.Logger.Infof("Checking no-format condition")
 	// User asked for no format, lets check if there are already those labeled file systems in the disk
-	for _, label := range []string{c.config.ActiveImage.Label, c.config.PassiveLabel} {
+	for _, label := range []string{c.config.Images[cnst.ActiveImgName].Label, c.config.PassiveLabel} {
```
mudler: nit: This is a bit unsafe here, I know that the active will be always present there, but would break easily also from testing

pkg/elemental/elemental.go
```diff
@@ -278,7 +278,7 @@ func (c Elemental) CreateFileSystemImage(img *v1.Image) error {
 func (c *Elemental) DeployImage(img *v1.Image, leaveMounted bool) error {
 	var err error
 
-	if !img.Source.IsFile {
+	if !img.Source.IsFile() {
 		//TODO add support for squashfs images
```
davidcassany, line 283: This is not yet needed with the current feature set. But it will be as long as we allow recovery images to come from different source types.

mudler (approved):
Looking good!

pkg/elemental/elemental.go
davidcassany: Yes I know, in general I assumed in `pkg/elemental.go` we have already sanitized data, but it is true that in some places it probably assuming too much. It is hard to balance, at a certain point it turns to be insane to always check for nil in all map accesses.

Anyway, here we can use `c.config.ActiveLabel` instead, which is clearly safer :+1:

PR #115, opened 2022-02-21
Fixups to upgrade process

mudler:
First commit enables more detailed logs when rsync fails, attaching them to the error, the second one actually adds `hosts` to the excludes which was missing https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/backports/installer/cos.sh#L974

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -117,7 +117,7 @@ func CopyFile(fs afero.Fs, source string, target string) (err error) {
 
 // Copies source file to target file using afero.Fs interface
 func CreateDirStructure(fs afero.Fs, target string) error {
-	for _, dir := range []string{"sys", "proc", "dev", "tmp", "boot", "usr/local", "oem"} {
+	for _, dir := range []string{"run", "sys", "proc", "dev", "tmp", "boot", "usr/local", "oem"} {
 		err := fs.MkdirAll(fmt.Sprintf("%s/%s", target, dir), 0755)
 		if err != nil {
 			return err
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -18,18 +18,19 @@ package utils
 
 import (
 	"encoding/json"
-	"errors"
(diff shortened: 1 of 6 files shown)


mudler:
This seems to fix upgrades on kubernetes https://github.com/rancher-sandbox/cOS-toolkit/issues/1136#issuecomment-1046788494

davidcassany (approved):

davidcassany:
Looks like there is a problem with the go.mod and go.sum

pkg/utils/common.go
```diff
@@ -18,18 +18,19 @@ package utils
...
-	"github.com/zloylos/grsync"
 	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/joho/godotenv"
+	"github.com/pkg/errors"
```
davidcassany: In other parts of the code we are using the multierror library from hashicorp, I think it would be good trying to be consistent and stick to one. I have no preference over one against the other, but I think there is no need for two different error third party libraries.

mudler: take in mind they do a different thing, but can likely replace it with a fmt.Errorf

davidcassany: Use it as you see fit, I mean, this is very minor detail. I was just surprised of seeing another error library, I am convinced it does not hurt, to me it just felt a bit like using different logging libraries within the same project 😅

mudler: there is really no gain except syntax sugar in this case (and my usual habits :sweat_smile: ), I've went back to Errorf in https://github.com/rancher-sandbox/elemental/pull/115/commits/e501baf9d45c91f8645c2c4dc8b70697c90b1f25

mudler:
> Looks like there is a problem with the go.mod and go.sum

yup, looks like an unwanted effect from `go mod tidy`. Looks something is not ok in the ginkgo test setup

PR #117, opened 2022-02-24
Increase shared flags between install, upgrade and reset

davidcassany:
This ensures upgrade, reset and install share `--directory`, `--docker-image` and related flags.


davidcassany:
I believe we could extend it to `--iso`, so reset and upgrade also make use of it. I have not done that since for the upgrade case this could be intrusive and probably it deserves its own devoted PR.

mudler (approved):
Looks good here!

PR #120, opened 2022-02-25
Use io.ReadAll to get the ISO

kkaempf:
resp.Body.Read(source) would only read up to the size of source, which is zero

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -19,6 +19,7 @@ package elemental
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -520,7 +521,7 @@ func (c *Elemental) GetUrl(url string, destination string) error {
 		if err != nil {
 			return err
 		}
-		_, err = resp.Body.Read(source)
+		source, err = io.ReadAll(resp.Body)
 		defer resp.Body.Close()
 	default:
 		c.config.Logger.Infof("Copying from %s to %s", url, destination)


kkaempf:
Just tested, installation works now.

We could also revert the `loop` option to mount :wink:

mudler (changes requested):
As this function might be used to download ISOs too which are big files, io.ReadAll is not safe. It would load and slurp the file content in memory, which will likely break on embedded devices.

We should or refactor GetUrl here, or have a separate one for big files. My suggestion is to not re-implement the wheel and go for something already specific for big files as https://github.com/cavaliergopher/grab .

kkaempf:
Or spawn a `curl` process ?! :wink:

davidcassany:
Yes I agree we should not use ReadAll for ISOs or any potentially big file. I would not distinguish cases between big or small files, I would just implement a safe download utility (using some library, `curl` or whatever) and just use it regardless the file type we are downloading. I just noted that the local file copy is also loading the whole file at once on memory, this needs to be changed too.

I am having a look

mudler:
> Or spawn a `curl` process ?! wink

that would defeat entirely the purpose of a golang bin!

> Yes I agree we should not use ReadAll for ISOs or any potentially big file. I would not distinguish cases between big or small files, I would just implement a safe download utility (using some library, `curl` or whatever) and just use it regardless the file type we are downloading. I just noted that the local file copy is also loading the whole file at once on memory, this needs to be changed too.
> 
> I am having a look

I'm suggesting [grab](https://github.com/cavaliergopher/grab) for specific features that in the long-run I think we are going to implement or face in a way or another, such as: 

```
    Monitor download progress concurrently
    Auto-resume incomplete downloads
    Guess filename from content header or URL path
    Safely cancel downloads using context.Context
    Validate downloads using checksums
    Download batches of files concurrently
    Apply rate limiters
```

which are, extremely useful when downloading large files. Consider we already passively import it as `luet` uses it to download packages

davidcassany:
> that would defeat entirely the purpose of a golang bin!

To be fair we are already shelling out tons of different utilities... elemental is far from being golang native.

> I'm suggesting [grab](https://github.com/cavaliergopher/grab) for specific features that in the long-run I think we are going to implement or face in a way or another

Yes, makes sense. It should not be that difficult to implement our current interface using this library.

davidcassany:
#121 is already merged fixing this issue, I think we can close this one.

PR #121, opened 2022-02-28
Use grab library as the http client

davidcassany:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ ifneq ($(shell id -u), 0)
 	@echo "This tests require root/sudo to run."
 	@exit 1
 else
-	ginkgo run --label-filter root --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage_root.txt -p -r ${PKG}
+	ginkgo run --label-filter root --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage_root.txt -procs=1 -r ${PKG}
 endif
 
 # Useful test run for local dev. It does not run tests that require root and it does not run tests that require systemctl checks
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -19,10 +19,10 @@ package config
 import (
 	"github.com/rancher-sandbox/elemental/pkg/cloudinit"
 	cnst "github.com/rancher-sandbox/elemental/pkg/constants"
+	"github.com/rancher-sandbox/elemental/pkg/http"
 	"github.com/rancher-sandbox/elemental/pkg/types/v1"
 	"github.com/spf13/afero"
 	"k8s.io/mount-utils"
-	"net/http"
 )
 
 func NewRunConfig(opts ...v1.RunConfigOptions) *v1.RunConfig {
@@ -32,7 +32,7 @@ func NewRunConfig(opts ...v1.RunConfigOptions) *v1.RunConfig {
 		Logger:  log,
 		Runner:  &v1.RealRunner{},
 		Syscall: &v1.RealSyscall{},
-		Client:  &http.Client{},
+		Client:  http.NewClient(),
 	}
 	for _, o := range opts {
 		err := o(r)
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	uri "net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -251,4 +252,45 @@ func GetUpgradeTempDir(config *v1.RunConfig) string {
 		return filepath.Join(persistent.MountPoint, "elemental-upgrade")
 	}
 	return filepath.Join("/", "tmp", "elemental-upgrade")
+}
+
+// IsLocalUrl returns true if the url has no scheme or it has "file" scheme
+// and returns false otherwise. Error is not nil if the url can't be parsed.
+func IsLocalUrl(url string) (bool, error) {
+	u, err := uri.Parse(url)
+	if err != nil {
+		return false, err
+	}
+	if u.Scheme == "file" || u.Scheme == "" {
+		return true, nil
+	}
+	return false, nil
+}
+
+// GetSource copies given source to destination, if source is a local path it simply
+// copies files, if source is a remote URL it tries to download URL to destination.
+func GetSource(config *v1.RunConfig, source string, destination string) error {
+	local, err := IsLocalUrl(source)
+	if err != nil {
+		config.Logger.Errorf("Not a valid url: %s", source)
+		return err
+	}
+
+	err = config.Fs.MkdirAll(filepath.Dir(destination), os.ModeDir)
+	if err != nil {
+		return err
+	}
+	if local {
+		u, _ := uri.Parse(source)
+		err = CopyFile(config.Fs, u.Path, destination)
+		if err != nil {
+			return err
+		}
+	} else {
+		err = config.Client.GetUrl(config.Logger, source, destination)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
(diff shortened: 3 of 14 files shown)


davidcassany:
Supersedes #120 to start using the grab library. I don't like we are not able to make use the of the afero filesystem with the grab library, but I guess this is a reasonable price to pay to get other features.

It also completely redefines the HTTPClient interface. I saw it was used basically to fetch the remote ISO or CloudConfig in the very same way, so I thought about creating an interface with a single method to download file to destination with the parameters that were needed for this use case. We can simply extend it as long as we need more features or flexibility.

davidcassany:
@mudler I am not sure why, but I started to face issues on test labeled for root, basically the upgrade tests. I started to wonder if those should be executed in parallel or not, since they call luet to install stuff. I assume luet has some sort of mutex, so running tests that make use of luet installation procedures in parallel seams to be a bad idea, what I can't figure out is why now started to be a problem and not before. I got ginkgo's timeout complaints, probably the issue just becomes really problematic when performance is bad? gingko complaints about not getting response of one of its processes.

[Here](https://github.com/rancher-sandbox/elemental/runs/5365619432?check_suite_focus=true) is the issue I see if I do not limit parallel tests for root tests:

```
** Ginkgo timed out waiting for all parallel procs to report back. **
action (./pkg/action)
Output from proc 1:
  PASS
  coverage: 30.3% of statements

Output from proc 2:

** End **ginkgo run failed
  Failed to combine cover profiles
  Unable to read coverage file /home/runner/work/elemental/elemental/pkg/action/coverage_root.txt.2:
  open /home/runner/work/elemental/elemental/pkg/action/coverage_root.txt.2: no such file or directory
 ```

**Comment by mudler** [comment]
> @mudler I am not sure why, but I started to face issues on test labeled for root, basically the upgrade tests. I started to wonder if those should be executed in parallel or not, since they call luet to install stuff. I assume luet has some sort of mutex, so running tests that make use of luet installation procedures in parallel seams to be a bad idea, what I can't figure out is why now started to be a problem and not before. I got ginkgo's timeout complaints, probably the issue just becomes really problematic when performance is bad? gingko complaints about not getting response of one of its processes.
> 
> [Here](https://github.com/rancher-sandbox/elemental/runs/5365619432?check_suite_focus=true) is the issue I see if I do not limit parallel tests for root tests:
> 
> ```
> ** Ginkgo timed out waiting for all parallel procs to report back. **
> action (./pkg/action)
> Output from proc 1:
>   PASS
>   coverage: 30.3% of statements
> 
> Output from proc 2:
> 
> ** End **ginkgo run failed
>   Failed to combine cover profiles
>   Unable to read coverage file /home/runner/work/elemental/elemental/pkg/action/coverage_root.txt.2:
>   open /home/runner/work/elemental/elemental/pkg/action/coverage_root.txt.2: no such file or directory
> ```

That's typically happening when there are some goroutine hanging, actually not sure if this is caused by luet, because you can actually run it in parallel (although, the cli will prevent that to not cause system inconsistencies due to dep solving ) 

One thing to notice, but I only gave it a quick look, is that if we don't instantiate a http.client with a timeout, the default is without, which means a potential http call could hang forever.

I'll give a throughout review tomorrow, but for the moment looks good

**Comment by mudler** [comment]
> Supersedes #120 to start using the grab library. I don't like we are not able to make use the of the afero filesystem with the grab library, but I guess this is a reasonable price to pay to get other features.

On afero: huh, yet another reason why I remember replacing it, as IIRC doesn't have a clean way to get the underlying path and implement a wrapper around it. As it also clashes with the standard fs interface, makes things more convoluted and force using it all the way down.

> 
> It also completely redefines the HTTPClient interface. I saw it was used basically to fetch the remote ISO or CloudConfig in the very same way, so I thought about creating an interface with a single method to download file to destination with the parameters that were needed for this use case

👍

File: Makefile [hunk 1]
```diff
@@ -47,7 +47,7 @@ ifneq ($(shell id -u), 0)
 	@echo "This tests require root/sudo to run."
 	@exit 1
 else
-	ginkgo run --label-filter root --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage_root.txt -p -r ${PKG}
+	ginkgo run --label-filter root --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage_root.txt -procs=1 -r ${PKG}
```
Itxaka, line 50: leftover?

pkg/utils/common.go
```diff
@@ -252,3 +253,44 @@ func GetUpgradeTempDir(config *v1.RunConfig) string {
 	}
 	return filepath.Join("/", "tmp", "elemental-upgrade")
 }
+
+// IsLocalUrl returns true if the url has no scheme or it has "file" scheme
+// and returns false otherwise. Error is not nil if the url can't be parsed.
```
Itxaka, line 258: why aren't we returning a proper err if this errors out? We cant be inconsistent, either there is an error and we return it or we don't, otherwise we need to be checking the error and the return value in order to know what happened no? Error out if there was an error IMO.

Itxaka (commented):
couple of nits

Makefile
davidcassany: No, for some reason I can't figure out root tests started to fail. It looks like some sort of ginkgo timeout while running upgrade from channel tests in parallel. I am not, fully convinced what is going on there, I thought that probable there is an issue about running luet install in parallel over the same host? Not sure, but forcing sequential execution on root tests seams to fixed it. I also don't understand why it started to fail in this PR, it looks unrelated to the PR... 🤷🏽‍♂️

pkg/utils/common.go
davidcassany: Yes, this is also an option, however you can't distinguish invalid URLs from remote ones in that case. I thought this could be also eventually used to sanitize URLs and/or tests remote urls. But yes, we can eventually have a IsLocalURl method and if needed another IsRemoteUrl method.

Makefile
Itxaka: blargh :/

Itxaka (approved):

pkg/utils/common.go
davidcassany: Note you can ignore the error with this method, you could eventually use:

```golang
if IsLocalUrl("someurl") {
    ...
}
```
It only returns `true` if the URL is local. If it returns `false` it is not a local url and it is eventually an invalid URL if err is not nil.

davidcassany:
> One thing to notice, but I only gave it a quick look, is that if we don't instantiate a http.client with a timeout, the default is without, which means a potential http call could hang forever.

@mudler I guess you mean initiating the requests with some specific context, right? I saw the context settings and I did not really understand how to use it, it is unclear to me what this timeout means, is it timeout to get a response? a timeout to finalize the response handling (finalize download)? Without this timeout in which context is falling into an infinite loop?

pkg/config/config.go
```diff
@@ -32,7 +32,7 @@ func NewRunConfig(opts ...v1.RunConfigOptions) *v1.RunConfig {
 		Logger:  log,
 		Runner:  &v1.RealRunner{},
 		Syscall: &v1.RealSyscall{},
-		Client:  &http.Client{},
+		Client:  http.NewClient(),
```
mudler, line 35: We need an `http.Client` here by ourselves with a http timeout. Keep in mind that without a timeout explictly set, golang goes foobar, see e.g. https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779 

Grab is just doing the same, but setting up a user agent text field for it: https://github.com/cavaliergopher/grab/blob/v2.0.0/client.go#L37

mudler: That might have been causing also problems with upgrade tests if ends hitting up this path: grab spawns a goroutine and waits for the request context, which in this case could hang indefinitely

kkaempf:
Confirmed working 💚

PR #122, opened 2022-03-01
Run parallel tests

davidcassany:


davidcassany:
This PR is just to try to reproduce and experiment with some flakyness on tests that seams to be related with parallel execution of specs.

mudler:
we might have some goroutine leak or either some context not properly closed which is tied to something in the background (which is still sitting there after tests ended). Maybe checking pprof/goroutines would help out to nail it down.

davidcassany:
Yes, I did not invest time anymore on that as with this PR I just wanted to ensure this was a real issue and not something I saw transitory while fixing the ISO download. My idea was to trying to identify the offending tests by reducing the executed ones, so start to check if with `procs=2` is still failing (I think we usually get 2 processes) and from there start reducing the tests and check if I can get it to fail with only 2 specs.

davidcassany:
Ok, as I expected this issue on channels tests is unrelated with #121, where we first faced it. So there is no connection with the httpclient, which makes sense as the upgrade action does not make use of the httpclient in any way.

There are three tests of the upgrade channel, no matter which two we run concurrently that it fails, there is the one going to process 2 that hangs. Process 1 succeeds, I assume it matches the order of the execution.

It feels like it could be something related to Luet, but I can't see what changed that caused these tests to start failing....

davidcassany:
I could not find the root cause of this issue. I have two ideas, without good arguments though:

* There is a lack of resources to run upgrade from channel tests in parallel: This could explain why it suddenly started to fail, probably we were already around its limits and by some cOS update it starts to consume more or caused by some update on the underlaying VM there are less resources available.
* Simultaneous `luet install`  commands fail to execute: at a glance I could not see anything that could explain it as all the paths are configured on per test basis (database, tmpdir, cache, etc). However I saw that few luet temporary files are not encompassed by this. There is a `$TMPDIR/tmpluet` folder dynamically created in each test, I think this is related to the garbase collector and it looks like there is no system wide information stored in there as each run creates a random folder under it and it does not look like there can be collisions.

I any case I believe not being capable of running elemental upgrade tests in parallel is a big issue, since this is a use case that only happens on tests. In real operation is doesn't make sense running multiple `elemental upgrade` on the same host in parallel.

So that I am closing this PR and leaving the parallel upgrade tests issue unsolved.

PR #123, opened 2022-03-02
Use golangci for lint+license check

Itxaka:
Also fixes the 100 or so lint issues


pkg/action/install.go
```diff
@@ -50,13 +51,14 @@ func InstallImagesSetup(config *v1.RunConfig) error {
...
+		return errors.New("error setting Active image")
 	}
 	// Set active image
 	activeImg := v1.Image{
-		Label:      config.ActiveLabel,
-		Size:       cnst.ImgSize,
-		File:       filepath.Join(partState.MountPoint, "cOS", cnst.ActiveImgFile),
+		Label: config.ActiveLabel,
+		Size:  cnst.ImgSize,
+		// TODO: partState.MountPoint can be nil? We are checking for nil below on line 84...
```
davidcassany: There is no such a TODO, it is just checked on previous statement, line 52, partState is not nil here, it can't segfault and MountPoint is a string so it can't be nil.

pkg/action/install.go
```diff
@@ -72,17 +74,18 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 
 	// Set passive image
 	passiveImg := v1.Image{
-		File:   filepath.Join(partState.MountPoint, "cOS", cnst.PassiveImgFile),
+		// TODO: partState.MountPoint can be nil? We are checking for nil below on line 84...
```
davidcassany: Same as a above

pkg/partitioner/disk.go
```diff
@@ -364,7 +361,9 @@ func (dev Disk) expandFilesystem(device string) (string, error) {
 	case "xfs":
 		// to grow an xfs fs it needs to be mounted :/
 		tmpDir, err := afero.TempDir(dev.fs, "", "yip")
-		defer dev.fs.RemoveAll(tmpDir)
+		defer func(fs afero.Fs, path string) {
+			_ = fs.RemoveAll(path)
+		}(dev.fs, tmpDir)
```
davidcassany, line 366: I guess the linter is takes being explicit to a hardcore level 😅

davidcassany (approved):
Nice looks good!

I only would remove a couple of TODOs, there can't be a nil de reference there.

pkg/utils/grub.go
```diff
@@ -116,7 +116,12 @@ func (g Grub) Install() (err error) {
 	}
 
 	grubConfTarget, err := g.config.Fs.Create(fmt.Sprintf("%s/grub.cfg", grubdir))
-	defer grubConfTarget.Close()
+	if err != nil {
+		return err
+	}
+	defer func(grubConfTarget afero.File) {
```
davidcassany, line 122: I guess it is a linter request passing variables as arguments instead of taking it from scope, right?

pkg/action/install.go
Itxaka: ah I see. Any idea why is checked also afterwards on line 84? is that wrong and should refer to `partRecovery` ?

pkg/partitioner/disk.go
Itxaka: yeah, I wish there was a way of disabling some of this stuff. Like I get it, its important to get your return vars and do something with them explicitly, but on some places we dont really care if it fails...

And I dont like setting the nolint stuff everywhere :/

pkg/utils/grub.go
Itxaka: line 122 is about not getting the return values from `grubConfTarget.Close()`

line 129 is that the err var from line 118 was not being used, and indeed it wasn't checked after getting it, just overwritten below on 138 and then checked there.

pkg/action/install.go
davidcassany: Damn you are right, there is a copy & paste error there. Good catch!

pkg/action/install.go
```diff
@@ -72,17 +74,18 @@ func InstallImagesSetup(config *v1.RunConfig) error {
...
+		File:   filepath.Join(partState.MountPoint, "cOS", cnst.PassiveImgFile), // nolint:staticcheck
 		Label:  config.PassiveLabel,
 		Source: v1.NewFileSrc(activeImg.File),
 		FS:     cnst.LinuxImgFs,
 	}
 
 	// Set recovery image
 	partRecovery := config.Partitions.GetByName(cnst.RecoveryPartName)
-	if partState == nil {
+	if partState == nil { // nolint:staticcheck
```
davidcassany: This should be `if partRecovery == nil {`. @Itxaka do you mind changing it in this PR?

Itxaka: no probs!

PR #125, opened 2022-03-03
Check partial yaml unmarshal errors and unify FS

### LINKED ISSUE #1109: Hide yip unmarshall errors in elemental
This is related to the `yaml: unmarshal errors: line1: cannot unmarshal !!str /usr/lo... into schema.YipConfig` messaged we see on elemental during the installation. Despite they are not relevant in that case and they can be ignored, they really give the impression something went wrong. We should hide them as debug messages or find some smarter criteria to weather show them or not.

This will be a problem as soon as harvester starts using the new elemental installer, since harvester installer interface has a frame where the stdout of the installation is shown and these are almost the last messages shown while the harvester installer is stuck downloading some harvester from registries. It really gives the impression something is broken 😓 
image

or 
image

mudler:
This is what I had in mind for https://github.com/rancher-sandbox/cOS-toolkit/issues/1109

Fixes: https://github.com/rancher-sandbox/cOS-toolkit/issues/1109

Apart from fixing https://github.com/rancher-sandbox/cOS-toolkit/issues/1109, this PR is multi-fold as unifies the FS interface in elemental by adopting vfs in place of AferoFS, due to the following:
- aferoFS breaks the standard FS interface by returning a custom struct: https://pkg.go.dev/io/fs#FS - this have utterly bad consequences down in the code, which in turn makes us create new code that can't be used by consumers of such interface
- aferoFS doesn't allow a clean way to read the underlying system path. While this is available for basepath, it is not streamlined to the interface, thus IF someone wants to convert something to aferoFS it forces to type casting all the way down
- several logical functions that makes sense (Lchown, Lstat) are not available on aeroFS and, since there is no clean way to access the underlying system path, makes the implementation around it convoluted ( to say at least )
- Mixture of 2 fs was problematic, expecially for the change at hand: the RunStage func was using afero, vfs, and the underlying system OS alltogether. This made almost impossible to test the code without going for convoluted hacks

Besides, It introduces a `v1.FS` interface to unify the code around FS logics. it tries to refactor some code areas where was possible, but I would go over with a specific PR to unify all the remaining code part bound to the system OS. vfs as exposes RawPath directly, makes possible to convert any kinda of code to use a fs implementation.


davidcassany (approved):
Looks like a nice approach to me

Itxaka:
wait, why would this change the afero to a vfs? how is that related to the change? either we change the whole elemental to use vfs or lets not mix 2 different virtual FS libraries please :pray:

pkg/utils/fs.go
```diff
@@ -0,0 +1,71 @@
+package utils
+
```
Itxaka, line 21: Maybe im not understanding this, but why would we create these methods ourselves when they are already supported in the usual VFS libs?

mudler: they aren't

mudler:
> wait, why would this change the afero to a vfs? how is that related to the change? either we change the whole elemental to use vfs or lets not mix 2 different virtual FS libraries please pray

that's my plan. Afero is really limited here and getting in the way so easily because it doesn't comply to anything standard as returns afero structures all the way down. It doesn't support even simple things like Lchown, Lchmod and so on.

mudler:
and - this also prevents me to add tests for this specific change too without having to cut very bad code

mudler:
> wait, why would this change the afero to a vfs? how is that related to the change? either we change the whole elemental to use vfs or lets not mix 2 different virtual FS libraries please pray

we do already mix them, the issue is that RunStage is half-fs aware which makes everything so terrible. it partially operates on the system FS directly, partially uses vfs bound on the os fs, and some part of it are tied to afero. What I want to do here is to unify to a single interface, then the implementation is really something that we can change anytime

Itxaka:
Ok, that sounds good to me. Could we Renae the PR then to match what is doing? I was very confused seeing all the changes here while expecting only the mkdir 🤣

mudler:
> Ok, that sounds good to me. Could we Renae the PR then to match what is doing? I was very confused seeing all the changes here while expecting only the mkdir rofl

yep I will, let me just get over this change subset to have an idea. I'd like to bring it back to 1 lib, but I might not make it in this changeset.

Note this PR also introduce skipping partial YAML unmarshal errors, it is not just about mkdir (and that's where I had issues with afero for testing :/)

mudler:
Bonus from vfs, since it exposes `RawPath(name string) (string, error)`, anything that is actually bound to os can be converted easily (also things that don't know about it, for e.g. Chroot). But I'd likely go for it in a separate PR.

At the moment struggling with tests, as vfs is a bit more picky on permissions

mudler:
I've been able to use only one fs now, just few tests are missing

mudler:
https://github.com/rancher-sandbox/elemental/pull/125/commits/07a536732636550cf6e04fc2e395559cb244ac80 is an example on how to make other underlying libs to be fs-aware

mudler:
@Itxaka is there a way to make the go header linter a bit less smart? It tries to match the header, but in case files have different license authors it fails miserably

mudler:
@davidcassany @Itxaka need another review, as this derailed from the original PR, to actually unify the FS interface into a single one

Itxaka:
> @Itxaka is there a way to make the go header linter a bit less smart? It tries to match the header, but in case files have different license authors it fails miserably

what do you mean by this? Author should be SUSE LLC no? At least for the files created by us.

mudler:
> > @Itxaka is there a way to make the go header linter a bit less smart? It tries to match the header, but in case files have different license authors it fails miserably
> 
> what do you mean by this? Author should be SUSE LLC no? At least for the files created by us.

yes and no... I've took 2 functions from afero (create temp file and temp dir) to bring feature parity with it and as afero is under apache licence, It is enough to add it as author in the file

EDIT: I was thinking to isolate those fn into a separate file so to apply only the license there, but I think would complain anyhow? would excluding files work for license checks too?

Itxaka:
> yes and no... I've took 2 functions from afero (create temp file and temp dir) to bring feature parity with it and as afero is under apache licence, It is enough to add it as author in the file

I see... Does the apache mentions who has to go first on the license headers? Cause we could just skip the check for the license text and only check for the `Copyright © {{year}} SUSE LLC` line....

go header is pretty bad to be honest, it assumes just one entity copyrighting the full code... kind of wanting to write a new one that allows for regex and checks the text before the first import stanza, 'cause if we add a `// build:linux` line it will also fail...

I think you can also start the code by setting a `//nolint:goheader` so the check is skipped for that file, which also would fit this, as I dont expect this to happen that many times.

mudler:
> > yes and no... I've took 2 functions from afero (create temp file and temp dir) to bring feature parity with it and as afero is under apache licence, It is enough to add it as author in the file
> 
> I see... Does the apache mentions who has to go first on the license headers? Cause we could just skip the check for the license text and only check for the `Copyright © {{year}} SUSE LLC` line....

In theory should go first as their code is older, but that file contains mostly function written by us, just two of them are attributable to afero..
 
> 
> go header is pretty bad to be honest, it assumes just one entity copyrighting the full code... kind of wanting to write a new one that allows for regex and checks the text before the first import stanza, 'cause if we add a `// build:linux` line it will also fail...

Argh I see, yeah that sucks a bit :/
> 
> I think you can also start the code by setting a `nolint:goheader` so the check is skipped for that file, which also would fit this, as I dont expect this to happen that many times.

Oh cool! completely oversight that in the golint-ci docs, gonna try!

Itxaka:
@mudler pkg/types/v1/fs.go is missing the full license also :P

pkg/cloudinit/cloudinit.go
```diff
@@ -32,7 +32,7 @@ type YipCloudInitRunner struct {
 
 // NewYipCloudInitRunner returns a default yip cloud init executor with the Elemental plugin set.
 // It accepts a logger which is used inside the runner.
-func NewYipCloudInitRunner(l v1.Logger, r v1.Runner) *YipCloudInitRunner {
+func NewYipCloudInitRunner(l v1.Logger, r v1.Runner, fs vfs.FS) *YipCloudInitRunner {
```
davidcassany, line 35: Cool! yes this was really bothering me.

davidcassany (approved):
Looks good to me :+1:

PR #126, opened 2022-03-05
runstage: create cloudinitpaths before passing them to yip runner

Itxaka:
So we avoid passing non.existing paths, which yip will try to parse as
a cloud-init config string and produce unmarshall error which are
indistinguishable from real unmarshall errors


mudler:
Note this is already included in https://github.com/rancher-sandbox/elemental/pull/125

Itxaka:
blergh

Issue #124, opened 2022-03-03
Some race conditions in our tests


Itxaka:
Looks like our rsync lib sometimes produces race conditions during tests: https://github.com/rancher-sandbox/elemental/runs/5405027256?check_suite_focus=true

davidcassany:
~mmm, it could be related to the errors on #122~

Edit: it is not related to #122

davidcassany:
Probably this is caused by https://github.com/zloylos/grsync/blob/3aefd55e3b8344bfb7c1fca5e02bfaa49055205c/task.go#L47-L64

To me it looks the rsync task execution does not wait for the stdout and stderr scanner to end before returning. So potentially the Task.Run could eventually end and the `t *Task` object could still be modified by the scanner go routines. It is hard to imagine a use case where it could harm, but I guess this is not `--race` analyzer business, it just reports that the task object could be silently modified after `task.Run()` returned, which could have unpredictable effects.

davidcassany:
In my env I can easily reproduce this issue by using the little script:

```bash
for i in {1..20}; do
  echo "Running test $i"
  ginkgo run --label-filter 'install' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -p -r ./pkg/action/ || break
done
```

It usually fails 1 out of 10, so with 20 I could not manage to succeed a single time.

I did a hack on `grsync` package to ensure `Task.Run()` waits for the go routines parsing stdout and the errors are simply gone. I executed the above with the patch several times I did not get a single failure.

I am preparing a PR to upstream with the patch.

mudler:
> In my env I can easily reproduce this issue by using the little script:
> 
> ```shell
> for i in {1..20}; do
>   echo "Running test $i"
>   ginkgo run --label-filter 'install' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -p -r ./pkg/action/ || break
> done
> ```
> 
> It usually fails 1 out of 10, so with 20 I could not manage to succeed a single time.
> 
> I did a hack on `grsync` package to ensure `Task.Run()` waits for the go routines parsing stdout and the errors are simply gone. I executed the above with the patch several times I did not get a single failure.
> 
> I am preparing a PR to upstream with the patch.

:+1: sounds good here, good catch! just a tip, ginkgo has `--until-it-fails` to run tests until a failure is observed, and `--repeat=N` to repeat it N times

[...]

PR #128, opened 2022-03-07
Fix variables initialization in tests

davidcassany:
This commit makes explicit initialization of the variables declared
in `Describe` sections of test specs. I realized that non initialized
variables in `BeforeEach` or explicitly in each tests might be shared
across tests causing erratic behaviors depending on how concurrency
is applied over the suite.

### DIFF
--- a/pkg/action/action_test.go
+++ b/pkg/action/action_test.go
@@ -80,6 +80,8 @@ var _ = Describe("Actions", func() {
 	Describe("Reset Setup", Label("resetsetup"), func() {
 		var lsblkTmpl, bootedFrom, blkidOut, label, cmdFail string
 		BeforeEach(func() {
+			label = ""
+			cmdFail = ""
 			fs.Create(constants.EfiDevice)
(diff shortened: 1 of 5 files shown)


pkg/action/action_test.go
```diff
@@ -142,6 +144,7 @@ var _ = Describe("Actions", func() {
 		var cmdFail, activeTree, activeMount string
 		var err error
 		BeforeEach(func() {
+			cmdFail = ""
```
davidcassany, line 147: I had expected that all these  `cmdFail, activeTree, activeMount ` to be set to the default value here, but apparently this is not the case. Before this change I noted  `cmdFail` was being set on tests where this variable is not set at all. I did a review of several tests suits and adapted few details here and there, but I think this group is the only one that was facing this issue.

davidcassany:
I found this issue while reviewing #124, but this is unrelated, it is a different bug.

mudler (approved):

PR #129, opened 2022-03-08
Add checkRoot helper for cmd and add it to root commands

Itxaka:
Some of the commands require root privs to run but currently they are
run without which could cause issues during the middle of the command
leaving the target system broken.

Now the commands that require root are checked on PreRun and fail if
they require privs and they are running without


davidcassany (approved):
Nice, looks good to me.

PR #130, opened 2022-03-08
Add CheckDeps to cmd

Itxaka:
To check for all deps before running a command

### DIFF
--- a/cmd/constants/constants.go
+++ b/cmd/constants/constants.go
@@ -14,29 +14,21 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package cmd
+package constants
 
-import (
-	"errors"
-	"os"
-	"os/exec"
-)
-
-// CheckRoot is a helper to return on PreRunE, so we can add it to commands that require root
-func CheckRoot() error {
-	if os.Geteuid() != 0 {
-		return errors.New("this command requires root privileges")
+func InstallDeps() []string {
+	return []string{
+		"rsync", "losetup",
+		"wipefs", "tune2fs",
+		"blkid", "lsblk",
+		"e2fsck", "resize2fs",
+		"mount", "umount",
+		"xfs_growfs", "udevadm",
+		"mksquashfs", "grub2-install",
+		"grub2-editenv",
 	}
-	return nil
 }
 
-// CheckDeps is used by cmd to check a list of deps for a command and fail if they don't exists
-func CheckDeps(deps []string) error {
-	for _, dep := range deps {
-		_, err := exec.LookPath(dep)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+func UpgradeDeps() []string {
+	return InstallDeps()
 }
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -21,6 +21,7 @@ import (
 	"os/exec"
 
 	"github.com/rancher-sandbox/elemental/cmd/config"
+	"github.com/rancher-sandbox/elemental/cmd/constants"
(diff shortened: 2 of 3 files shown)


Itxaka:
This si just a WIP, an example on how we could check deps for each cmd individually.

cmd/constants/constants.go
```diff
@@ -0,0 +1,34 @@
...
+package constants
+
+func InstallDeps() []string {
+	return []string{
+		"rsync", "losetup",
+		"wipefs", "tune2fs",
+		"blkid", "lsblk",
+		"e2fsck", "resize2fs",
+		"mount", "umount",
+		"xfs_growfs", "udevadm",
```
Itxaka, line 26: should probably mark the xfs and resize2fs/tune2fs/e2fsck as optional deps, and emit a warning saying that the dependencies for X were not found, you will find errors if your layout includes a system with X fs?

cmd/constants/constants.go
```diff
@@ -0,0 +1,34 @@
...
+		"mksquashfs", "grub2-install",
```
Itxaka, line 27: same here for the squashfs

Itxaka:
it is a bit flimsy as we can have some optional deps like xfs, but those may be required for the given path and Im not sure we can cover checking everything on the cmd part...and not even sure if this should be done on the cmd part even...

cmd/constants/constants.go
davidcassany: mmm I don't know, in general I hate soft dependencies, but if at some point we aim to build tiny OSs makes sense having optionals. Probably I'd not even show a warning and make our runner to check the command existence every time, so if command X is used and missing we get then a proper message about missing command.

This could go together with the idea we discussed a while ago about making our runner to always log the command and its output as debug messages. For that we need to include a logger in our Runner implementation. Then we can easily check the command existence there and log a proper message if command X is missing. Probably this is just enough for optional deps?

Just first ideas that came to my mind, I do not have a real strong opinion on that.

cmd/helpers.go
```diff
@@ -28,3 +29,14 @@ func CheckRoot() error {
 	}
 	return nil
 }
+
+// CheckDeps is used by cmd to check a list of deps for a command and fail if they don't exists
+func CheckDeps(deps []string) error {
+	for _, dep := range deps {
+		_, err := exec.LookPath(dep)
```
davidcassany, line 36: It does not make a big difference, but here I would use the `utils.CommandExists`.

davidcassany (approved):
Looks good to me

cmd/constants/constants.go
Itxaka: Well, this is mostly trying to advice the user on what it will need beforehand, like the root stuff, so we dont fail in the middle of a run, i.e. if we do the whole install and then we are missing grub...that is bad UX as we leave the target system in a bad state and we may return the user a bad/misleading error. So this is kind of a first try to avoid that and advice beforehand as much as we can so the user of the tool is not frustrated afterwards.

cmd/helpers.go
Itxaka: ah good one :D

cmd/constants/constants.go
```diff
@@ -0,0 +1,34 @@
...
+		"grub2-editenv",
```
mudler, line 28: tried to run an install from the docker image this weekend and `mkfs.vfat` is missing, I'd also add `mkfs.*` at the list

Itxaka:
will revisit down the line

PR #131, opened 2022-03-08
Add dockerfile

Itxaka:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+ARG GOLANG_IMAGE_VERSION=1.17-alpine
+ARG COSIGN_VERSION=1.4.1-5
+ARG LEAP_VERSION=15.4
+
+FROM quay.io/costoolkit/releases-green:cosign-toolchain-$COSIGN_VERSION AS cosign-bin
+
+
+FROM golang:$GOLANG_IMAGE_VERSION as elemental-bin
+ENV CGO_ENABLED=0
+WORKDIR /src/
+# Add specific dirs to the image so cache is not invalidated when modifying non go files
+ADD go.mod .
+ADD go.sum .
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.17-alpine AS golang
+FROM gcr.io/projectsigstore/cosign:v1.4.0 AS cosign-bin
```
mudler: I'd use our images for cosign here from the cos releases repo, they just contain the binary and it's tested and built by our CI

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+
+FROM golang as elemental-bin
+WORKDIR /src/
+COPY . /src/
+RUN CGO_ENABLED=0 go build -o /usr/bin/elemental
+
+FROM alpine AS elemental-alpine
```
mudler: we should stick with opensuse here imho (besides obvious reasons), we have less chance to have incompatibilities and failures due to binary which are misaligned in versions.

Dockerfile
Itxaka: Problem with that is our version scheme is not that nice with all the -number stuff :/

Dockerfile
Itxaka: Yep, this as mentioned is a PPC and wanted to see if it was possible to build a valid image from alpine as I know it's possible from an opensuse base.

After all the size difference is not that much, ends up about 50Mb bigger so..

Dockerfile
mudler: Why? it is just something like:

```
quay.io/costoolkit/releases-green:cosign-toolchain-1.4.1
```

mudler (approved):
Looks good here!

PR #132, opened 2022-03-09
add channel support to install action

### LINKED ISSUE #1160: Provide a replacement for `cos-deploy`
Currently we do not have replacement for `cos-deploy` command in elemental and I think this is the last piece missing before we can actually drop `package/installer/cos.sh`.

So far by looking at the old command code and docs this command is expected to do the following:

* Sets active and passive from channel
* Sets active and passive from a docker image if this is provided with the `--docker-image` flag
* Does all the above regardless being booted from active, passive or recovery

If we add the `channel` source to `element reset` and we assume we only support this feature set when booting from recovery (this is the only use case we documented), then `elemental reset` would could be a drop in replacement to `cos-deploy`.

Itxaka:
Partially fixes https://github.com/rancher-sandbox/cOS-toolkit/issues/1160

Cant be this easy...can it?

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -80,7 +80,7 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 	// Error out if we could not find the source
 	if activeImg.Source.Value() == "" {
 		config.Logger.Error("Could not find source for the install")
-		return errors.New("could not find source for the install")
+		return &v1.SourceNotFound{}
 	}
 
 	// Set passive image
@@ -125,9 +125,14 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 // the given configuration flags
 func InstallSetup(config *v1.RunConfig) error {
 	SetPartitionsFromScratch(config)
-	_ = InstallImagesSetup(config)
-	SetupLuet(config)
+	err := InstallImagesSetup(config)
 
+	// Only error out if we can't find the source
+	switch e := err.(type) {
+	case *v1.SourceNotFound:
+		return e
+	}
+	SetupLuet(config)
 	return nil
 }
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -65,20 +65,18 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 		activeImg.Source = v1.NewDockerSrc(config.DockerImg)
 	} else if config.Directory != "" {
 		activeImg.Source = v1.NewDirSrc(config.Directory)
-	} else {
+	} else if _, err := config.Fs.Stat(cnst.IsoBaseTree); err == nil {
 		// If cnst.IsoBaseTree exists we are booting from iso, use that as source
-		if _, err := config.Fs.Stat(cnst.IsoBaseTree); err == nil {
-			activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
-		} else {
-			// We cannot fall back to channel upgrades always as they may be disabled, so check before setting it as source
-			if config.ChannelUpgrades {
-				activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
-			}
+		activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
+	} else {
+		// We cannot fall back to channel upgrades always as they may be disabled, so check before setting it as source
+		if config.ChannelUpgrades {
+			activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
 		}
 	}
 
-	// Error out if we could not find the source
-	if activeImg.Source.Value() == "" {
+	// Error out if we could not find the source and we haven't set the iso file, which fill the source later
+	if activeImg.Source.Value() == "" && config.Iso == "" {
 		config.Logger.Error("Could not find source for the install")
 		return &v1.SourceNotFound{}
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -61,13 +61,26 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 		MountPoint: cnst.ActiveDir,
 	}
 
-	//TODO add installation from channel
 	if config.DockerImg != "" {
 		activeImg.Source = v1.NewDockerSrc(config.DockerImg)
 	} else if config.Directory != "" {
 		activeImg.Source = v1.NewDirSrc(config.Directory)
 	} else {
-		activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
+		// If cnst.IsoBaseTree exists we are booting from iso, use that as source
+		if _, err := config.Fs.Stat(cnst.IsoBaseTree); err == nil {
+			activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
+		} else {
+			// We cannot fall back to channel upgrades always as they may be disabled, so check before setting it as source
+			if config.ChannelUpgrades {
+				activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
+			}
+		}
+	}
+
+	// Error out if we could not find the source
+	if activeImg.Source.Value() == "" {
+		config.Logger.Error("Could not find source for the install")
+		return errors.New("could not find source for the install")
 	}
 
 	// Set passive image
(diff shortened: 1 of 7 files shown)


pkg/action/install.go
```diff
@@ -67,6 +67,8 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 		activeImg.Source = v1.NewDockerSrc(config.DockerImg)
 	} else if config.Directory != "" {
 		activeImg.Source = v1.NewDirSrc(config.Directory)
+	} else if config.ChannelUpgrades {
+		activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
```
davidcassany: I am not sure how ChannelUggrades is set by default. But to me this looks a bit tricky as there is no flag for it. I suggest to give priority to the installation from `cnst.IsoBaseTree` (aka the installation media). So what about the following:

```golang
} else if exists, _ := afero.Exists(u.Config.Fs, cnst.IsoBaseTree); exists {
    activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
} else if config.ChannelUpgrades {
    activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
} else {
    return errors.New("No installation media defined")
}
```

This way running `elemental install /dev/sda` forma  device media will install whatever is in the ISO and `elemental install /dev/loop0` (not running from an ISO) will install latest `system/cos` from the channel.

The the exact same change could be applied to the `reset` command and we are done :)

davidcassany (commented):
Oh, is this that easy? If so please add it to `reset` command too :) Looks really good

What is unclear to me is from where it comes the `config.ChannelUpgrades` value. Basically this is the main reason I did not implement it at a time. Since the behavior was not clear to me and I did not want to add too many features on a single PR.

My question is, if we run `elemental install /dev/sda`  what is the value of `confing.ChannelUpgrades`, if `true` it means it will not attempt to install from the installation media. Probably we can do something smarter.

pkg/action/install.go
Itxaka: :+1: I had that same question when adding it, what would go first.

Itxaka:
> What is unclear to me is from where it comes the config.ChannelUpgrades value. Basically this is the main reason I did not implement it at a time. Since the behavior was not clear to me and I did not want to add too many features on a single PR.

So this comes from loading the default config files from cos: https://github.com/rancher-sandbox/elemental/blob/main/cmd/config/config.go#L100 which loads `/etc/cos-upgrade-image` which usually has the CHANNEL_UPGRADES and the UPGRADE_IMAGE vars, which is why on a running system the channel upgrades should always be the last one/fallback if nothing specified.

In the current case, we would need to identify if we are running from iso or `cnst.IsoBaseTree` exists before falling back to channel upgrades as otherwise from iso/recovery/running system there is always gonna be a  `/etc/cos-upgrade-image`  which those values

pkg/action/install.go
davidcassany: > The the exact same change could be applied to the `reset` command and we are done :)

@Itxaka after thinking about it twice I think that better forget about  `reset` command. Channel source doesn't make much sense there and the expected behavior is not that simple, as in `reset` command there will always be the default source dir available, hence the upgrades from channel should be an opt in for that case. Probably better leave it for now.

mudler (approved):
looks good!

pkg/action/install.go
```diff
@@ -61,13 +61,26 @@ func InstallImagesSetup(config *v1.RunConfig) error {
...
 	}
 
-	//TODO add installation from channel
 	if config.DockerImg != "" {
 		activeImg.Source = v1.NewDockerSrc(config.DockerImg)
 	} else if config.Directory != "" {
 		activeImg.Source = v1.NewDirSrc(config.Directory)
 	} else {
-		activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
+		// If cnst.IsoBaseTree exists we are booting from iso, use that as source
```
davidcassany, line 69: nit: can't it be a plain sequence of `else if`? I don't see the difference with nested ifs. just a matter of style, as you wish.

pkg/action/install.go
```diff
@@ -61,13 +61,26 @@ func InstallImagesSetup(config *v1.RunConfig) error {
...
+		} else {
+			// We cannot fall back to channel upgrades always as they may be disabled, so check before setting it as source
+			if config.ChannelUpgrades {
+				activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
+			}
+		}
+	}
+
+	// Error out if we could not find the source
+	if activeImg.Source.Value() == "" {
```
davidcassany: This will fail on `--iso`, if you use this flag from a non installation media and without channel upgrades on Souce will be empty. Probably we can make iso with the highest priority, in fact `--iso`, `--docker-image` and `--directory` should not be used together.

pkg/action/install.go
```diff
@@ -61,13 +61,26 @@ func InstallImagesSetup(config *v1.RunConfig) error {
 		MountPoint: cnst.ActiveDir,
 	}
 
-	//TODO add installation from channel
 	if config.DockerImg != "" {
```
davidcassany, line 64:
```suggestion
	if config.Iso != "" {
		activeImg.Source = v1.NewDirSrc("downloadedISO")
	} else if config.DockerImg != "" {
```

pkg/action/install.go
davidcassany: Probably we can make sources of the downloaded ISO predictable, so mount the downloaded ISO into some constant paths so we can set the Source here and forget about it. The current impletmentation of GetISO downloads the ISO into tmp folder and mounts it there and sets the Sources for active and recovery acording to the temporary path.

But I would leave this to a follow up PR.

davidcassany (changes requested):

pkg/action/install.go
Itxaka: umm, but this is a new thingie, isnt it? I dont mind adding it tot he PR but maybe it gets more complicated that expected and should be properly tracked. Do we have a card for this already?

pkg/action/install.go
Itxaka: sure, I just find nested if more readable (I know, I know, should be the other way `¯\_(ツ)_/¯`)

pkg/action/install.go
Itxaka: ah I see, this complements the comment below about the --iso

davidcassany: No, is is not new, it is parsed during the install. `elemental.GetIso()` method gets called based on `config.Iso != ""` and downloads the ISO, mounts the ISO and then sets activeImg and recoveryImg to the proper sources. Yes, I dislike this approach of setting it dynamically after downloading, but this can be fixed in another PR. The problem with the current PR is that we can hit the empty Source error if we use the `--iso` flag, which is currently supported.

pkg/action/install.go
Itxaka: is there anything that prevents use from downloading the iso during the installsetup?

pkg/action/install.go
```diff
@@ -61,13 +61,24 @@ func InstallImagesSetup(config *v1.RunConfig) error {
...
 		activeImg.Source = v1.NewDirSrc(cnst.IsoBaseTree)
+	} else {
+		// We cannot fall back to channel upgrades always as they may be disabled, so check before setting it as source
+		if config.ChannelUpgrades {
+			activeImg.Source = v1.NewChannelSrc(cnst.ChannelSource)
+		}
+	}
+
+	// Error out if we could not find the source and we haven't set the iso file, which fill the source later
+	if activeImg.Source.Value() == "" && config.Iso == "" {
```
Itxaka, line 79: cc @davidcassany

davidcassany: Yes that's also a good option

pkg/action/install.go
davidcassany: I guess we could consider it but this is a bigger change and makes the iso download happen before the `before-install` hook (though I don't think we should care about it).

Also I like this setup methods to be super light, this is good for testing and from an API PoV.

davidcassany (approved):

davidcassany:
@Itxaka the grubtest test fails because now `InstallImagesSetup` method does not have default Source for ActiveImage, it can error out depending on the host and configs context, I guess for this test there is no source defined.

Itxaka:
Yep, good eye @davidcassany. I have fixed the other tests but failed to catch this one :)

PR #134, opened 2022-03-10
Bump grsync wrapper

### LINKED ISSUE #124: Some race conditions in our tests

davidcassany:
This commit bumps grsync package to fix some race issues that were time to time detected in our unit tests.

The version bump includes zloylos/grsync#19 however it should include zloylos/grsync#20
Related to #124

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/vishvananda/netns v0.0.0-20211101163701-50045581ed74 // indirect
 	github.com/xanzy/ssh-agent v0.3.1 // indirect
 	github.com/zcalusic/sysinfo v0.0.0-20210905121133-6fa2f969a900 // indirect
-	github.com/zloylos/grsync v1.5.1
+	github.com/zloylos/grsync v1.6.1
 	golang.org/x/crypto v0.0.0-20220126234351-aa10faf2a1f8 // indirect
 	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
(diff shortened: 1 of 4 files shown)


Itxaka:
wow, that was fast, very nice!

Itxaka (approved):

davidcassany:
@Itxaka @mudler believe it or not, but now the situation is way worse... it always fail with the new grsync 😅 And it also fails on my local env. I am currently debugging it.

I think the grsync issue was worse than what I initially thought, probably the problem is that there is a goroutine waiting on `scanner.Scan()` for stdout and stderr, but it could happen that one of the two pipes (stdout and/or stderr) does not produce a single character, hence no EOF and it just blocks there... trying to validate it 🤔  I hope I did not break grsync with my PR 😓

davidcassany:
Ok there were several issues here, first SyncData had a small issue while using the fs, the trailing `/` needs to be added after the convertion to the real fs. On the way I discovered the super ugly #135 😞  and that my fix of the grsync is not as good as I wanted, it can lead to infinite loops 😞 

Damn! Too many things at once to something that was supposed to be a low hanging fruit! 😭

Itxaka:
should probably close this now that we have our own fork @davidcassany :)

davidcassany:
Yes lets close this, no longer needed!

PR #136, opened 2022-03-10
Fix directory permissions

### LINKED ISSUE #135: Wrong use of os.ModeDir
We have been using `os.ModeDir` as constant to `0755`, however this is not the case... 😞 It is a constant for the directory type of file. Hence directories created with utils.MkdirAll(fs, "/somedir", os.ModeDir) have wrong permissions. We mostly have not seen it because we run in root mode all the time, hence no permissions complaints. @mudler faced it while leaving from aftero, so in some tests started to require root permissions.... this was the underlaying reason, the tests did not require root. Our code is just horribly wrong in that regard.

This is what is left on my host after running `elemental install` to a loop device:

```bash 
sudo ls /run/cos/ -lah 
total 0
d---------  8 root root  160 Mar 10 13:55 .
drwxr-xr-x 44 root root 1.2K Mar 10 13:55 ..
d---------  2 root root   40 Mar 10 13:55 active
d---------  2 root root   40 Mar 10 13:55 efi
d---------  2 root root   40 Mar 10 13:55 oem
d---------  2 root root   40 Mar 10 13:55 persistent
d---------  2 root root   40 Mar 10 13:55 recovery
d---------  2 root root   40 Mar 10 13:55 state
```

davidcassany:
Fixes #135

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -90,8 +90,8 @@ const (
 	GPT                    = "gpt"
 
 	// Default directory and file fileModes
-	DirPerm  = os.ModeDir | 0755
-	FilePerm = 0644
+	DirPerm  = os.ModeDir | os.ModePerm
+	FilePerm = 0666
 )
 
 func GetCloudInitPaths() []string {
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -17,6 +17,7 @@ limitations under the License.
 package constants
 
 import (
+	"os"
 	"runtime"
 )
 
@@ -87,6 +88,10 @@ const (
 	PassiveImgName         = "passive"
 	RecoveryImgName        = "recovery"
 	GPT                    = "gpt"
+
+	// Default directory and file fileModes
+	DirPerm  = os.ModeDir | 0755
+	FilePerm = 0644
 )
 
 func GetCloudInitPaths() []string {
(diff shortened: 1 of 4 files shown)


davidcassany:
Apparently os.ModeDir is not the equivalent to `0755` according to [docs](https://pkg.go.dev/io/fs#ModeDir).

Itxaka:
Can't we use something like `os.ModeDir | os.ModePerm` ? I'm pretty sure that there is a combination which is clear. Or we can set this as a constant in our side as to not repeat it that much and maintain the same dirs all across

davidcassany:
> Can't we use something like `os.ModeDir | os.ModePerm` ? I'm pretty sure that there is a combination which is clear. Or we can set this as a constant in our side as to not repeat it that much and maintain the same dirs all across

I tried but I could not find a  simple match, see [here](https://go.dev/play/p/Ce-Ha6A2ixa). But yes, probably a constant makes sense.

pkg/constants/constants.go
```diff
@@ -87,6 +88,10 @@ const (
 	PassiveImgName         = "passive"
 	RecoveryImgName        = "recovery"
 	GPT                    = "gpt"
+
+	// Default directory and file fileModes
+	DirPerm  = os.ModeDir | 0755
```
Itxaka: :+1:

Itxaka (approved):

Itxaka:
`os.ModeDir | os.ModePerm` should work for us as long as the system has the proper default umask (0022 I think?) which would make the dir end up with 755 as expected.  Anyway, this is good enough for now, no need to go further IMO.

PR #140, opened 2022-03-14
Keep partition device path after creating a new partition

davidcassany:
This commit stores the device path of a partition after creating it.
This is helpful to reduce the amount of calls to tools such as `blkid`
or `lsblk` during the installation.


Itxaka (commented):
good one

Itxaka (approved):

davidcassany:
This is a little change to reduce `blkid` and `lsblk` calls. In fact, this allows to run `elemental install` without calling `blkid` not even once.

Also, I realized we can definitely completely remove `blkid` usage from our code, having only `lsblk` should be enough. That would make it easier to also remove `lsblk`  in the future if want to use some golang native lib, since we can manage to have only a single `lsblk` call in a single utils method to pull devices data.

This makes calling elemental from docker container more robust.

PR #141, opened 2022-03-15
Use auto for FS in upgrade

Itxaka:
Dont require the actual FS from the partition, instead let mount be
smart about it. This should make us depend less on blkid/lsblk data for
the upgrade


Itxaka:
Tested this manually on a vm to upgrade from 0.8.5 to 0.8.6 and it mounted everything correctly.

davidcassany (approved):
Looks good :+1:

PR #142, opened 2022-03-15
Add logger to runner and log commands on debug

Itxaka:
Adds a logger to the runner and logs commands on debug calls.
Takes care of trying to use a nil Logger and doesnt force a logger to
exists

Side effect of checking some commands locally, I implemented this real quick to see the commands being executed :)

### DIFF
--- a/pkg/types/v1/runner.go
+++ b/pkg/types/v1/runner.go
@@ -18,15 +18,20 @@ package v1
 
 import (
 	"os/exec"
+	"strings"
 )
 
 type Runner interface {
 	InitCmd(string, ...string) *exec.Cmd
 	Run(string, ...string) ([]byte, error)
 	RunCmd(cmd *exec.Cmd) ([]byte, error)
+	GetLogger() Logger
+	SetLogger(logger Logger)
 }
 
-type RealRunner struct{}
+type RealRunner struct {
+	Logger Logger
+}
 
 func (r RealRunner) InitCmd(command string, args ...string) *exec.Cmd {
 	return exec.Command(command, args...)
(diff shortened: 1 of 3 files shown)


mudler (approved):
Looks good

Itxaka:
@davidcassany I know you wanted to do this, not sure if you like the implementation, opened to any ideas.

pkg/types/v1/runner.go
```diff
@@ -38,5 +43,16 @@ func (r RealRunner) RunCmd(cmd *exec.Cmd) ([]byte, error) {
 
 func (r RealRunner) Run(command string, args ...string) ([]byte, error) {
 	cmd := r.InitCmd(command, args...)
+	if r.Logger != nil {
+		r.Logger.Debugf("Running cmd: '%s %s'", command, strings.Join(args, " "))
+	}
```
davidcassany, line 48: I'd also log the output as a debug message.

davidcassany (commented):
Looks good to me :+1: I am just missing logging the output of the command.

My idea was even simpler, I was not considering exposing the logger to the interface. So just initiating a RealRunner with something like `NewRealRunner(logger v1.Logger)`. I like having logger control on the interface, so we can eventually modify it at runtime if there is some command horribly verbose or we want to hide simple repetitive commands :+1:

davidcassany (approved):

pkg/types/v1/runner.go
Itxaka: Good point, I was wondering if the output should be logged for everything or not, like it could get a bit obnoxious if we log the out of everything? Maybe its better like now that we return the out and err and if we consider that its important we log the out when erroring out?

Itxaka:
Im gonna go ahead with the merge of this @davidcassany and we can move on the output of the command later on, like I said I think it migth turn out to be a bit obnoxious if done all the time.

PR #143, opened 2022-03-15
Dockerfile changes

Itxaka:
- Build dockerfile on PR only when the file is changed. Does not affect master/tag building
 - Add a target to the Makefile to build docker image locally
 - Use version and commit while building docker image elemental
 - Use version and commit for image tags on docker image
 - Fix version on makefile getting the long version (tag+commit) instead of the short version (tag)
 - Master merged docker images are now pushed with a tag+short_commit instead of `sha-long_commit` as its easier to differentiate
 - Built image is now locally pushed in the worker so it can be reused in the job

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -48,4 +48,4 @@ jobs:
           # mount loop device and get the device
           LOOP=`sudo losetup -fP --show loopbackfile.img`
           # run the install tot he loop device as a small test
-          docker run --privileged ${{ steps.meta.outputs.tags }} install --debug $LOOP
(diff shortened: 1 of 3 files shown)


mudler:
mmm but the underlying code might be changed?

Itxaka:
> mmm but the underlying code might be changed?

Yes, but that should not affect the dockerfile building, especially if we dont test it :D -That why the tests are there, the docker stuff should only be triggered **on PRs** when there is a change to the Dockerfile. Its still building and pushing on merges to main and on new tags.

mudler (dismissed):

mudler (approved):

PR #147, opened 2022-03-16
Remove blkid dependency and limit lsblk usage

davidcassany:
This commit removes all blkid calls and wraps lsblk in a more
normalized and generic way. lsblk is only called in two methods:

* GetDevicePartitions: returns a v1.PartitionList with all the
partitions or loop devices for the given device path.

* GetAllPartitions: returns a v1.PartitionList with all the partitions
and loop devices of the host.

### DIFF
--- a/pkg/partitioner/disk.go
+++ b/pkg/partitioner/disk.go
@@ -19,6 +19,7 @@ package partitioner
 import (
 	"errors"
 	"fmt"
+	"os"
 	"regexp"
 	"strings"
 	"time"
@@ -94,9 +95,20 @@ func (dev Disk) GetLabel() string {
 	return dev.label
 }
 
-func (dev Disk) Exists() bool {
-	exists, _ := utils.Exists(dev.fs, dev.device)
-	return exists
+func (dev *Disk) Exists() bool {
+	fi, err := dev.fs.Stat(dev.device)
+	if err != nil {
+		return false
+	}
+	// resolve symlink if any
+	if fi.Mode()&os.ModeSymlink != 0 {
+		d, err := dev.fs.Readlink(dev.device)
+		if err != nil {
+			return false
+		}
+		dev.device = d
+	}
+	return true
 }
 
 func (dev *Disk) Reload() error {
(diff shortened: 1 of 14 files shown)


davidcassany:
With this PR installing from a docker container just worked, I did something like:
```
docker run --privileged -v /dev:/dev -ti elemental:v0.0.13-770286d install --docker-image quay.io/costoolkit/releases-green:cos-system-0.8.3-1 /dev/loop0
```
and it all went good.

Note that, by far, most of the changes of this PR are around tests (in fact it took me way longer than expected to adapt them 😓 ). Here I just adapted the current tests, I have not coded yet specific unit tests for the new `lsblk` utilities, that's why I marked it as a draft, however this is ready to review, as I do not expect to add anything else than few unit tests.

In between I also found a tiny bug in the layout yip plugin, we were not passing the new vfs.FS to the partitioner, hence there were two FS instances around, I bet this is why we had to make yip tests to run as root. Now they are back to a non root execution.

pkg/partitioner/disk.go
```diff
@@ -267,25 +279,24 @@ func (dev Disk) WipeFsOnPartition(device string) error {
 }
 
 func (dev Disk) FindPartitionDevice(partNum int) (string, error) {
-	re, _ := regexp.Compile(fmt.Sprintf("(?m)^(/.*%d) part$", partNum))
+	re := regexp.MustCompile(`.*\d+$`)
```
Itxaka, line 282: hell yeah, always happy when I see a regex simplified :D

pkg/utils/lsblk.go
```diff
@@ -0,0 +1,109 @@
...
+func GetAllPartitions(runner v1.Runner) (v1.PartitionList, error) {
+	out, err := runner.Run("lsblk", "-p", "-b", "-n", "-J", "--output", "LABEL,SIZE,FSTYPE,MOUNTPOINT,PATH,PKNAME,TYPE")
+	if err != nil {
+		return nil, err
+	}
+
+	return unmarshalLsblk(out)
+}
+
+// GetDevicePartitions gets an array of partition devices mapped into v1.Partition
```
Itxaka: same comment on both methods but they do a different thing!

davidcassany: ouch! copy & paste is evil...

pkg/utils/lsblk.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, err
+	}
+
+	return parts, nil
+}
+
+// GetAllPartitions gets an array of partition devices mapped into v1.Partition
+// objects.
+func GetAllPartitions(runner v1.Runner) (v1.PartitionList, error) {
+	out, err := runner.Run("lsblk", "-p", "-b", "-n", "-J", "--output", "LABEL,SIZE,FSTYPE,MOUNTPOINT,PATH,PKNAME,TYPE")
```
Itxaka, line 92: Im just thinking out loud here, but maybe having an object that gets the partitions on init and then having the possibility of calling the object.Reload() to reload partitions may be better? I havent checked how often we need the refreshed partition list (probably on reset/install) but it could be a way of calling lsblk only when needed, otherwise we have to keep calling GetDevicePartitions, which means a call to lsblk each time.

Of course if we want to substitute lsblk ASAP, this is a moot point :D

pkg/utils/common.go
```diff
@@ -55,40 +49,43 @@ func BootedFrom(runner v1.Runner, label string) bool {
 // attempts value sets the number of attempts to find the device, it
 // waits a second between attempts.
 func GetDeviceByLabel(runner v1.Runner, label string, attempts int) (string, error) {
```
Itxaka, line 51: does it make sense to have 2 methods for this anymore? Like if we return a structured Partition, we can easily use the part.Path directly instead of having 2 methods :thinking: 

Not for this PR thougth.

pkg/utils/lsblk.go
```diff
@@ -0,0 +1,109 @@
...
+		if part.Type == "part" || part.Type == "loop" {
+			partitions = append(partitions, part.Partition())
+		}
+	}
+	*p = partitions
+	return nil
+}
+
+func unmarshalLsblk(lsblkOut []byte) ([]*v1.Partition, error) {
+	var objmap map[string]*json.RawMessage
```
Itxaka, line 70: interesting :eyes:

Itxaka (approved):

pkg/utils/lsblk.go
davidcassany: Interesting idea indeed. Probably we could have some sort of `Host` struct and keep partitions and whatever else we need to know from the host in it. Probably I would consider that in the scope of a RunConfig refactor, as by doing that I think we can encounter that this new structure does not fully substitute our current usage for some of the fields. For instance `v1.Partition` in the `install` action is used to express what we want to create not what we load from the host.

I like the idea, but I think it might require a proper reorganization of RunConfig.

pkg/utils/common.go
davidcassany: Yes good point, probably we can simply get rid of it.

Itxaka (approved):

pkg/utils/utils_test.go
```diff
@@ -199,6 +199,109 @@ var _ = Describe("Utils", Label("utils"), func() {
 			Expect(runner.CmdsMatch(append(cmds, cmds...))).To(BeNil())
 		})
 	})
+	Describe("GetAllPartitions", Label("lsblk"), func() {
```
Itxaka: NIT: I would add partitions to the labels here as this deals with partitions, so we can easily check that we havent broke anything when we change anything on the partition type

PR #148, opened 2022-03-16
Move config options into the config package

Itxaka:
It makes no sense to have the config options under types, lets move it
to the config package instead

Also adds some missing tests


davidcassany (approved):
Nice :+1:

PR #149, opened 2022-03-16
Test the docker image install over loop device

Itxaka:
Some other improvements:

 - mv arg/env to after go mod download as to not ivalidate the steps
before
 - use gha cache for building
 - add elemental install test for docker image
 - add docker.yaml file to docker job
 - user qemu-img to create the empty disk image
 - only build the image before tests, do not push
 - push the image after tests if not on PR

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -46,13 +46,14 @@ jobs:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_TOKEN }}
-      - name: Build and push
+      - name: Build image
         uses: docker/build-push-action@v2
         with:
-          # push and load can't be used at the same time
-          push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
-          load: ${{ github.event_name == 'pull_request' }} # loads it locally, so it can be used from docker client
+          load: true # loads it locally, so it can be used from docker client
+          # cache into GitHub actions cache, nice
+          cache-from: type=gha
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yaml
```diff
@@ -55,9 +60,9 @@ jobs:
...
         run: |
           # create a 30Gb file
-          # qemu-img create -f raw disk.img 30G
+          qemu-img create -f raw disk.img 30G
           # mount loop device and get the device
-          #LOOP=`sudo losetup -fP --show disk.img`
+          LOOP=`sudo losetup -fP --show disk.img`
           # unfortunately loop devices are not properly refreshed under a container :/
-          #docker run --privileged ${{ steps.meta.outputs.tags }} install --debug -d quay.io/costoolkit/releases-green:cos-system-0.8.6 $LOOP
+          docker run -v /dev/:/dev/ --privileged ${{ steps.meta.outputs.tags }} install --debug -d quay.io/costoolkit/releases-green:cos-system-0.8.6 $LOOP
```
davidcassany, line 78: Probably we can use a `--force-efi` flag here and just run two tests, one with EFI and one without. See #150

Itxaka: I will wait for #150 to merge then rebase this and add the efi test :D

Itxaka: done!

Itxaka:
We should probably do something else here? Maybe? I dunno, seems like this covers what we need, checking that the required tools are in the container. Maybe we should set a layout with an XFS fs to check those as well...

davidcassany (approved):

PR #150, opened 2022-03-16
Add grub2-efi image into the docker image

davidcassany:


davidcassany:
I needed that to run an EFI installation

Itxaka (approved):
good catch!

Itxaka:
No need to wait for the tests to run. Maybe we should skip running the tests if only the dockerfile is changed or there are no code changes....


EDIT: https://github.com/tj-actions/changed-files looks really nice....

PR #155, opened 2022-03-20
Drop lsblk in favour of ghw library

Itxaka:
- ghw uses udev database and files under /sys and /proc to gather all
info
 - This uses that exclusively so no more lsblk
 - Probably means that the partition/disk identification utils can now
without root privs
 - Adds a new mock to mock a full disk(s) with partition(s) for tests
only
 - Makes tests use the new mock to mock the lsblk calls
 - Fixes a couple of leftovers on dir permissions

Unofrtunately this needs my own fork of ghw as upstream is missing a
patch to get the labels from the udev database, but it should be merged
upstream soon.

It also means that when running elemental from docker we need to link
the /run/udev/data volume, but this can be done on RO as we only read,
which would be a bit nicer. Commadn will look uglier thougth, but as
long as we explain why it needs that volume, it should all be good.


davidcassany (approved):
Looks good to me. Just a couple of nits


If we are changing from `GetAllPartitions(runner v1.Runner)` to `GetAllPartitionsV2()`, probably without keeping the former method, probably we can simply avoid the `V2` into the method name. Looks like there could be an alternate version of the method. Also, the `pkg/utils/lsbk.go` file I think it should be renamed to something else if it is not wrapping `lsblk`.

Beyond this two naming details it all looks good to me. :+1: I think this is a good move. And as discussed already, using the installer within a container will always be tricky or require specific uncommon volumes and privilages, it should not be the preferred way to use elemental.

Itxaka:
yup, the naming and old functions are there 'cause...I was testing them side by side :D

In practice this was gonna be a work on top of the existing stuff to be substituted, but I saw it was only called in a few places, tried changing those and fixing the tests and suddenly those functions are not called anymore :o did not expect it to be this straightforward.

Itxaka:
oh yeah what about the ghw Partition type VS our Partition type? We could drop our partition as they are identical but relaying on a external type....not good if they change it, but I would like to avoid doing the absurd transformation from one struct to the other which IMHO is wasted cycles.

Itxaka:
I changed it to return our type, we can work out if we want to drop the transformation afterwards

davidcassany:
> oh yeah what about the ghw Partition type VS our Partition type? We could drop our partition as they are identical but relaying on a external type....not good if they change it, but I would like to avoid doing the absurd transformation from one struct to the other which IMHO is wasted cycles.

I'd leave it for another PR and probably after a while. The good thing of this the current approach is that ghw is only imported within the utils wrapper and in tests. v1.Partitions is used in several places, so if we encounter some issue with ghw and consider going back to something else the changes are mostly affecting tests, the elemental logic is pretty isolated from it.

kkaempf:
Not sure about the long-term implications of this:
- `lsblk` is part of the core operating system
- there's someone to blame if lsblk has problems :wink:
- `ghw` increases our external dependencies
- problems in `ghw` are _on us_ to fix

mudler:
> Not sure about the long-term implications of this:
> 
>     * `lsblk` is part of the core operating system
> 
>     * there's someone to blame if lsblk has problems wink
> 
>     * `ghw` increases our external dependencies
> 
>     * problems in `ghw` are _on us_ to fix

Let's not confuse `blkid` `lsblk` and `ghw`. `ghw` is a library for hardware detection, and for partitions and such it uses also blkid behind the scene, among other tools (@Itxaka also improved it to use udev!). 

This is about avoiding us to re-invent the wheel and write from scratch a detecting library, with all the issues that it might bring.. it makes sense instead to contribute to an open source one that we just adopt.

Itxaka:
>   lsblk is part of the core operating system

Not on cos, only 'cause we need it for elemental :P

>   there's someone to blame if lsblk has problems wink

well...same? we blame ghw :P

>   ghw increases our external dependencies

Not really, its compiled into a final binary babyyyyy. ghw is really ligth and we only use a small part of it, not sure why the problem with this and not the other 300 dep that we have :rofl: 

>   problems in ghw are on us to fix

Good. I like that, cause I already fixed issues on it and expanded. Good luck getting me to fix something on lsblk :rofl:

Itxaka:
upstream PR missing https://github.com/jaypipes/ghw/pull/310

Itxaka:
upstream patch merged, this should be ready tomorrow!

davidcassany (approved):
LGTM :+1:

pkg/action/action_test.go
```diff
@@ -86,40 +90,50 @@ var _ = Describe("Actions", func() {
...
+						Type:  "ext4",
+					},
+					{
+						Name:  "device3",
+						Label: "COS_PERSISTENT",
+						Type:  "ext4",
+					},
+				},
+			}
+			ghwTest = v1mock.GhwMock{}
```
mudler, line 131: :heart:

mudler (approved):
Looks good!

PR #158, opened 2022-03-24
Relocate GPT headers for expanded disks

### LINKED ISSUE #1201: Test raw disk deploy fails
Seems the test raw disk deploy started to fail recently:

```
• Failure [165.664 seconds]
cOS Recovery deploy tests
/Users/runner/work/cOS-toolkit/cOS-toolkit/tests/recovery-raw-disk/recovery_raw_disk_test.go:9
  after running recovery from the raw_disk image
  /Users/runner/work/cOS-toolkit/cOS-toolkit/tests/recovery-raw-disk/recovery_raw_disk_test.go:23
    uses cos-deploy to install [It]
    /Users/runner/work/cOS-toolkit/cOS-toolkit/tests/recovery-raw-disk/recovery_raw_disk_test.go:24

    Unexpected error:
        <*errors.withStack | 0xc0000b0d80>: {
            error: <*errors.withMessage | 0xc0000d92c0>{
                cause: <*ssh.ExitError | 0xc0000f1a80>{
                    Waitmsg: {status: 1, signal: "", msg: "", lang: ""},
                },
                msg: "\x1b[36mINFO\x1b[0m[2022-03-24T01:05:53Z] Starting elemental version v0.0.13           \nError: no device found\n\x1b[31mERRO\x1b[0m[2022-03-24T01:05:54Z] State partition 'COS_STATE' not found        \n",
            },
            stack: [0x12cefdd, 0x12d129a, 0x12d1280, 0x126d27a, 0x126cc45, 0x126c33b, 0x126f4a9, 0x126ee87, 0x1284e85, 0x1284ba5, 0x12843e5, 0x1286692, 0x1291ca5, 0x129196b, 0x12d0f65, 0x10dc202, 0x1069ca1],
        }
        INFO[2022-03-24T01:05:53Z] Starting elemental version v0.0.13           
        Error: no device found
        ERRO[2022-03-24T01:05:54Z] State partition 'COS_STATE' not found        
        : Process exited with status 1
    occurred

    /Users/runner/work/cOS-toolkit/cOS-toolkit/tests/recovery-raw-disk/recovery_raw_disk_test.go:28
------------------------------

```

**Additional context**
- https://github.com/rancher-sandbox/cOS-toolkit/runs/5668948084?check_suite_focus=true
- https://github.com/rancher-sandbox/cOS-toolkit/runs/5668591417?check_suite_focus=true
- https://github.com/rancher-sandbox/cOS-toolkit/runs/5668740862?check_suite_focus=true
- https://github.com/rancher-sandbox/cOS-toolkit/runs/5684036928?check_suite_focus=true

davidcassany:
This commit ensure the partitioner module run 'sgdisk -e /dev/disk' to
fix GPT headers in case the disk was expanded or some smaller image was
dumped into the disk.

Fixes rancher-sandbox/cOS-toolkit#1201

### DIFF
--- a/pkg/partitioner/parted.go
+++ b/pkg/partitioner/parted.go
@@ -155,17 +155,6 @@ func (pc PartedCall) Print() (string, error) {
 	return string(out), err
 }
 
-// HasUnallocatedSpace parses a parted print command output and checks for a
-// warning within the output complaining about wrong GPT end header. If
-// the warning is found it assumes the GPT end header is not at the end of
-// the disk, hence it needs to be adapted to get valid sizes from parted
-// print. Parted has no proper command to verify GPT headers neither a
-// proper command to fix them whenever is possible in a non interactive mode.
-func (pc PartedCall) HasUnallocatedSpace(printOut string) bool {
-	unallocated, _ := regexp.MatchString("Not all of the space available", printOut)
-	return unallocated
-}
-
 // Parses the output of a PartedCall.Print call
 func (pc PartedCall) parseHeaderFields(printOut string, field int) (string, error) {
 	re := regexp.MustCompile(`^(.*):(\d+)s:(.*):(\d+):(\d+):(.*):(.*):(.*);$`)
--- a/pkg/partitioner/parted.go
+++ b/pkg/partitioner/parted.go
@@ -155,6 +155,17 @@ func (pc PartedCall) Print() (string, error) {
 	return string(out), err
 }
 
+// HasUnallocatedSpace parses a parted print command output and checks for a
+// warning within the output complaining about wrong GPT end header. If
+// the warning is found it assumes the GPT end header is not at the end of
+// the disk, hence it needs to be adapted to get valid sizes from parted
+// print. Parted has no proper command to verify GPT headers neither a
+// proper command to fix them whenever is possible in a non interactive mode.
+func (pc PartedCall) HasUnallocatedSpace(printOut string) bool {
+	unallocated, _ := regexp.MatchString("Not all of the space available", printOut)
+	return unallocated
+}
+
 // Parses the output of a PartedCall.Print call
 func (pc PartedCall) parseHeaderFields(printOut string, field int) (string, error) {
 	re := regexp.MustCompile(`^(.*):(\d+)s:(.*):(\d+):(\d+):(.*):(.*):(.*);$`)
--- a/pkg/partitioner/disk.go
+++ b/pkg/partitioner/disk.go
@@ -118,6 +118,19 @@ func (dev *Disk) Reload() error {
 		return err
 	}
 
+	if pc.HasUnallocatedSpace(prnt) {
+		// Parted has not a proper way to doing it in non interactive mode,
+		// because of that we use sgdisk for that...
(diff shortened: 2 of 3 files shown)


davidcassany:
I am not happy of having to use `sgdisk` just for that but it looks like this is a very long standing limitation of [parted](https://bugzilla.redhat.com/show_bug.cgi?id=217979). Found some discussions from 2020 about patching `parted` to better support this use case and it seams a `--fix` flag was [included upstream around early 2021](https://git.savannah.gnu.org/cgit/parted.git/commit/?id=7c9a96209af2b84e1eec98ad3527f82ddd966cdf) but this is not yet released in any release.

Just validated that the patch is not included in TW and Ubuntu, only Fedora includes a patched parted including this change. So, hopefully, this is something we can revisit at some point in the future.

Itxaka (approved):

pkg/partitioner/parted.go
```diff
@@ -155,6 +155,17 @@ func (pc PartedCall) Print() (string, error) {
 	return string(out), err
 }
 
+// HasUnallocatedSpace parses a parted print command output and checks for a
+// warning within the output complaining about wrong GPT end header. If
+// the warning is found it assumes the GPT end header is not at the end of
+// the disk, hence it needs to be adapted to get valid sizes from parted
+// print. Parted has no proper command to verify GPT headers neither a
+// proper command to fix them whenever is possible in a non interactive mode.
+func (pc PartedCall) HasUnallocatedSpace(printOut string) bool {
```
mudler: small nit here, better keep this private. It is parsing a string for a known pattern so also the regexp could be compiled instead of being a func wrapper

mudler (approved):

pkg/partitioner/parted.go
mudler: besides, it is safe to assume that we want to always call sgdisk here in any case this pattern matches?

davidcassany: > besides, it is safe to assume that we want to always call sgdisk here in any case this pattern matches?

I think so and it is aligned with the `--fix` flag that parted should include within the next release (I am not sure which is the criteria to cut releases, since it is more than a year already since last one). So if we'd use parted from master upstream we would not need this check at all and we would simply call `parted print` command with `--fix` flag, which essentially _fixes_ gpt headers if there is some inconsistency. So the `Disk.Reload` method would essentially run `parted -s --fix -m <device> unit s print` and it would load the disk partition table at the time it fixes GTP header inconsistencies, so the values of the disk size print outputs matches the actual disk geometry.

In parted there is no such command to actually _fix_ headers, it can only do it on the fly in interactive mode or fully automated with the `--fix` flag in non interactive mode (up coming releases). `sgdisk` has `verify` and `expand` commands, so you have the tools to add logic to handle this situation in different scenarios if you want.

`sgdisk` is a well designed tool to script on top of it, `parted` is not that good in that regard, scripted commands are convoluted (probably to match GPT and MSDOS support in a single UI) and not as powerfull as in interactive mode. The only reason to use parted is that is supports MSDOS partition tables and it allows us to support GPT and MSDOS with a single logic or wrapper.

PR #159, opened 2022-04-07
Move out generic configuration from RunConfig

davidcassany:
This commit is a first step to move generic configuration from
RunConfig to a new Config struct. In this commit, to narrow the scope
of changes the new Config struct is embedded into RunConfig, so
from a RunConfig PoV the change is transparent.

This allows to reuse Config struct in BuildConfig, so all the
interfaces initalizations can be reused there.

Related to rancher-sandbox/cOS-toolkit#1217
Partial #33 

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -25,101 +25,107 @@ import (
 	"k8s.io/mount-utils"
 )
 
-type RunConfigOptions func(a *v1.RunConfig) error
+type GenericOptions func(a *v1.Config) error
 
-func WithFs(fs v1.FS) func(r *v1.RunConfig) error {
-	return func(r *v1.RunConfig) error {
+func WithFs(fs v1.FS) func(r *v1.Config) error {
+	return func(r *v1.Config) error {
 		r.Fs = fs
 		return nil
 	}
 }
 
-func WithLogger(logger v1.Logger) func(r *v1.RunConfig) error {
-	return func(r *v1.RunConfig) error {
+func WithLogger(logger v1.Logger) func(r *v1.Config) error {
+	return func(r *v1.Config) error {
 		r.Logger = logger
 		return nil
 	}
 }
 
-func WithSyscall(syscall v1.SyscallInterface) func(r *v1.RunConfig) error {
-	return func(r *v1.RunConfig) error {
+func WithSyscall(syscall v1.SyscallInterface) func(r *v1.Config) error {
+	return func(r *v1.Config) error {
 		r.Syscall = syscall
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


davidcassany:
While working on rancher-sandbox/cOS-Toolkit#1217 I started to need the runner, luet, logger, fs, etc. At first I started moving around only the fs and logger, but then luet came and I could not stand it anymore. I did not want to reproduce all this logic for the `BuildConfig` struct.

I do not really like this approach, as I would completely strip out from RunConfig the generic configuration, but including such a change in rancher-sandbox/cOS-Toolkit#1217 scope is way too much. This is way the change is minimal and transparent and it already allows to reuse the initialization logic. I consider it some sort of a first step for #33 and it already prevents some logic duplication without having to revisit all tests and many functions signatures.

pkg/config/config.go
```diff
@@ -165,3 +171,10 @@ func NewRunConfig(opts ...RunConfigOptions) *v1.RunConfig {
 	}
 	return r
 }
+
+func NewBuildConfig(opts ...GenericOptions) *v1.BuildConfig {
+	b := &v1.BuildConfig{
+		Config: *NewConfig(opts...),
```
Itxaka, line 177: doesnt this mean that to access the logger now we need to use the buildconfig.Config.Logger ? Like it adds an extra level to access those clients? I dont get how this works then :/

Like if we had something creating an instance of runconfig and accessing its luet, which used to be at top level (i.e. config.Luet) how does that work now? Shouldnt it fail as now it needs config.Config.Luet???? Or does adding the Config struct in the RunConfig makes some magic and adds those values into the RunConfig somehow??

Itxaka (approved):
Looks good!

mjura (approved):
Implementation looks good

pkg/config/config.go
davidcassany: > Config struct in the RunConfig makes some magic and adds those values into the RunConfig somehow??

Yes this is the case. So you can embed the struct as you would for interfaces. I have to admit the literal initialization syntax gives the impression of a nested struct, I was also surprised by this syntax. I guess it is hard do express the assignment of all fields at once and they came out with this.

Note there is no filed Config in RunConfig, it might be not obvious to see in the github diff. We have something like:

```golang
type Config struct {
   ...
}

type RunConfig struct {
   ...
   Config
}
```

Itxaka: woah, nice. I did not expect that, especially with the `Config: *NewConfig(opts...),` I expected it to be a field in the struct called config. Very nice!

PR #160, opened 2022-04-12
build-iso command first round implementation

davidcassany:
This commit includes a first round of the 'build-iso' subcommand
implementation. It is functional enough to create a hybrid ISO bootable
for both, BIOS and EFI firmwares.

Related with rancher-sandbox/cOS-Toolkit#1217

### DIFF
--- a/pkg/action/build-iso.go
+++ b/pkg/action/build-iso.go
@@ -242,38 +242,8 @@ func burnISO(c *v1.BuildConfig, root string) error {
 		"-volid", c.ISO.Label, "-joliet", "on", "-padding", "0",
 		"-outdev", outputFile, "-map", root, "/", "-chmod", "0755", "--",
 	}
+	args = append(args, constants.GetDefaultXorrisoBooloaderArgs(root, c.ISO.BootFile, c.ISO.BootCatalog, c.ISO.HybridMBR)...)
 
-	// TODO: make this detection more robust
-	// Assume ISOLINUX bootloader is used if boot file is includes 'isolinux'
-	// in its name, otherwise assume an eltorito based grub2 setup
-	if strings.Contains(c.ISO.BootFile, "isolinux") {
-		args = append(args, []string{
-			"-boot_image", "isolinux", fmt.Sprintf("bin_path=%s", c.ISO.BootFile),
-			"-boot_image", "isolinux", fmt.Sprintf("system_area=%s/%s", root, c.ISO.HybridMBR),
-			"-boot_image", "isolinux", "partition_table=on",
-		}...)
-	} else {
-		args = append(args, []string{
-			"-boot_image", "grub", fmt.Sprintf("bin_path=%s", c.ISO.BootFile),
-			"-boot_image", "grub", fmt.Sprintf("grub2_mbr=%s/%s", root, c.ISO.HybridMBR),
-			"-boot_image", "grub", "grub2_boot_info=on",
-		}...)
-	}
-
-	args = append(args, []string{
-		"-boot_image", "any", "partition_offset=16",
-		"-boot_image", "any", fmt.Sprintf("cat_path=%s", c.ISO.BootCatalog),
-		"-boot_image", "any", "cat_hidden=on",
-		"-boot_image", "any", "boot_info_table=on",
-		"-boot_image", "any", "platform_id=0x00",
-		"-boot_image", "any", "emul_type=no_emulation",
-		"-boot_image", "any", "load_size=2048",
-		"-append_partition", "2", "0xef", fmt.Sprintf("%s/boot/uefi.img", root),
-		"-boot_image", "any", "next",
-		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
-		"-boot_image", "any", "platform_id=0xef",
-		"-boot_image", "any", "emul_type=no_emulation",
-	}...)
 	out, err := c.Runner.Run(cmd, args...)
 	c.Logger.Debugf("Xorriso: %s", string(out))
 	if err != nil {
--- a/pkg/action/build-iso.go
+++ b/pkg/action/build-iso.go
@@ -1,0 +1,331 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package action
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/rancher-sandbox/elemental/pkg/constants"
+	"github.com/rancher-sandbox/elemental/pkg/partitioner"
+	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
+	"github.com/rancher-sandbox/elemental/pkg/utils"
+)
+
+// BuildISORun will install the system from a given configuration
+func BuildISORun(cfg *v1.BuildConfig) (err error) {
+	cleanup := utils.NewCleanStack()
+	defer func() { err = cleanup.Cleanup(err) }()
+
+	rootDir, err := utils.TempDir(cfg.Fs, "", "elemental-iso-rootfs")
+	if err != nil {
+		return err
+	}
+	cleanup.Push(func() error { return cfg.Fs.RemoveAll(rootDir) })
+
+	uefiDir, err := utils.TempDir(cfg.Fs, "", "elemental-iso-uefi")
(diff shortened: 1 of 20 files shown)


davidcassany:
marked as draft because most unit tests are still missing. In addition there are few details to agree about, the current implementation supports the following manifest:

```yaml
iso:                                                                            
  rootfs:                                                                       
    - system/cos                                                                
  uefi:                                                                         
    - live/grub2-efi-image                                                      
  isoimage:                                                                     
    - live/grub2                                                                
    - live/grub2-efi-image                                                      
  label: "COS_LIVE"                                                             
                                                                                
name: "cOS-0"                                                                   
date: true
```

Not the same but similar to the current manifest. All these values can be defined from the command line. To find kernel and initrd there is an heuristic logic to determine the path, hence I believe it doesn't make much sense to expose those too. Regarding other values such as the old

```yaml
boot_file: "boot/x86_64/loader/eltorito.img"                                    
boot_catalog: "boot/x86_64/boot.catalog"                                        
isohybrid_mbr: "boot/x86_64/loader/boot_hybrid.img" 
```

are currently set the these defaults and exposed to the schema. However I doubt this should be user facing as those values are strictly coupled with `live/grub2` package. How to set them, I see three options:

1. Set those values in `live/grub2` package somehow and force elemental to read them, in such a case we should document this procedure.
2. Assume some expectations and discover them according to that. Assume certain filenames and locations and define some heuristics to discover them. I like this idea, however our current syslinux/systemd-boot and syslinux/grub2 packages are radically different, so it might require some refactor on syslinux or grub2 package to better align them.
3. Keep current status exposing it to the manifest. I dislike this idea since this values are really tied to the internals of the bootloader packages and having some defaults there is strictly tied to some specific implementation. IMHO elemental should not be coupled to luet packages, ideally it should be the other way round to facilitate customization from a derivative PoV.

davidcassany:
In addition this PR is raising the need of #33, currently elemental API is coupled with RunConfig, API should only be coupled to basic and fundamental configurations (Luet, Runner, Logger, Cloud-Init, Filesystem, etc.)

pkg/action/build-iso.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	// End of CreateFileSystemImage copy
+
+	files, err := c.Fs.ReadDir(root)
+	if err != nil {
+		return err
+	}
+
+	for _, f := range files {
+		_, err = c.Runner.Run("mcopy", "-s", "-i", img, filepath.Join(root, f.Name()), "::")
```
Itxaka, line 213: Is there any way we can avoid forking to mcopy here? Maybe mouting the img, copying the files and then umounting it?

davidcassany: Indeed this is a good point. The mcopy use was here to allow non privileged builds in a container. I also prefer mounting and unmounting, code wise it way cleaner. However then we will loose the capability of building ISOs using the elemental container in unprivileged mode.

I am wondering to what extend should we care about it. To build disks we already assumed this will be case. Using privileges opens the door to better code reuse.

pkg/action/build-iso.go
```diff
@@ -0,0 +1,331 @@
...
+	} else {
+		args = append(args, []string{
+			"-boot_image", "grub", fmt.Sprintf("bin_path=%s", c.ISO.BootFile),
+			"-boot_image", "grub", fmt.Sprintf("grub2_mbr=%s/%s", root, c.ISO.HybridMBR),
+			"-boot_image", "grub", "grub2_boot_info=on",
+		}...)
+	}
+
+	args = append(args, []string{
+		"-boot_image", "any", "partition_offset=16",
```
Itxaka: I guess this should all end up in constants somehow, althougth I wonder about the naming of all of this LOL, thats gonna be the worst

pkg/utils/common.go
```diff
@@ -292,3 +296,44 @@ func GetSource(config *v1.RunConfig, source string, destination string) error {
...
+	return true
+}
+
+// NewSrcGuessingType returns new v1.ImageSource instance guessing its type
+// applying somne heuristic techniques (by order of preference):
+//   1. Assume it is Dir/File if value is found as a path in host
+//	 2. Assume it is a container registry reference if it matches [<domain>/]<repositry>:<tag>
+//      (only domain is optional)
+//	 3. Fallback to a channel source
+func NewSrcGuessingType(c v1.Config, value string) v1.ImageSource {
```
Itxaka, line 328: niiiice

pkg/action/build-iso.go
davidcassany: I am not that sure there is a gain having those in constants, I doubt we will call xorriso anywhere else. Anyway I guess we can have a xorriso default arguments method in constants that constructs all this. :+1:

pkg/action/build-iso.go
mudler: seems actually this is actually being used, in context of rolling up an ISO directly within kubernetes and making it available in a webserver. For the moment being I 'd keep it as is. 

cc: @agracey is the "building ISO" in kubernetes something we want to stick our interest in?

mudler:
> marked as draft because most unit tests are still missing. In addition there are few details to agree about, the current implementation supports the following manifest:
> 
> ```yaml
> iso:                                                                            
>   rootfs:                                                                       
>     - system/cos                                                                
>   uefi:                                                                         
>     - live/grub2-efi-image                                                      
>   isoimage:                                                                     
>     - live/grub2                                                                
>     - live/grub2-efi-image                                                      
>   label: "COS_LIVE"                                                             
>                                                                                 
> name: "cOS-0"                                                                   
> date: true
> ```
> 
> Not the same but similar to the current manifest. All these values can be defined from the command line. To find kernel and initrd there is an heuristic logic to determine the path, hence I believe it doesn't make much sense to expose those too. Regarding other values such as the old
> 
> ```yaml
> boot_file: "boot/x86_64/loader/eltorito.img"                                    
> boot_catalog: "boot/x86_64/boot.catalog"                                        
> isohybrid_mbr: "boot/x86_64/loader/boot_hybrid.img" 
> ```
> 
> are currently set the these defaults and exposed to the schema. However I doubt this should be user facing as those values are strictly coupled with `live/grub2` package. How to set them, I see three options:
> 
>     1. Set those values in `live/grub2` package somehow and force elemental to read them, in such a case we should document this procedure.
> 
>     2. Assume some expectations and discover them according to that. Assume certain filenames and locations and define some heuristics to discover them. I like this idea, however our current syslinux/systemd-boot and syslinux/grub2 packages are radically different, so it might require some refactor on syslinux or grub2 package to better align them.
> 
>     3. Keep current status exposing it to the manifest. I dislike this idea since this values are really tied to the internals of the bootloader packages and having some defaults there is strictly tied to some specific implementation. IMHO elemental should not be coupled to luet packages, ideally it should be the other way round to facilitate customization from a derivative PoV.

Yes I do agree those settings do not really belong here. I was thinking to even have hardcoded default on the CLI for the time being, or even not be able to specify at all and just keep support to grub2.
 On point 2 I see it difficult to put in practice, with the additional complexity of false positive hitting the road. I see appealing going over option 1 of shipping a config in a package that is read by our tool. 

At the end of the day we stick with a simple documented rule, that will likely change not much often, and as we will merge the build commands in elemental we can than have a more obvious link in the codebase

davidcassany:
I consider main logic to be ready, however some portions of the `pkg/action/build-iso.go` should be adapted, restructured or replaced based on #33 outcomes.

Code and logic wise I think this sufficient for a first round. It enables us to start working on integration items from rancher-sandbox/cOS-Toolkit#1217

Note syslinux support is not there yet, we are not transparent to grub2 or syslinux cOS-Toolkit packages.

pkg/constants/constants.go
```diff
@@ -88,6 +89,18 @@ const (
...
+	BuildImgName           = "elemental"
+
+	//TODO these paths are abitrary, coupled to package live/grub2 and assuming xz
+	// I'd suggest using `/boot/kernel` and `/boot/initrd`
+	IsoKernelPath = "/boot/kernel.xz"
+	IsoInitrdPath = "/boot/rootfs.xz"
+
+	// TODO would be nice to discover these ISO loader values instead of hardcoding them
+	// These values are coupled with package live/grub2
+	IsoHybridMBR   = "/boot/x86_64/loader/boot_hybrid.img"
```
mudler, line 104: this actually make me wonder, do we inject default repositories somewhere within changes? wasn't able to find it out if it's already there, or is it expected still to be passed by in the config file?

mudler (approved):
Looks good here!

pkg/action/build-iso.go
agracey: I do think it's very useful to allow the builds to be done in k8s. It allows for both more automation and a reduction of tooling needed on the admin's laptop.

Also, we could eventually build and publish a Tekton task that allows more people to access the project!

agracey:
~~Would this be to replace the luet command that we use currently?~~ nvm, I read the attached ticket...

pkg/constants/constants.go
davidcassany: This is still to be completed, I am aware repositories are not customizable yet. It requires an extension of the Luet interface. I already had a look at it and is not a big deal. We need to agree on how to express it in the yaml form. I'd like to make the feature available also for the other commands (upgrade, install, etc.), since they share logic regarding channels.

PR #161, opened 2022-04-13
Extend LuetInterface to support custom repositories

davidcassany:
This commit extends the luet unpack from channel feature by adding
an optional list of custom repositories.

In addition this repository list is mapped to build config manifest and
into the build-iso flags.

related to rancher-sandbox/cOS-Toolkit#1217


davidcassany:
@Itxaka @mudler I moved luet implementation to its own package to prevent circular dependencies, as long as I added more logic to manage repositories I needed it, as it happened with cloud-init interface a while ago.

pkg/types/v1/config.go
```diff
@@ -163,11 +163,21 @@ type LiveISO struct {
 	HybridMBR   string   `yaml:"hybrid_mbr,omitempty" mapstructure:"hybrid_mbr,omitempty"`
 }
 
+// Repository represents the basic configuration for a package repository
+type Repository struct {
```
Itxaka, line 167: should we just use the luet repository struct instead of our own? I mean, this has to be in sync with upstream anyway as its passed to luet so...

Itxaka: oh, I see there is some initialization done on our side. Cant we offload that onto luet as well??

Itxaka: ok, I see this is for convenience, we just pass an url instead of a full repo config. Ok, good enough for me and better for end users :)

davidcassany:
This allows a build manifest such as:

 ```yaml
iso:
  rootfs:
    - system/cos
  uefi:
    - live/grub2-efi-image
  isoimage:
    - live/grub2
    - live/grub2-efi-image
  label: "COS_LIVE"

# For repositories only the URI field is mandatory:
#  * repo type is guessed form the URI if not provided.
#  * name is set to a md5sum from URI if not provided (deterministic POSIX compliant name)
repositories:
  - uri: quay.io/elemental/releases
  - name: facny repo
    uri: http://some.repo.org
    type: http
    priority: 1

name: "cOS-0"
date: true
 ```

pkg/types/v1/config.go
davidcassany: I would not, we would lose part of the benefits from the LuetInterface if so. Currently luet is only imported within the Luet implementation and not exposed in other parts of the code. I think it makes sense keeping it this way, makes or code cleaner and constraints luet dependency to the luet interface implementation. This way we could even think of a package manager interface and an OCI manipulator interface, where Luet could implement both.

pkg/luet/luet.go
```diff
@@ -0,0 +1,285 @@
...
+		} else if utils.ValidContainerReference(repo.URI) {
+			repoType = "docker"
+		} else {
+			return luetTypes.LuetRepository{}, fmt.Errorf("Invalid Luet repository URI: %s", repo.URI)
+		}
+	}
+
+	return luetTypes.LuetRepository{
+		Name:     name,
+		Priority: repo.Priority,
```
davidcassany, line 180: @mudler if priority is unset in config I understand this will be set to `0`, does it have any ugly side effect on Luet's side? Should we have other default instead of  `0`?

mudler: I think we should be good, this would set it as max prio. we have to just keep in mind that in the CI this is the other way around as we set up the final repo lower to run our tests on: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/make/Makefile.iso#L93 while the local path the artifacts to test is with defaults (`0`)

davidcassany: I see, all this will need to be adapted though. I am fine having highest priority by default, since the provided repositories are the only ones that will be used, hence it doesn't really matter the priority if it is not provided, it they will all be on the same level.

mudler (dismissed):
This looks good! also nice refactor and separate the struct from the interface

cmd/command_test.go
```diff
@@ -50,3 +55,22 @@ func executeCommandC(cmd *cobra.Command, args ...string) (c *cobra.Command, outp
 
 	return c, string(out), nil
 }
+
+// TODO instead of this  hack probably is simpler to find a way to use a new cmd instances per test
+func resetFlags(cmd *cobra.Command) {
```
davidcassany: This approach feels hackish and I do not really like it. To me it feels like a better approach would be instance a new command in each test so there is no need to cleanup anything. Since we use global variables this requires a bit of a refactor, but at this point it is unclear to me if this challenging or not. I am missing some details about Cobra command initialization.

Cobra has no `AllFlagsToDefaults` method or strategy, so the cleanup has to be implemented by us if we reuse the command instance across tests. So that I opted for this ugly reflection code I found googling.

cmd/build-iso.go
```diff
@@ -103,15 +122,17 @@ var buildISO = &cobra.Command{
...
 	buildISO.Flags().String("overlay-rootfs", "", "Path of the overlayed rootfs data")
 	buildISO.Flags().String("overlay-uefi", "", "Path of the overlayed uefi data")
 	buildISO.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
-	// TBC expose rootfs as flags?
-	//buildISO.Flags().StringArray("rootfs", []string{}, "A list of sources for the rootfs image. Can be repeated to add more than one source.")
-	buildISO.Flags().StringArray("isoimage", []string{}, "A list of sources for the ISO image. Can be repeated to add more than one source.")
-	buildISO.Flags().StringArray("uefi", []string{}, "A list of sources for the UEFI image. Can be repeated to add more than one source.")
+
+	// The dot notation is used to express nested maps in viper settings which is needed to unmarshal nested structs.
+	buildISO.Flags().String("iso.label", "", "Label of the ISO volume")
```
davidcassany: @mudler in Luet are you mapping flags to nested struct keys? I am wondering if there is any better or nicer way to handle this. In this particular case having to prepend the `iso.` into flags to bind into nested `ISO` struct is not that bad, it actually looks nice IMHO, however i assume this will problematic in the scope of #33. You have any experience on that regard?

cmd/command_test.go
davidcassany: Just managed to recreate the cobra.Command instance for each tests using  `BeforeEach`, gonna provide a commit for that. Unfortunately the PR is getting big, but I really want to fix it and get rid of this ugly `resetFlags`.

davidcassany:
I know there is huge diff, sorry 🙏🏽 I tried to make as simpler as possible though.

You can review it per commits:
*  first two are the actual feature
*  next comes the unit tests (there I had to refactor a little the commands structure so we are not recycling the same cobra.Command instance in every test)
* last commit is just applying the refactor to all commands (the diff looks horrible, but if you see actual code you will see I just wrapped subcommand initialization into a function).

cmd/build-iso.go
mudler: more or less it's the same in `luet`: https://github.com/mudler/luet/blob/6674abb256156d1df896ca3581c3e2ae4dbd45c8/cmd/util/config.go#L287, there are just few initializators in place to handle this a bit more nicely

mudler (approved):
Looks good!

cmd/install.go
```diff
@@ -27,68 +27,76 @@ import (
...
-	RunE: func(cmd *cobra.Command, args []string) error {
-		path, err := exec.LookPath("mount")
-		if err != nil {
-			return err
-		}
-		mounter := mount.New(path)
+// NewInstallCmd returns a new instance of the install subcommand and appends it to
+// the root command. requireRoot is to initiate it with or without the CheckRoot
+// pre-run check. This method is mostly used for testing purposes.
+func NewInstallCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
```
mudler, line 33: Just a small nit feel free to ignore, maybe better call them `Register` or something among the lines so are more indicative on the fact that they add themselves to the root command, the fact that they return the instance is more or less not significant if not for testing

Itxaka (approved):
very nice, especially the command stuff. I wanted to change them to that a long time ago but had no clear justification, so this is awesome. This also means we can expand the cmd testing further! very nice!

cmd/install.go
Itxaka: I think this is a commonly used patter across cobra projects, to call this NewXCmd

PR #162, opened 2022-04-18
Add build-disk command

Itxaka:
This allows a config as follows:

```
raw_disk:
  x86_64:
    repositories:
      - uri: quay.io/costoolkit/releases-green
    # which packages to install and the target to install them at
    packages:
      - name: system/grub2-efi-image
        target: efi
      - name: system/grub2-config
        target: root
      - name: system/grub2-artifacts
        target: root/grub2
      - name: recovery/cos-img
        target: root/cOS
  aarch64:
    repositories:
      - uri: quay.io/costoolkit/releases-green-arm64
    packages:
      - name: system/grub2-efi-image
        target: efi
      - name: system/grub2-config
        target: root
      - name: system/grub2-artifacts
        target: root/grub2
      - name: recovery/cos-img
        target: root/cOS
  odroid_c2:
    repositories:
      - uri: quay.io/costoolkit/releases-green-arm64
    packages:
      - name: system/grub2-efi-image
        target: efi
      - name: system/grub2-config
        target: root
      - name: system/grub2-artifacts
        target: root/grub2
```

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -36,7 +36,7 @@ import (
 	"k8s.io/mount-utils"
 )
 
-func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig, error) {
+func ReadConfigBuild(configDir string, mounter mount.Interface, manifestRequired bool) (*v1.BuildConfig, error) {
 	logger := v1.NewLogger()
 	cfg := config.NewBuildConfig(
 		config.WithLogger(logger),
@@ -46,21 +46,21 @@ func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig
 
 	configLogger(cfg.Logger, cfg.Fs)
 
-	if exists, _ := utils.Exists(cfg.Fs, configDir); exists {
-		viper.AddConfigPath(configDir)
-		viper.SetConfigType("yaml")
-		viper.SetConfigName("manifest")
-		// If a config file is found, read it in.
-		err := viper.MergeInConfig()
-		if err != nil {
-			cfg.Logger.Warnf("error merging config files: %s", err)
+	viper.AddConfigPath(configDir)
+	viper.SetConfigType("yaml")
+	viper.SetConfigName("manifest.yaml")
+	// If a config file is found, read it in.
+	err := viper.MergeInConfig()
+	if err != nil {
+		if manifestRequired {
+			return nil, err
 		}
 	}
 
 	viperReadEnv()
 
 	// unmarshal all the vars into the config object
-	err := viper.Unmarshal(cfg)
+	err = viper.Unmarshal(cfg)
 	if err != nil {
 		cfg.Logger.Warnf("error unmarshalling config: %s", err)
 	}
--- a/pkg/action/build_test.go
+++ b/pkg/action/build_test.go
@@ -17,7 +17,10 @@
 package action_test
 
 import (
+	"bytes"
 	"errors"
+	"github.com/sirupsen/logrus"
+	"os"
 	"path/filepath"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -43,12 +46,15 @@ var _ = Describe("Runtime Actions", func() {
 	var cloudInit *v1mock.FakeCloudInitRunner
 	var luet *v1mock.FakeLuet
 	var cleanup func()
+	var memLog *bytes.Buffer
 	BeforeEach(func() {
 		runner = v1mock.NewFakeRunner()
 		syscall = &v1mock.FakeSyscall{}
 		mounter = v1mock.NewErrorMounter()
 		client = &v1mock.FakeHTTPClient{}
-		logger = v1.NewNullLogger()
+		memLog = &bytes.Buffer{}
+		logger = v1.NewBufferLogger(memLog)
+		logger.SetLevel(logrus.DebugLevel)
 		cloudInit = &v1mock.FakeCloudInitRunner{}
 		luet = &v1mock.FakeLuet{}
 		fs, cleanup, _ = vfst.NewTestFS(map[string]interface{}{})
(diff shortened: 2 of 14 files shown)


Itxaka:
missing tests

davidcassany (commented):
Looks good. Event I'd try to abstract `sgdisk` and try to map into the raw disk the partition parameters and reuse the v1.Partition struct, this way we can make all easily configurable (partitions sizes, labels, etc.). It can all be done in follow up PRs.

A bit off-topic regarding the implementation I realized this is only for building vanilla images (images without state and persistent partitions). Note this does not build a complete disk image, it is done this way to allow the operation in non privileged containers. It also relies in several linux utilities (`sgdisk`, `mkfs.*`, `mtools`...). This opens few questions:

* Shouldn't we provide a default mechanism for auto deployment? I mean embed the required cloud-init for a full deployment that can be easily parametrized/customized at build time.
* Is building disks in on privileged containers a really relevant requirement? Note that if not we can simply use `elemental install` over a loop device and provider way more fine grained control over the installation and reusing most of the code. We could even consider making `elemental install` more flexible so it can install as it was a vanilla (so, no state and persistent partitions) and reuse the already existing API.
* If these vanilla images in non privileged containers are relevant shouldn't we provide also a real OEM image build? aka running an installation over a loop device.
* Which are the use cases from a derivative developer of the `build-disk` command? To me it feels the common use case would be running a full installation, rather than a vanilla image.

@itxaka @mudler @mjura @ldevulder To me it is kind a blurry the need of this and I am not sure we are just doubling logic for the sake of being capable to build disks in non privileged containers, which is cool, but really needed?

Itxaka:
> Looks good. Event I'd try to abstract sgdisk and try to map into the raw disk the partition parameters and reuse the v1.Partition struct, this way we can make all easily configurable (partitions sizes, labels, etc.). It can all be done in follow up PRs.

Yes, I was having issues trying to build a working raw image and it was failing on sgdisk, so I decided to keep it 1to1 with the script until I found the reason. That is fixed now (didn't seek the file after truncation lmao) so I could have a look.

> Shouldn't we provide a default mechanism for auto deployment? I mean embed the required cloud-init for a full deployment that can be easily parametrized/customized at build time.

Mmh, this is so we can build off images by providing the cloud init at the start already, not sure what else would we need here? This is the base for deploying/building your aws, gce, azure images. Adding that wouldnt break the point of this bootstrap/vanilla image?

> Is building disks in on privileged containers a really relevant requirement?

I dont think so. I dont even remember where that feature came from or what asked for. Maybe we should revisit that.

> If these vanilla images in non privileged containers are relevant shouldn't we provide also a real OEM image build? aka running an installation over a loop device.

What are the advantages of those VS this image? This is a smaller, recovery only image that allows you to use a cloud-init to configure everything in a cloud operator, I see that as an advantage.

> Which are the use cases from a derivative developer of the build-disk command? To me it feels the common use case would be running a full installation, rather than a vanilla image.

I guess maintaining their custom recovery cloud image to build their final cloud images from? I mean, at the moment there doesnt seems to be any advantage for a derivative that I can see, but maybe in the future there is? Like adding extra drivers for a cloud operator maybe?


> @Itxaka @mudler @mjura @ldevulder To me it is kind a blurry the need of this and I am not sure we are just doubling logic for the sake of being capable to build disks in non privileged containers, which is cool, but really needed?

Again, I think the purpose of this is to be a base cloud image. Not all cloud operators allow you to boot an image from an iso and you want to have a small base image with nothing extra to be able to generate your custom cloud images. At least that is how we are using them, and I think os2 is using this base vanilla image to build (or at least its pre-configured to do so) AWS images for os2. Different purposes I think.

davidcassany:
> What are the advantages of those VS this image? This is a smaller, recovery only image that allows you to use a cloud-init to configure everything in a cloud operator, I see that as an advantage.

Sure I fully agree on that, I also see the value on these images. My point is that if we assume we do not need to circumvent privileges everything turns to be easier more flexible and we can reuse elemental API. So instead of this convoluted way of creating partitions we could simply use the partitioner and loop mount them to copy data in there. In such a case we would be just dealing with a custom installation (basically a half way installation). So 90% of the code is already there. I coded this weird disk creation script the way it is to make it possible to run in non privileged containers, so no mount, no parted and the bloody squashfs image :sweat_smile: 💢 This was one of the random important features that were asked at a time and completely forgotten afterwards.

Probably we could describe the vanilla image as something like:

```yaml
disk:
  # if system is set to `none`   omit/ignore persistent and state partitions
  system: none
  recovery:
    source: docker://my.registry.org/cos:latest
```

and full image build as:

```yaml
disk:
  system:
     source: docker://my.registry.org/my/derivative:latest
  recovery:
    source: docker://my.registry.org/cos:latest
```

This could be mostly aligned with the `install` struct from #33

Itxaka:
@davidcassany umm that makes a lot of sense to me. Let meblock this until we agree on #33

Itxaka:
umm, I still see some issues with that, as we need to get extra packages other than recovery for the raw image... are we gonna hardcode those? Derivatives are screwed in that case....

mudler:
> Looks good. Event I'd try to abstract `sgdisk` and try to map into the raw disk the partition parameters and reuse the v1.Partition struct, this way we can make all easily configurable (partitions sizes, labels, etc.). It can all be done in follow up PRs.
> 
> A bit off-topic regarding the implementation I realized this is only for building vanilla images (images without state and persistent partitions). Note this does not build a complete disk image, it is done this way to allow the operation in non privileged containers. It also relies in several linux utilities (`sgdisk`, `mkfs.*`, `mtools`...). This opens few questions:
> 
>     * Shouldn't we provide a default mechanism for auto deployment? I mean embed the required cloud-init for a full deployment that can be easily parametrized/customized at build time.
> 
>     * Is building disks in on privileged containers a really relevant requirement? Note that if not we can simply use `elemental install` over a loop device and provider way more fine grained control over the installation and reusing most of the code. We could even consider making `elemental install` more flexible so it can install as it was a vanilla (so, no state and persistent partitions) and reuse the already existing API.
> 
>     * If these vanilla images in non privileged containers are relevant shouldn't we provide also a real OEM image build? aka running an installation over a loop device.
> 
>     * Which are the use cases from a derivative developer of the `build-disk` command? To me it feels the common use case would be running a full installation, rather than a vanilla image.
> 
> 
> @Itxaka @mudler @mjura @ldevulder To me it is kind a blurry the need of this and I am not sure we are just doubling logic for the sake of being capable to build disks in non privileged containers, which is cool, but really needed?

At the moment we don't strictly need this to be able to build on containers, however, consider that building ISO currently is, and I would not exclude that for disks in the future. I do agree that recovery images are a very specific case indeed, and instead a full blown system install is mostly prefered, altough I see how this could be potentially being used in kubernetes pipelines as well. 

I'm not much fancying copying 1:1, because I see this should be using the API. Maybe we could do the API aware if the build context is rather privileged or not and adapt accordingly, so the code leveraging it would stay the same, but that's room for a follow-up

Itxaka:
umm, do we want to move this forwards as a v1 of the build-disk stuff then rework it with the new config and so on?

currently this is just a 1to1 copy of the script, takes advantage of nothing at all from elemental building block other than config.

CC @davidcassany @mudler

davidcassany:
> umm, do we want to move this forwards as a v1 of the build-disk stuff then rework it with the new config and so on?

Sounds good to me preparing it as first round.

Itxaka:
umm, elemental is really tied with the runconfig, so to use its API for all the heavy lifting (creating the images, mounting, blablabla) it would require a big rework of elemental to use either a config interface or something like that :/

davidcassany:
> umm, elemental is really tied with the runconfig, so to use its API for all the heavy lifting (creating the images, mounting, blablabla) it would require a big rework of elemental to use either a config interface or something like that :/

Exactly that's why #33 turned to be relevant now. Build-iso has quite some duplicated code because of that. Elemental should not be coupled with `v1.RunConfig`, but with `v1.Config` and pass the relevant data of each method as a parameter (partition/image structs or image source...).

Itxaka:
@mudler @davidcassany @mjura ready for review as a first pass. Raw image generated boots as expected with a valid manifest.yaml (see under cmd/config/config/manifest.yaml for a valid test one)

cmd/config/config.go
```diff
@@ -46,21 +46,21 @@ func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig
...
-		viper.AddConfigPath(configDir)
-		viper.SetConfigType("yaml")
-		viper.SetConfigName("manifest")
-		// If a config file is found, read it in.
-		err := viper.MergeInConfig()
-		if err != nil {
-			cfg.Logger.Warnf("error merging config files: %s", err)
+	viper.AddConfigPath(configDir)
+	viper.SetConfigType("yaml")
+	viper.SetConfigName("manifest.yaml")
```
mudler, line 51: :+1: I'm wondering how it was working before? did it tried with all extensions?

pkg/action/build_test.go
```diff
@@ -185,4 +191,78 @@ var _ = Describe("Runtime Actions", func() {
 			Expect(err).Should(HaveOccurred())
 		})
 	})
+	FDescribe("Build disk", Label("disk", "build"), func() {
```
mudler: Leftover?

pkg/luet/luet.go
```diff
@@ -189,7 +189,7 @@ func (l Luet) initLuetRepository(repo v1.Repository) (luetTypes.LuetRepository,
 func (l Luet) UnpackFromChannel(target string, pkg string, repositories ...v1.Repository) error {
 	var toInstall luetTypes.Packages
 	toInstall = append(toInstall, l.parsePackage(pkg))
-	l.log.Debugf("Luet config: %+v", l.context.Config)
+	//l.log.Debugf("Luet config: %+v", l.context.Config)
```
mudler:
```suggestion

```

pkg/action/build_test.go
Itxaka: woops

mudler (approved):
Looks good here! this is looking good as a first pass, I see placeholders for the other disk types which is nice :+1:

cmd/config/config.go
Itxaka: Not really, it worked because it would infer that the file would have a yaml extension due to the configtype. But that resulted in an ugly error if not found in which the file was mentioned without extension, i.e `Error: Config File "manifest" Not Found in "[/home/itxaka/projects/elemental-cli]"`

As that could result in confusion if we mention manifest.yaml but the cli errors out with a manifest file, I just changed it so it can throw a proper error message. We could even drop the configtype altogether, but its clearer like this to me.

PR #165, opened 2022-04-20
Exclude only root-located directories

### LINKED ISSUE #164: Install excludes directories without '/' prefix
https://github.com/rancher-sandbox/elemental/blob/6fd0d6d61a6a33306cd674c33a62e68df9c29921/pkg/elemental/elemental.go#L342
Implies that the directories considered are located in root (e.g. `/run`, `/mnt`, etc ) however it looks like it's not the case and dirs are passed by as-is down the path, hence it excludes also directories like `/var/run`.

mudler:
This commit fixes https://github.com/rancher-sandbox/elemental/issues/164.

It was wrongly implied to exclude root level directories when copying from a directory,
resulting in other folder, such as /var/run to be excluded as well.

It also includes specific unit tests to avoid falling into this in the
future if we ever change the underlying lib.

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -339,7 +339,7 @@ func (c *Elemental) CopyImage(img *v1.Image) error { // nolint:gocyclo
 			return err
 		}
 	} else if img.Source.IsDir() {
-		excludes := []string{"mnt", "proc", "sys", "dev", "tmp", "host", "run"}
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Nice test expansion, rsync is pretty nit-picky with all these slash details, having extended tests on few corner cases is really helpful.

PR #166, opened 2022-04-21
Spkie about new runconfig loading values from config.yaml

davidcassany:
Note this is not any implementation of anything, this is just playground to experiment with different configurations. It basically attempts to load a config.yaml file similar to what is discussed in #33 to check how feasible it would be and how it would be translated into go structures.

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -23,8 +23,10 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
+	"github.com/mitchellh/mapstructure"
 	"github.com/rancher-sandbox/elemental/internal/version"
 	"github.com/rancher-sandbox/elemental/pkg/config"
 	"github.com/rancher-sandbox/elemental/pkg/luet"
@@ -35,6 +37,30 @@ import (
 	"github.com/spf13/viper"
 	"k8s.io/mount-utils"
 )
+
+type Unmarshaler interface {
+	CustomUnmarshal(interface{}) error
+}
+
+func UnmarshalerHook() mapstructure.DecodeHookFunc {
+	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
+		// If the destination implements the unmarshaling interface
+		u, ok := to.Interface().(Unmarshaler)
+		if !ok {
+			return from.Interface(), nil
+		}
+		// If it is nil and a pointer, create and assign the target value first
+		if to.IsNil() && to.Type().Kind() == reflect.Ptr {
+			to.Set(reflect.New(to.Type().Elem()))
+			u = to.Interface().(Unmarshaler)
+		}
+		// Call the custom unmarshaling method
+		if err := u.CustomUnmarshal(from.Interface()); err != nil {
+			return to.Interface(), err
+		}
+		return to.Interface(), nil
+	}
+}
 
 func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig, error) {
 	logger := v1.NewLogger()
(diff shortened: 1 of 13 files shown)


cmd/config/config.go
```diff
@@ -36,6 +38,30 @@ import (
 	"k8s.io/mount-utils"
 )
 
+type Unmarshaler interface {
+	CustomUnmarshal(interface{}) error
+}
+
+func UnmarshalerHook() mapstructure.DecodeHookFunc {
```
davidcassany, line 45: This magic is executed every time viper tries to unmarshal someting defined in `maspstructure` to some object. If the destination object implements the `Unmarshaler` interface it executes the custom unmarshal logic of this object. If the destination object does not implement the interface it just by passes and continues to the following decode hook up to the default decoder.

config.yaml
```diff
@@ -0,0 +1,51 @@
+install:
+  # target is the only mandatory
+  target: /dev/sda
+
+  # basic disk configs for formatting
+  firmware: efi
+  partition-table: gpt
+
+  # partitions setup, any missing value is taken from defaults
+  partitions:
```
davidcassany, line 10: partitions is a list so it is ordered and the `id` parameter is to identify the partition (oem, recovery, state or persistent). I am not sure if there is a better way to express that. 

Elemental expects a sorted list (iterable in a for loop) of partitions during the installation. So this maps to it directly straight forward, but probably we could apply some custom marshal logic if we find a better way to express it in yaml form.

config.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+  # use this iso as installation media (overwrites 'image' and 'recoveryImage' sources with iso contents)
+  iso: https://my.domain.org/some/powerful.iso
+
+  # main OS image
+  system:
+    label: COS_ACTIVE
+    size: 1024
+    fs: ext2
+    source: docker://quay.io/costoolkit/green-releases:v1.2.3
```
davidcassany, line 31: this could be converted into `--system-image` flag in elemental client all in one instead of `--docker-image` and `--directory`. So we could have a single flag to express system source for all types of sources.

config.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+  recovery:
```
davidcassany, line 33: similarly we could have `--recovery-image` flag, so we can independently set a different recovery image if we want to, if unset just use the system one.

cmd/installnew.go
```diff
@@ -0,0 +1,93 @@
...
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			path, err := exec.LookPath("mount")
+			if err != nil {
+				return err
+			}
+			mounter := mount.New(path)
+
+			cfg, err := config.ReadConfigRunNew(viper.GetString("config-dir"), mounter)
```
davidcassany, line 52: from that config then the method `action.InstallSetup(cfg)` would fill all the missing values with defaults and set the struct with all the data required for the installation so `action.InstallRun(cfg)` should be minimally impacted by all these changes.

davidcassany:
@Itxaka @mudler ideas and opinions are more than welcome.

Itxaka:
looks good to me. Although Im worried about going down the hole of the custom masrshall/unmarshall I guess its a necessary evil :/

Now I wonder, what do we exactly want to expose as overridable by flag? I would say the minimum we can, as you can use config files to override values and even the config-dir flag to point to an extra dir with the configs, because starting to allow overriding all this stuff via flag is a terrible idea. Its ugly, its terrible for normal usage and also overrides the nice thing of having a config written and even tracked via git

davidcassany:
> Although Im worried about going down the hole of the custom masrshall/unmarshall I guess its a necessary evil :/

Yes agree, the alternative is to perfectly use a loadable as is yaml but then you have process it into new structures and/or code a bunch helpers and alias.

For instance, the custom unmarshaler is only used for the v1.ImageSource in this example, we could also define image source as:

```golang
type ImageSource string

func (i ImageSource) Type() error {
    // parse string as url and/or guess type somehow
}

func (i ImageSource) GetValue() string {
    // return value without type, basically skip scheme
}

...
```

this could also work, but then we should be really careful to not use this type as string and relay on its methods the get usable data out of it... This is how I would avoid custom unmarshal, but it doesn't convinces me either, feels unsafe and convoluted too. It is easy to miss use.

pkg/types/v1/common.go
```diff
@@ -16,51 +16,104 @@ limitations under the License.
...
 package v1
 
+import (
+	"fmt"
+	"net/url"
+	"path/filepath"
+)
+
+const (
+	docker     = "docker"
```
Itxaka, line 26: btw, I would start calling this container instead of docker as it sounds like a docker image, but its more an OCI image no?

pkg/types/v1/common.go
```diff
@@ -16,51 +16,104 @@ limitations under the License.
...
+	file       = "file"
+	dir        = "dir"
+	channelPkg = "channelpkg"
```
Itxaka, line 29: would also leave this as channel with no pkg added

pkg/types/v1/common.go
davidcassany: that's a good point, probably we should support both and treat them equally? I prefer using OCI, but I wonder if this is what what people expects out there.

mudler (approved):
Looks good here! I like the approach so it is transparent to the code already there, we just hook over on how we read the config by satisfying a viper interface, so this looks really clean as well

pkg/types/v1/common.go
mudler: good point, maybe we can add support for alias with a `map[string][]string` here , we could also cover cases like `channel-pkg`, `channel`, etc. like below

davidcassany:
superseded by #189

PR #168, opened 2022-04-21
Fix build-iso kernel discovery and root permissions

davidcassany:
This commit fixes the kernel discovery to avoid 'vmlimux' images,
which are usually and intermediate uncompressed kernel image.

In addition it makes sure that root folder of the iso and squashfs are
having the appropriate 0755 permissions mask.


davidcassany:
At some I had add `vmlinux` file as valid kernel candidate, I can't really remember why and when and it broke iso builder since it was picking the wrong kernel image.

In addition while debugging this I noted that the root of the iso had 0700 permissions (should have 0755), this is also fixing it.

mudler (approved):

PR #169, opened 2022-04-22
Fix --date default

davidcassany:


cmd/build-iso.go
```diff
@@ -127,7 +127,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	}
 	root.AddCommand(c)
 	c.Flags().StringP("name", "n", "", "Basename of the generated ISO file")
-	c.Flags().Bool("date", true, "Adds a date suffix into the generated ISO file")
+	c.Flags().Bool("date", false, "Adds a date suffix into the generated ISO file")
```
davidcassany, line 130: apparently this default value is the default when the `--date` value is unsed. I thought it was the other way round...

mudler (approved):

PR #171, opened 2022-04-25
Expand build disk

Itxaka:
- Small fix for image existing, just refuse to overwrite by default
 - Supports output for GCE and azure images
 - No external tools required, all done in pure golang!
 - Autoclean middle image if generating a gce/azure image
 - Adds new command transform-disk
 - Transform-disk uses the same code as build-disk
 - Transform-disk allows to keep the base raw image
 - Adds some vhd tooling for the build/transform stuff

### DIFF
--- a/cmd/convert-disk.go
+++ b/cmd/convert-disk.go
@@ -18,10 +18,10 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/rancher-sandbox/elemental/pkg/action"
 	"strings"
 
 	"github.com/rancher-sandbox/elemental/cmd/config"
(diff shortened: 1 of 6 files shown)


Itxaka:
second part of the build-disk. This should cover all existing functionality of the raw image stuff well enough to drop the scripts.

As device stuff is a bit different, that is not covered in here and I wonder if it should as the device stuff is more closer to a normal install than to a vanilla/recovery disk image....

Itxaka:
Ready to review @mjura @mudler @davidcassany

davidcassany (approved):
Looks good here :+1:

PR #174, opened 2022-04-26
Add default values if config is empty

Itxaka:

### DIFF
--- a/pkg/action/build-disk.go
+++ b/pkg/action/build-disk.go
@@ -19,6 +19,7 @@ package action
 import (
 	"archive/tar"
 	"compress/gzip"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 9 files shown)


pkg/action/build-disk.go
```diff
@@ -35,6 +36,23 @@ var GB = 1024 * MB
...
+	}
+
+	if len(cfg.RawDisk[arch].Packages) == 0 {
+		msg := fmt.Sprintf("no packages in the config for arch %s", arch)
+		cfg.Logger.Error(msg)
+		return errors.New(msg)
+	}
+
+	if len(cfg.Config.Repos) == 0 {
+		msg := fmt.Sprintf("no varepositories configured for arch %s", arch)
```
mudler:
```suggestion
		msg := fmt.Sprintf("no repositories configured for arch %s", arch)
```

pkg/constants/constants.go
```diff
@@ -113,6 +113,8 @@ const (
 
 	// Eject script
 	EjectScript = "#!/bin/sh\n/usr/bin/eject -rmF"
+
+	BuildDiskDefaultRepo = "quay.io/costoolkit/releases-green"
```
mudler: Can't we use the same default for both ISO and build disk here?

mudler (approved):
Just some nits but looks good!

pkg/action/build-disk.go
Itxaka: just noticed this locally.... spaghetti fingers!

pkg/constants/constants.go
Itxaka: I thougth about it, then thougth about if we want to change one vs the other....I guess it could be difficult to need 2 different default repos for the same stuff...Will change it

PR #177, opened 2022-04-27
Improve build-iso flags

davidcassany:
This commit changes `--iso.label` flag to `--label` and drops
`--iso.image` and `--iso.uefi` flags since this are unlikely to be used
and they can still be set by using a custom manifest.yaml.

Finally it also changes `--repo` flag behavior to append repos on top
of whatever is defined in the manifest or by default. Appended
repositories with `--repo` flag are always having highest priority.


davidcassany:
In addition this is also assuming the maximum repo priority to be 1.

Itxaka (approved):

mjura (commented):
LGTM

PR #181, opened 2022-05-02
Set default config dir to /etc/elemental

mudler:
By default we used to set "" as config-dir, this in turn made it scan yaml files from the "/". By setting also a default, we allow derivatives to store yaml config files directly in `/etc/elemental/conf.d/*`

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -88,28 +88,31 @@ func ReadConfigRun(configDir string, mounter mount.Interface) (*v1.RunConfig, er
 		}
 	}
 
-	if exists, _ := utils.Exists(cfg.Fs, configDir); exists {
-		viper.AddConfigPath(configDir)
-		viper.SetConfigType("yaml")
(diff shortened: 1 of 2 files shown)


mudler:
Context: https://github.com/rancher-sandbox/cOS-toolkit/pull/1269#issuecomment-1113775170 ARM builds are currently failing tests now due to a typo. Besides the typo of the flavor/arch I think there is also a separate issue, that is https://github.com/rancher-sandbox/cOS-toolkit/blob/master/packages/cloud-config/oem/02_upgrades.yaml#L19 is not took into account anymore after switching to elemental-cli. It was passing x86_64 tests, but not arm (which is good that we cought up)

davidcassany (approved):
Looks good

I am wondering if shouldn't we also rename the `/etc/cos` to `/etc/elemental` at some point.

Itxaka (approved):

PR #182, opened 2022-05-04
Introduce generic arch flag for build commands

Itxaka:

### DIFF
--- a/cmd/build-disk.go
+++ b/cmd/build-disk.go
@@ -67,22 +67,21 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			cmd.SilenceUsage = true
 			cmd.SilenceErrors = true // Do not propagate errors down the line, we control them
 			imgType, _ := cmd.Flags().GetString("type")
-			archType, _ := cmd.Flags().GetString("arch")
 			output, _ := cmd.Flags().GetString("output")
 			oemLabel, _ := cmd.Flags().GetString("oem_label")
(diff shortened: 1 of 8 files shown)


Itxaka:
@davidcassany would something like this work? Or were you thinking something more different?

This is a very simplistic arch implementation, as we only modify the default repos to add the proper ones and allow cross-build of iso (disk was already possible). 

Another implementation could be to add the arch to the repositories, so we could enable/disable them based of the required given arch, like luet does, but that would be more complicated and would require more changes to our whole thing.

davidcassany:
> This is a very simplistic arch implementation, as we only modify the default repos to add the proper ones and allow cross-build of iso (disk was already possible).

Yes this is the major and first step
 
> Another implementation could be to add the arch to the repositories, so we could enable/disable them based of the required given arch, like luet does, but that would be more complicated and would require more changes to our whole thing.

Rather than another implementation, it should be added on top. So it can easily be a second round and also get rid of nested repositories in `build-disk`. I'd do that as a follow up once all other stuff is settled.

davidcassany (approved):
Looks good to me

mudler:
uhm. weird. after this was merged to master tests started to fail? https://github.com/rancher-sandbox/elemental/runs/6287438234?check_suite_focus=true

Itxaka:
No really, that command test has a weird fetiah for failing randomly, but only on ci

Itxaka:
Just retriggered and now it passed... Probably should open a bug

PR #185, opened 2022-05-09
Add --local option to install, build-iso, upgrade

### LINKED ISSUE #1279: elemental build-iso: `--local` flag to pick-up images from docker daemon
Currently elemental build-iso doesn't load images from the local docker daemon (as previously `luet-makeiso` was capable of), breaking the development workflow https://github.com/rancher-sandbox/elemental/blob/edc412b0dd1c219a07e023d16cafb6251f7280cd/pkg/elemental/elemental.go#L337

This card is about to port back that functionality to `elemental build-iso`

mjura:
Add --local option to pull-image, install, build-iso
and upgrade subcommands which uses Unpack() function.

To use image from local cache set option --local

Fixes https://github.com/rancher-sandbox/cOS-toolkit/issues/1279

Signed-off-by: Michal Jura <mjura@suse.com>

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -334,7 +334,7 @@ func (c *Elemental) CopyImage(img *v1.Image) error { // nolint:gocyclo
 				return err
 			}
 		}
-		err = c.config.Luet.Unpack(img.MountPoint, img.Source.Value(), false)
+		err = c.config.Luet.Unpack(img.MountPoint, img.Source.Value(), c.config.LocalImage)
 		if err != nil {
 			return err
 		}
--- a/pkg/luet/luet.go
+++ b/pkg/luet/luet.go
@@ -18,6 +18,7 @@ package luet
 
 import (
 	"crypto/md5"
+	"errors"
 	"fmt"
 	"runtime"
 	"strings"
@@ -135,21 +136,28 @@ func NewLuet(opts ...Options) *Luet {
 }
 
 func (l Luet) Unpack(target string, image string, local bool) error {
-	l.log.Infof("Unpacking docker image: %s", image)
-	if !local {
+	l.log.Infof("Unpacking a container image: %s", image)
+
+	if local {
+		l.log.Infof("Using an image from local cache")
+		info, err := docker.ExtractDockerImage(l.context, image, target)
+		if err != nil {
+			if strings.Contains(err.Error(), "reference does not exist") {
+				return errors.New("Container image does not exist locally")
+			}
+			return err
+		}
+		l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
+	} else {
+		l.log.Infof("Pulling an image from remote repository")
 		info, err := docker.DownloadAndExtractDockerImage(l.context, image, target, l.auth, l.VerifyImageUnpack)
 		if err != nil {
 			return err
 		}
 		l.log.Infof("Pulled: %s %s", info.Target.Digest, info.Name)
 		l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
-	} else {
-		info, err := docker.ExtractDockerImage(l.context, image, target)
-		if err != nil {
-			return err
-		}
-		l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
 	}
+
 	return nil
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+.vscode/
 bin/
 coverage.txt
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -147,6 +147,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	addArchFlags(c)
 	addCosignFlags(c)
 	addSquashFsCompressionFlags(c)
+	addLocalImageFlag(c)
 	return c
 }
 
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -49,6 +49,11 @@ func addSharedInstallUpgradeFlags(cmd *cobra.Command) {
 	addPowerFlags(cmd)
 }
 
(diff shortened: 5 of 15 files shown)


pkg/luet/luet.go
```diff
@@ -128,23 +130,58 @@ func NewLuet(opts ...Options) *Luet {
...
+		if imgLocal {
+			info, err := docker.ExtractDockerImage(l.context, image, target)
+			if err != nil {
+				return err
+			}
+			l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
+			return nil
+		}
+
+		l.log.Infof("An image does not exist locally, pulling from remote")
```
mudler: I'm not completely sure about this, failing back looks good but it might not be always what we want. Let's suppose I'm iterating locally against a tag which was already published and accidentally I remove it from docker, It would go straight building and I would not be able to notice that I'm building an image that I did not want to.

mudler: maybe instead of taking a `local` flag, should it take a `localOnly` boolean?

pkg/luet/luet.go
```diff
@@ -128,23 +130,58 @@ func NewLuet(opts ...Options) *Luet {
...
 			return err
 		}
-		l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
+
+		if imgLocal {
+			info, err := docker.ExtractDockerImage(l.context, image, target)
+			if err != nil {
+				return err
+			}
+			l.log.Infof("Size: %s", units.BytesSize(float64(info.Target.Size)))
```
Itxaka: Should probably log here that we found the image locally just in case.

pkg/luet/luet.go
Itxaka: Agreed, this should be flag-able so users can control. I do think that making it the default might be good though

pkg/elemental/elemental.go
```diff
@@ -334,7 +334,7 @@ func (c *Elemental) CopyImage(img *v1.Image) error { // nolint:gocyclo
 				return err
 			}
 		}
-		err = c.config.Luet.Unpack(img.MountPoint, img.Source.Value(), false)
+		err = c.config.Luet.Unpack(img.MountPoint, img.Source.Value(), true)
```
Itxaka: umm, I wonder why this is not configurable and its hardcoded....

mudler: yes, I wondered as well too. I think it is a side-effect of the fallback logic, it is always enabled as it tries local first.

mjura: This is because `elemental install` doesn't support `--local` option, we can add it if you think that it is needed

pkg/luet/luet.go
mjura: I have added information about it, thank you

pkg/luet/luet.go
mjura: Following logic is the most natural for me:
1. To use image, check it in local cache first
2. If not present then try to pull it from remote

Additional flags will only confuse our users. But if you want to have flag `--local-only` or `--remote-only` we can add them

mudler: the problem with this approach is that you don't guarantee anymore what is the result of your build artifact unless you look up at the logs, so any call to the same command could potentially yields different results.

Being too smart here might just be more problematic - while developing I would never want `build-iso` to pull a remote image unless I specifically tell it to.

pkg/elemental/elemental.go
Itxaka: umm, Im not sure if that is good or not. I mean, it would work greatly for air-gapped system maybe? You could build an iso with all the images on a local docker registry and then install systems from there?

But I think thats out of scope of this card, I would leave it as is.

pkg/luet/luet.go
mjura: After our discussion, current logic will be as follows:
1. To use image, check it in local cache first
2. If not present then try to pull it from remote
3. If --local-only option was set then don't use remote

pkg/elemental/elemental.go
mjura: After changes, it will work for air-gapped environments, we have added --local-only option which will rely only on local cache.

Itxaka: Awesome, but now I wonder if we want 2 different flags for this, like --local for try local then remote and --local-only for only try local.

Anyway, we can come back to this later to add that.

Itxaka (approved):

pkg/elemental/elemental.go
mudler: This should be false by default in this case here, right? otherwise we would fail upgrading as we would always try from local cc: @Itxaka

Itxaka: EDIT: oh no, you are rigth, we are passing the localOnly yes, so this will fail on install, upgrade, reset? and pull-image unless you unset --localOnly

Itxaka: As we dropped using the real luet for upgrade tests, this path is not properly tested, we should probably try to test this further.

mudler: ok, so yes, this needs to be properly instrumented otherwise hardcoded value here have a different meaning depending on what context this is being used on. we can't hardcode any value, we must pass this accordingly depending on how it is being used.

mudler (changes requested):
`elemental.CopyImage` can't take an hardcoded value here, otherwise it will just work during building the iso, but it would fail during upgrades/install and all the other scenarios relying on `CopyImage`.

pkg/elemental/elemental.go
mjura: ok, my idea would be then to add --local-only flag to elemental install, upgrade and build-iso and then reuse it in CopyImage function. What do you think ?

Itxaka:
Easy solution here would be to add a --local flag to all the commands using the unpack and pass that down to the unpacker

mjura:
Ok, let's add local flag to all the commands using the unpack and pass that down to the unpacker

mudler (approved):
Looks good here!

PR #187, opened 2022-05-09
Squashfs compression

Itxaka:

### DIFF
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -145,6 +145,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().StringArray("repo", []string{}, "A repository URI for luet. Can be repeated to add more than one source.")
 	addArchFlags(c)
 	addCosignFlags(c)
+	addSquashFsCompressionFlags(c)
 	return c
 }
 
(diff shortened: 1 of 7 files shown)


Itxaka:
Not really happy with this. Its the most straightforward path, to allow a full dump of the options directly, but its kind of crappy. On the other hand I dont really want to introduce the 5 compressors + 9 different options divided by compressor in the flags as it will be a mess. Plus validation, usage, and so on.... not sure to be honest what to do with this.

Itxaka:
I mean, look at this:

```
	lzo
	  -Xalgorithm <algorithm>
		Where <algorithm> is one of:
			lzo1x_1
			lzo1x_1_11
			lzo1x_1_12
			lzo1x_1_15
			lzo1x_999 (default)
	  -Xcompression-level <compression-level>
		<compression-level> should be 1 .. 9 (default 8)
		Only applies to lzo1x_999 algorithm
```
This means that in the case of lzo1x_999 suboption of the lzo compressor, Xcompression-level is valid and should be use....what a mess :/

mudler:
yeap, indeed passing by args looks like the only sane way without having to re-implement the whole squashfs CLI logic, no need imho, best to keep it simple.

mudler (approved):

PR #188, opened 2022-05-09
Squashfs compression

Itxaka:

### DIFF
--- a/pkg/types/v1/config.go
+++ b/pkg/types/v1/config.go
@@ -44,7 +44,7 @@ type Config struct {
 	CosignPubKey              string       `yaml:"cosign-key,omitempty" mapstructure:"cosign-key"`
 	Repos                     []Repository `yaml:"repositories,omitempty" mapstructure:"repositories"`
 	Arch                      string       `yaml:"arch,omitempty" mapstructure:"arch"`
-	SquashFsCompressionConfig []string
+	SquashFsCompressionConfig []string     `yaml:"squash-compression,omitempty" mapstructure:"SquashFsCompressionConfig"`
 }
 
 // RunConfig is the struct that represents the full configuration needed for install, upgrade, reset, rebrand.
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -39,6 +39,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 		Short: "builds bootable installation media ISOs",
 		Args:  cobra.MaximumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			bindSquashFsCompressionFlags(cmd)
 			_ = viper.BindPFlags(cmd.Flags())
 			if addCheckRoot {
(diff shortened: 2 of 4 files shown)


pkg/types/v1/config.go
```diff
@@ -44,7 +44,7 @@ type Config struct {
 	CosignPubKey              string       `yaml:"cosign-key,omitempty" mapstructure:"cosign-key"`
 	Repos                     []Repository `yaml:"repositories,omitempty" mapstructure:"repositories"`
 	Arch                      string       `yaml:"arch,omitempty" mapstructure:"arch"`
-	SquashFsCompressionConfig []string
+	SquashFsCompressionConfig []string     `yaml:"squash-compression,omitempty" mapstructure:"squash-compression"`
```
mudler: :sweat_smile:

mudler (approved):

Itxaka:
somehow, somewhere, all at once, I cant make this work at all. `SquashFsCompressionConfig` always comes empty, even when the flag has the correct params...

@mudler any ideas?

Itxaka:
oh god, I think its because the `SquashFsCompressionConfig` is under `config`, so it cannot bind to the cfg we are loading into, becuase its not on the first level, i.e. is under `RunConfig.Config.SquashFsCompressionConfig`


blergh

mudler:
> oh god, I think its because the `SquashFsCompressionConfig` is under `config`, so it cannot bind to the cfg we are loading into, becuase its not on the first level, i.e. is under `RunConfig.Config.SquashFsCompressionConfig`
> 
> blergh

mmm that would be a problem also for the other options like repositories. That would explain why testing #181 I could not set repositories.

I wonder how we ended up with this limitation? In luet this mapping works

mudler:
(not related to this change) What I miss in general is that we cannot marshal our config as contains interfaces and as thus options are not visible

Itxaka:
> I wonder how we ended up with this limitation? In luet this mapping works

Im not sure, it may be the datastructure lib that we use? Maybe its just a side effect of our config disorganization

mudler:
> > I wonder how we ended up with this limitation? In luet this mapping works
> 
> Im not sure, it may be the datastructure lib that we use? Maybe its just a side effect of our config disorganization

I just double checked - I don't see any big difference, except that the config in luet does not contain any interface. there are just a couple of different gotchas as `luet` runs the viper binding parts all in `PersistentPreRun` and it goes unmarshalling right before when running the commands https://github.com/mudler/luet/blob/b5da2fa7b4819b9fffa649c9d22a5779518fb146/cmd/util/config.go#L150.

Itxaka:
There is also some constraints in how we are passing configs around which kind of restricts you on adding stuff somewher. For example, this squashfs options would fit rigth into the RunConfig, but as we dont pass that to the `prepareISORoot` (we pass a v1.Config instead) I had to put it on the Config struct directly or change the whole thing.

I guess that once we do the whole config rebuild, we can re-tackle the viper flag stuff to make it better and will probably work much nicer

PR #189, opened 2022-05-09
Refactor runconfig

davidcassany:
I know this is a huge PR... It was impossible to refactor the configuration data without touching everything... :disappointed:

Gonna add some comments on the code to explain changes and relate a bullet list of behavior changes.

Related to #33


davidcassany:
With this PR we gain the following:

* fine grain setup of install, reset and upgrade commands from a yaml file. Consider `config.yaml.example` file:
* structured human readable yaml configuration file
* strict configuration prio: defaults (low) -> config files -> environment variables -> client flags (high)
  everything starts from defaults and then changes, if any, applied on on top in the above order
* all configuration is loaded at startup according the above criteria
  it opens the door to run early sanity checks if needed. Ideally this would be a postDecodeHook, but mapstructure does not support it yet.

Config example:
```yaml
# DISCLAIMER: this document show most of all available options. Elemental
# uses defaults if any of this value is missing. Values shown here
# only represent an example and they should not be used as defaults.

# installation configuration for 'install' command
install:
  # target is the only value that has no default, it must be provided by
  # config, flags or env variables.
  target: /dev/sda

  # basic disk configs for partitioning ('efi|bios' and 'gpt|msdos')
  firmware: efi
  part-table: gpt

  # partitions setup
  # only 'oem', 'recovery', 'state' and 'persistent' objects allowed
  # size in MiB
  partitions:
    oem:
      label: COS_OEM
      size: 60
      fs: ext4
    recovery:
      label: COS_RECOVERY
      size: 4096
      fs: ext4

  # no-format: true skips any disk partitioning and formatting
  # if set to true installation procedure will error out if expected
  # partitions are not already present within the disk.
  no-format: false

  # if no-format is used and elemental is running over an existing deployment
  # force cane be used to force installation.
  force: false

  # use this iso as installation media (overwrites 'system.uri' and 'recoverys-system.uri'
  # according to the ISO contents.
  iso: https://my.domain.org/some/powerful.iso

  # main OS image
  # size in MiB
  system:
    label: COS_ACTIVE
    size: 1024
    fs: ext2
    uri: docker:some.registry.org/cos/image:latest

  # recovery OS image
  recovery-system:
    fs: squashfs
    uri: channel:recovery/cos

  # extra cloud-init config file URI to include during the installation
  cloud-init: "https://some.cloud-init.org/my-config-file"

  # grub menu entry, this is the string that will be displayed
  grub-default-entry: cOS

  # tty console to add into the kernel parameters
  tty: ttyS0



# configuration for the 'reset' command
reset:
  # if set to true it will format persistent partitions ('oem 'and 'persistent')
  reset-persistent: false

  # OS image used to reset disk
  # size in MiB
  system:
    label: COS_ACTIVE
    size: 1024
    fs: ext2
    uri: docker:some.registry.org/cos/image:latest

  # grub menu entry, this is the string that will be displayed
  grub-default-entry: cOS

  # tty console to add into the kernel parameters
  tty: ttyS0



# configuration used for the 'ugrade' command
upgrade:
  # if set to true upgrade command will upgrade recovery system instead
  # of main active system
  upgrade-recovery: false

  # image used to upgrade main OS
  # size in MiB
  system:
    uri: channel:system/cos

  # image used to upgrade recovery OS
  # recovery images can be set to use squashfs
  recovery-system:
    fs: squashfs
    uri: channel:recovery/cos

# use cosing to validate images from container registries
cosign: true
# cosign key to used for validation
cosign-key: myKey

# attempt a verify process
no-verify: false

# fail on cloud-init hooks errors
strict: false

# Additional paths for look for cloud-init files
cloud-init-paths:
- "/some/path"

#  compress options passed to `mksquashfs` tool
squash-compression: ["-comp", "xz", "-Xbcj", "x86"]
  
# reboot/power off when done
reboot: false
poweroff: false
```

davidcassany:
What we loose in this current PR:

* environment variables gone or completely changed
  ~I did not aim to code elemental variables here as it seamed already too much and think it would be nice to strictly
  manage subset of env variables rather than mapping the whole yaml. So this is lost because it is a TODO.~
  Suported environment variables are explicitly passed on each command. See: https://github.com/rancher-sandbox/elemental/blob/37d42a6614a8ac6ecbbcbe2907f8ab00db0f24b9/cmd/install.go#L68-L74

* partition layout install flag
  It does not make sense anymore. This is was an alternate way of providing custom layout options by using a cloud-init
  plugin. Now all this is still possible by using the config file.
  
* upgrade channel config (luet package to upgrade to)
The file `/etc/cos-upgrade-image` is no longer being read by elemental. It relies on the `/etc/elemental/config.yaml` config file to configure upgrades.

* There is no rebrand procedure executed on recovery upgrades
  Rebrand currently only means setting the display grub menu entry. During upgrade this value is taken from 
  `/etc/os-release`, I doubt it makes much sense to apply it for recovery upgrades as this should come from
  the active upgrades. Removing this particular use case simplifies significantly upgrades logic.
  
* Using custom labels (connected to the missing environment variables)
  The use of custom labels can't be set at install, upgrade or reset time. This is a known limitation of cOS-toolkit,
  basically because some of these labels are hardcoded in few critical places of our packages, such as grub config
  and so on. Those files are templated and rendered at build time. 
  
  Note this means `/etc/cos/config` is still required at build time, however is no longer used by the elemental client in any way. IMHO we should reconsider this need at build time and find strategies to apply it at install time, probably storing these values under a grubenv file?

cmd/config/config.go
```diff
@@ -33,10 +34,69 @@ import (
...
+var decodeHook = viper.DecodeHook(
+	mapstructure.ComposeDecodeHookFunc(
+		UnmarshalerHook(),
+		mapstructure.StringToTimeDurationHookFunc(),
+		mapstructure.StringToSliceHookFunc(","),
+	),
+)
+
+type Unmarshaler interface {
+	CustomUnmarshal(interface{}) (bool, error)
```
davidcassany, line 51: Any type implementing this interface can eventually decode a `map[string]interface{}` into the type or just run some logic before the default decoder is applied, this is controller with the returned boolean. If it returns true the mapstructure continues with the default decoder logic after running the custom logic. If it returns false it assumes the decoding is completely done after this custom logic.

cmd/config/config.go
```diff
@@ -33,10 +34,69 @@ import (
...
+// setDecoder sets ZeroFields mastructure attribute to true
+func setDecoder(config *mapstructure.DecoderConfig) {
+	// Make sure we zero fields before applying them, this is relevant for slices
+	// so we do not merge with any already present value and directly apply whatever
+	// we got form configs.
+	config.ZeroFields = true
+}
+
+// BindGivenFlags binds to viper only passed flags, ignoring any non provided flag
+func bindGivenFlags(flagSet *pflag.FlagSet) {
```
davidcassany: I noted there is no way in viper to bind only the flags passed at runtime, this is a problem if we want to distinguish between set or unset. This logic only binds the flags that were actually provided, since defautls are all managed in `pkg/config` in this PR and not via flag, since flags are just a small set of available options.

cmd/config/config.go
```diff
@@ -134,6 +194,58 @@ func ReadConfigRun(configDir string, mounter mount.Interface) (*v1.RunConfig, er
 	return cfg, nil
 }
 
+func viperRemapKeys(vp *viper.Viper, keyRemap map[string]string) {
```
davidcassany: this is to map flags to nested structres with or without the same name. I could not find any other way to do that in viper other than remapping global flags to another sub viper instance.

cmd/flags.go
```diff
@@ -40,15 +40,38 @@ func addPowerFlags(cmd *cobra.Command) {
...
+	cmd.Flags().String("recovery-system", "", "Sets the recovery image source and its type (e.g. 'docker:registry.org/image:tag')")
+}
+
+// addResetFlags add flags shared between reset, install and upgrade
+func addResetFlags(cmd *cobra.Command) {
 	cmd.Flags().String("directory", "", "Use directory as source to install from")
+	_ = cmd.Flags().MarkDeprecated("directory", "'directory' is deprecated please use 'system' instead")
+
 	cmd.Flags().StringP("docker-image", "d", "", "Install a specified container image")
+	_ = cmd.Flags().MarkDeprecated("docker-image", "'docker-image' is deprecated please use 'system' instead")
```
davidcassany, line 54: This is deprected as it can be better expressed as: `--system.uri docker:registry.org/image:tag`

cmd/flags.go
```diff
@@ -40,15 +40,38 @@ func addPowerFlags(cmd *cobra.Command) {
...
 // addSharedInstallUpgradeFlags add flags shared between install, upgrade and reset
 func addSharedInstallUpgradeFlags(cmd *cobra.Command) {
+	addResetFlags(cmd)
+	cmd.Flags().String("recovery-system", "", "Sets the recovery image source and its type (e.g. 'docker:registry.org/image:tag')")
+}
+
+// addResetFlags add flags shared between reset, install and upgrade
+func addResetFlags(cmd *cobra.Command) {
 	cmd.Flags().String("directory", "", "Use directory as source to install from")
+	_ = cmd.Flags().MarkDeprecated("directory", "'directory' is deprecated please use 'system' instead")
```
davidcassany, line 51: This is deprected as it can be better expressed as: `--system.uri dir:/some/path`

cmd/flags.go
```diff
@@ -40,15 +40,38 @@ func addPowerFlags(cmd *cobra.Command) {
...
+
+	cmd.Flags().String("system", "", "Sets the system image source and its type (e.g. 'docker:registry.org/image:tag')")
 	cmd.Flags().BoolP("no-verify", "", false, "Disable mtree checksum verification (requires images manifests generated with mtree separately)")
 	cmd.Flags().BoolP("strict", "", false, "Enable strict check of hooks (They need to exit with 0)")
 
 	addCosignFlags(cmd)
 	addPowerFlags(cmd)
 }
 
+func adaptDockerImageAndDirectoryFlagsToSystem() {
```
davidcassany: This is just to be backward compatible old flags.

pkg/action/install.go
```diff
@@ -17,228 +17,147 @@ limitations under the License.
...
 			extraMounts[oem.MountPoint] = "/oem" //nolint:goconst
 		}
-		return ChrootHook(config, hook, config.Images.GetActive().MountPoint, extraMounts)
+		return ChrootHook(&i.cfg.Config, hook, i.cfg.Strict, i.spec.ActiveImg.MountPoint, extraMounts, i.cfg.CloudInitPaths...)
 	}
-	return Hook(config, hook)
+	return Hook(&i.cfg.Config, hook, i.cfg.Strict, i.cfg.CloudInitPaths...)
 }
 
-// InstallImagesSetup defines the parameters of active, passive and recovery
```
davidcassany, line 47: Defaults logic is moved under `pkg/config`

pkg/action/install_test.go
```diff
@@ -0,0 +1,376 @@
+/*
```
davidcassany, line 1: split `runtime_test.go` into `install_test.go`, `reset_test.go` and `upgrade_test.go` as the tests file was growing a lot.

pkg/config/config.go
```diff
@@ -126,77 +139,339 @@ func NewConfig(opts ...GenericOptions) *v1.Config {
...
+		Config: *config,
 	}
-	// Set defaults if empty
-	if r.GrubConf == "" {
-		r.GrubConf = cnst.GrubConf
+	return r
+}
+
+// NewInstallSpec returns an InstallSpec struct all based on defaults and basic host checks (e.g. EFI vs BIOS)
+func NewInstallSpec(cfg v1.Config) *v1.InstallSpec {
```
davidcassany, line 165: This creator applies defaults into `v1.InstallSpec` based on basic on constants and basic host checks (like running from install media or not, efi firmware capabilities, etc.).

pkg/config/config.go
```diff
@@ -126,77 +139,339 @@ func NewConfig(opts ...GenericOptions) *v1.Config {
...
+
+	statePart, ok := partitionMap[constants.StatePartName]
+	if !ok {
+		return nil, fmt.Errorf("state partition not found")
+	} else if statePart.MountPoint == "" {
+		statePart.MountPoint = constants.StateDir
+	}
+
+	// TODO find a way to pre-load current state values such as SystemLabel
+	bootedRec := utils.BootedFrom(cfg.Runner, constants.RecoverySquashFile) || utils.BootedFrom(cfg.Runner, constants.SystemLabel)
```
davidcassany: This `constants.SystemLabel` is the label of the recovery image if not squashed, if this is customized we do not have a proper way to know it from an upgrade command perspective, since we are not keeping anywhere installation details. We could have some heuristics to guess it, but I think this not the appropriate road. I miss state metadata once the system is installed (labels, versions and any other metadata that could be handy).

pkg/elemental/elemental.go
```diff
@@ -30,98 +30,82 @@ import (
 
 // Elemental is the struct meant to self-contain most utils and actions related to Elemental, like installing or applying selinux
 type Elemental struct {
-	config *v1.RunConfig
+	config *v1.Config
```
davidcassany, line 33: Now elemental is only coupled to  generic stuff `v1.Config`, it can be easily reused into build commands too.

pkg/types/v1/config.go
```diff
@@ -44,76 +46,223 @@ type Config struct {
...
+		if !pm.validName(k) {
+			// Removing the invalid entry causes to completely ignore it
+			delete(m, k)
+		}
+	}
+	return true, nil
+}
+
+// OrderedByLayoutPartitions sorts partitions according to the default layout
+func (pm PartitionMap) OrderedByLayoutPartitions() PartitionList {
```
davidcassany: maps in golang are explicitly unordered (order randomly changes on each initialization) so something like:

```yaml
partitiions:
   oem: 
   # some data
   persistent:
   # some data
```
is map and does not keep partitions order, so partitions order in yaml is irrelevant and they are statically ordered by this method.

pkg/types/v1/config.go
```diff
@@ -44,76 +46,223 @@ type Config struct {
...
+		if part, present = pm[name]; present {
+			partitions = append(partitions, part)
+		}
+	}
+	return partitions
+}
+
+// OrderedByMountPointPartitions sorts partitions according to its mountpoint, ignores partitions
+// with an empty mountpoint, these are excluded
+func (pm PartitionMap) OrderedByMountPointPartitions(descending bool) PartitionList {
```
davidcassany: Orders partitions by its mountpoint. This is relevant if mountpoints are nested, so we ensure partent is mounted first. `descending` bool flag set reverse order, so it can be used to unmount.

pkg/utils/common.go
```diff
@@ -254,21 +255,66 @@ func LoadEnvFile(fs v1.FS, file string) (map[string]string, error) {
...
+			}
+		}()
+	}
+	return Exists(config.Fs, filepath.Join(cnst.LiveDir, "cOS", cnst.RecoverySquashFile))
+}
+
+// GetElementalTempDir returns the dir for storing temporary files. It will respect TMPDIR and use
+// that if exists, fallback to persistent partition if mounted and finally the default /tmp/ dir.
+// This is specially useful when big temporary areas are required, such as creating squashfs images.
+func GetElementalTempDir(config *v1.Config) string {
```
davidcassany: This is the former `GetUpgradeTempDir` I renamed it since it is also used in  `elemental.DeployImage` to support squashfs images (we need the root under a temporary dir in a disk before squashing).

pkg/utils/grub.go
```diff
@@ -41,52 +40,40 @@ func NewGrub(config *v1.RunConfig) *Grub {
 }
 
 // Install installs grub into the device, copy the config file and add any extra TTY to grub
-func (g Grub) Install() (err error) { // nolint:gocyclo
+func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool) (err error) { // nolint:gocyclo
```
davidcassany, line 43: I am passing most of grubs config by parameter instead of passing a whole config struct so it can be used in any context (not only for install command struct).

davidcassany:
New unit tests to raise coverage up to the prior level are still missing. Mostly under `pkg/config` and `pkg/types/v1` and some utilities.

pkg/utils/common.go
Itxaka: This has been changed unfortunately

cmd/flags.go
Itxaka: Don't we need to bind this value to the new one so it still works even if deprecated?

cmd/install.go
```diff
@@ -37,7 +39,6 @@ func NewInstallCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 		Args:  cobra.MaximumNArgs(1),
```
Itxaka, line 39: As far as I understood, bonding flags put of preRun didn't work on my tests. How does this work now?

cmd/config/config.go
```diff
@@ -175,17 +283,20 @@ func configLogger(log v1.Logger, vfs v1.FS) {
 	log.Infof("Starting elemental version %s", v.Version)
 }
 
-func viperReadEnv() {
+func viperReadEnv(vp *viper.Viper, keyMap map[string]string) {
```
Itxaka: do we really need to pass viper here? AFAIK is a singleton to viper.GetViper() should get the global instance? https://github.com/spf13/viper/blob/65293ecec2f2d7237de9261f637c42d451f94a3d/viper.go#L870

cmd/flags.go
Itxaka: already done in a different place :D

cmd/config/config.go
```diff
@@ -76,7 +131,7 @@ func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig
 	return cfg, nil
 }
 
-func ReadConfigRun(configDir string, mounter mount.Interface) (*v1.RunConfig, error) {
+func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interface) (*v1.RunConfig, error) {
```
Itxaka, line 134: why do we need the `flags` here for?

cmd/config/config.go
Itxaka: oh wait, wtf, we are getting different vipers here! Why is that needed?

Itxaka:
will continue reviewing later but its looking good on my side. As usual, questions about flags and viper of course. I also dont really like that we need to pass the flagset so much around, but I think we cannot use a global viper for this, so we need to do that.

cmd/flags.go
davidcassany: Yes, it gets bound by `bindGivenFlags` method from `cmd/config/config.go`. This method binds runtime given flags to a viper instance. If we bind all flags it sucks because then default values from flags are applied in every field. This is problematic if we want to load config files first and then flags on top.

pkg/utils/common.go
davidcassany: Yes and already adapted no problem ;)

cmd/config/config.go
davidcassany: To access into nested structures I used another viper instance see the following:

```golang
func ReadInstallSpec(r *v1.RunConfig, flags *pflag.FlagSet, keyRemap map[string]string) (*v1.InstallSpec, error) {
	install := config.NewInstallSpec(r.Config)
	vp := viper.Sub("install")  // loads the install object from the global instance into a new viper instance
	if vp == nil {
		vp = viper.New()  // use an empty one if there was no 'install' object in configs
	}
	// Bind install cmd flags
	bindGivenFlags(vp, flags) // binds given subcommand flags to the new viper instance
	// Bind install env vars
	viperReadEnv(vp, keyRemap) // binds custom env vars to the new viper instance
	err := vp.Unmarshal(install, setDecoder, decodeHook) // unmarshll the nested 'instal'l object to the install struct
	r.Logger.Debugf("Loaded install spec: %+v", install)
	return install, err
}
```

This is the way I found to keep `RunConfig` struct and `InstallSpec` structs completely independent, one does not include the other.

cmd/config/config.go
davidcassany: To bind given ones at runtime as noted in https://github.com/rancher-sandbox/elemental/pull/189#discussion_r869497679 

So for a call like `elemental install --force /dev/sda` only `--force` flag gets bound to viper.

The configs are loaded in this order:
1. Create default object with a constructor (aka `config.NewRunConfig()`)
2. Bind config file to viper (with `omitempty` only provided values will be applied)
3. Bind given flags
4. Bind environment variables

cmd/config/config.go
Itxaka: niiice!

pkg/types/v1/config.go
```diff
@@ -41,118 +46,227 @@ type Config struct {
...
+		if p.Label == label {
+			return p
+		}
+	}
+	return nil
 }
 
-func (im ImageMap) SetPassive(img *Image) {
-	im[constants.PassiveImgName] = img
+type ElementalPartitions struct {
```
davidcassany, line 197: I deeply dislike having a struct like that, to me this is a clear `map[string]*Partition` container, however it turned to be really complex to unmarshal and merge data into a map. Viper and mapstructure limitations were too hard to workaround, so I finally opted for the dumb and explicit workaround. At least this way the mapping between struct/yaml is clear and explicit.

pkg/types/v1/luet.go
```diff
@@ -19,4 +19,6 @@ package v1
 type LuetInterface interface {
 	Unpack(string, string, bool) error
 	UnpackFromChannel(string, string, ...Repository) error
+	SetPlugins(...string)
```
davidcassany, line 22: This interface extension it to allow setting plugins after luet initialization. This way we can initialize a `v1.Config.Luet` without any specific plugin and add plugins according to elemental client configuration or flags in a later stage.

cmd/config/config.go
```diff
@@ -76,64 +131,122 @@ func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig
...
 	cfg := config.NewRunConfig(
 		config.WithLogger(v1.NewLogger()),
 		config.WithMounter(mounter),
 	)
 
 	configLogger(cfg.Logger, cfg.Fs)
 
-	cfgDefault := []string{"/etc/os-release", "/etc/cos/config", "/etc/cos-upgrade-image"}
```
mudler, line 87: that is totally fine to drop, altough this needs to be reflected in the docs https://rancher-sandbox.github.io/cOS-toolkit/docs/customizing/general_configuration/

mudler: I'm still thorn to keep one path excluded from the fs-layout, that is to prevent users need to overwrite the config that might be shipped by another derivative.  maybe keep `/etc/cos/config`?

cmd/config/config.go
```diff
@@ -76,64 +131,122 @@ func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig
...
+	// Bind reset cmd flags
+	bindGivenFlags(vp, flags)
+	// Bind reset env vars
+	viperReadEnv(vp, "RESET", keyRemap)
+	err = vp.Unmarshal(reset, setDecoder, decodeHook)
+	r.Logger.Debugf("Loaded reset spec: %+v", reset)
+	return reset, err
+}
+
+func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet, keyRemap map[string]string) (*v1.UpgradeSpec, error) {
```
mudler, line 239: :100: :heart:

cmd/install.go
```diff
@@ -50,50 +49,75 @@ func NewInstallCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
+			// Adapt 'docker-image' and 'directory'  deprecated flags to 'system' syntax
+			adaptDockerImageAndDirectoryFlagsToSystem(cmd.Flags())
 
-			if cfg.Target == "" {
-				return errors.New("at least a target device must be supplied")
+			//Adapt 'force-efi' and 'force-gpt' to 'firmware' and 'part-table'
+			adaptEFIAndGPTFlags(cmd.Flags())
+
+			// Map install sub viper keys to environment variables
+			// without the ELEMENTAL_INSTALL prefix
```
mudler: small nit: this doesn't feel right to belong here. maybe move the keyEnvMap into constants?

cmd/reset.go
```diff
@@ -45,31 +44,41 @@ func NewResetCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
+			if err := validateInstallUpgradeFlags(cfg.Logger, cmd.Flags()); err != nil {
 				return err
 			}
 
+			// Adapt 'docker-image' and 'directory'  deprecated flags to 'system' syntax
+			adaptDockerImageAndDirectoryFlagsToSystem(cmd.Flags())
+
+			// Map reset sub viper keys to environment variables
+			// without the ELEMENTAL_RESET prefix
+			keyEnvMap := map[string]string{
```
mudler: ditto

cmd/upgrade.go
```diff
@@ -53,33 +48,39 @@ func NewUpgradeCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
-			if cfg.DockerImg != "" || cfg.Directory != "" {
-				// Force channel upgrades to be false, because as its loaded from the config files,
-				// it will probably always be set to true due to it being the default value
-				cfg.ChannelUpgrades = false
+			// Adapt 'docker-image' and 'directory'  deprecated flags to 'system' syntax
+			adaptDockerImageAndDirectoryFlagsToSystem(cmd.Flags())
+
+			// Map upgrade sub viper keys to environment variables
+			// without the ELEMENTAL_UPGRADE prefix
+			keyEnvMap := map[string]string{
```
mudler: ditto, maybe a common keyEnvMap?

config.yaml.example
```diff
@@ -0,0 +1,130 @@
+# DISCLAIMER: this document show most of all available options. Elemental
```
mudler, line 1: This needs to be updated and reflected in https://rancher-sandbox.github.io/cOS-toolkit/docs/customizing/general_configuration/

cmd/install.go
davidcassany: yes I was thinking about it too, adding a default static method for each command is just better.

pkg/action/install.go
```diff
@@ -17,228 +17,149 @@ limitations under the License.
...
 	// Check no-format flag
-	if config.NoFormat {
+	if i.spec.NoFormat {
 		// Check force flag against current device
-		err = newElemental.CheckNoFormat()
-		if err != nil {
-			return err
+		labels := []string{i.spec.Active.Label, i.spec.Recovery.Label}
+		if e.CheckActiveDeployment(labels) && !i.spec.Force {
+			return fmt.Errorf("use `force` flag to run an installation over the current running deployment")
```
mudler, line 82: nit: I'm a bit confused here between the error message and the flag, now if the Noformat flag is given and if there is an active deployment we bail out? From the error message looks like we would partition something, as we have to specify `--force` to it

pkg/action/upgrade.go
```diff
@@ -28,362 +28,201 @@ import (
...
+	u.Info("Applying '%s' hook", hook)
 	if chroot {
 		mountPoints := map[string]string{}
 
-		oemDevice, err := utils.GetFullDeviceByLabel(config.Runner, config.OEMLabel, 2)
-		if err == nil && oemDevice.MountPoint != "" {
-			mountPoints[oemDevice.MountPoint] = "/oem"
+		oemDevice := u.spec.Partitions.OEM
+		if oemDevice != nil && oemDevice.MountPoint != "" {
+			mountPoints[oemDevice.MountPoint] = "/oem" //nolint:goconst
```
mudler: nit: Just thinking, maybe we should move this common default logics when we load the config data instead?

pkg/action/upgrade.go
```diff
@@ -28,362 +28,201 @@ import (
...
-	// create transition.img
-	img := v1.Image{
-		File:       transitionImg,
-		Size:       u.Config.ImgSize,
-		Label:      u.Config.ActiveLabel,
-		FS:         constants.LinuxImgFs,
-		MountPoint: upgradeTempDir,
-		Source:     upgradeSource, // if source is a dir it will copy from here, if it's a docker img it uses Config.DockerImg IN THAT ORDER!
+		mountPart = u.spec.Partitions.State
+		if mountPart == nil || mountPart.MountPoint == "" {
```
mudler: nit: similarly this could take place in an eventually Validate() step somewhere in the configs

pkg/action/upgrade.go
```diff
@@ -28,362 +28,201 @@ import (
...
-			// if mounted it may be RO only, so remount it RW?
-			err = u.Config.Mounter.Mount(statePartForRecovery.Path, statePartForRecovery.MountPoint, "auto", []string{"remount", "rw"})
-			if err != nil {
-				u.Error("Could not remount state partition with label %s for rebrand: %s", u.Config.StateLabel, err)
-				return err
-			}
+	// Only apply rebrand stage for system upgrades
+	if !u.spec.RecoveryUpgrade {
+		u.Info("rebranding")
+		osRelease, err := utils.LoadEnvFile(u.config.Config.Fs, filepath.Join(upgradeImg.MountPoint, "etc", "os-release"))
```
mudler, line 164: nit: shouldn't this read the env config file? aren't we re-reading it here, decoupling the paths defined already in https://github.com/rancher-sandbox/elemental/pull/189/files#diff-c6dc0c872cdd5407c8299f5b807a80a79ddf78e91e86c3c618508ca84c048d68R142 ?

pkg/action/upgrade.go
```diff
@@ -28,362 +28,201 @@ import (
...
-	osRelease, err := utils.LoadEnvFile(u.Config.Fs, filepath.Join(upgradeTempDir, "etc", "os-release"))
-	// override grub vars with the new system vars
-	u.Config.GrubDefEntry = osRelease["GRUB_ENTRY_NAME"]
-
-	err = ele.Rebrand()
 
-	if err != nil {
-		u.Error("Error running rebrand: %s", err)
-		return err
+		err = e.SetDefaultGrubEntry(mountPart.MountPoint, osRelease["GRUB_ENTRY_NAME"])
```
mudler, line 169: nit: similarly here, seems we are getting the same data from two different places

pkg/elemental/elemental.go
```diff
@@ -172,240 +161,255 @@ func (c Elemental) UnmountPartitions() error {
...
 // creates the filesystem image file, mounts it and unmounts it as needed.
-func (c *Elemental) DeployImage(img *v1.Image, leaveMounted bool) error {
+func (e *Elemental) DeployImage(img *v1.Image, leaveMounted bool) error {
 	var err error
 
+	target := img.MountPoint
 	if !img.Source.IsFile() {
-		//TODO add support for squashfs images
-		err = c.CreateFileSystemImage(img)
+		if img.FS != cnst.SquashFs {
```
mudler, line 272: nice!

cmd/config/config.go
davidcassany: I am not sure I understand your concern here. I am fine keeping some static configs around, however I'd follow the `/etc/elemental/config.yaml` approach, we could make it static if needed. Another options could be to load environment files as environment variables then they would be treated and read as such (which is no obvious IMHO).

To me it is weird keeping dual syntax logic, yaml and env files. Note that these env config files as they uesd to be apply over the same yaml syntax. So the environment file turns to something really weird, with var names such as `INSTALL.PARTITIONS.OEM.LABEL=COS_OEM` to give an example. We could use another delimiter instead of dots, but you see the essence 😓

mudler:
This looks good overall to me, it's also taking a really nice coverage on some parts and cleaning up some code, and adding some deliverables which weren't really part of the card... so kudos! I'd like to give it another look before approving as it is a big chunk of changes, but this is good!

config.yaml.example
davidcassany: Sure, docs will come next.

pkg/action/install.go
davidcassany: > now if the Noformat flag is given and if there is an active deployment we bail out?

Yes, now and before too. Basically it tries to use the already existing partitions, however if they are in use it errors out. With the `--force` command you can enforce installation even if the partitions are in use, meaning it will create active, passive and recovery images regardless the fact they could be already mounted. With the  `--no-format` flag no re-partitioning and formatting is done, it just dumps the images to the partitions.

pkg/action/upgrade.go
davidcassany: yes we can add constants for `/oem` and `/usr/local`, i was just uncertain about how to name them 😓 what about something like `constants.OEMPath` and  `constants.UsrLocalPath`? I want a name unrelated to this specific chroot use case.

pkg/action/upgrade.go
davidcassany: Well I configs there is already a similar check, config fails if no state partition or recovery partitions are found. However in some tests I realized not I was running the command without a proper configuration and resulting into hard to debug issues, so that I explicitly added simple sanity checks in all three commands at the very start. Probably they could be wrapped in a validate method and call it after unmarshalling as part of the `cmd/config`. For tests we could manually call it before running.

In fact the real thing I am missing is something like https://github.com/mitchellh/mapstructure/issues/253 

But yes this is something that can be done manually. We can create a validate method in each struct so we call it after unmarshaling.

pkg/action/upgrade.go
davidcassany: This feels odd to me too, this just aims to keep the same behavior as it had (probably I missed something). This is reading the `/etc/os-release` from the new upgraded image, the one we are deploying, so it is not the same file as it was read before. Moreover the `/etc/os-release` that is read in `cmd/config` is mapped against vipper, so only variables matching configuration fields are actually doing something. In fact I doubt reading the hosts `/etc/os-release` as in `cmd/config` is much valuable.

Here this is just a way to modify the default grub entry names via upgrades, which is slightly inconsistent with the installation procedure, where this is part of the configuration and not read from the installed image... It was already like this before, so I did not want to change it in this PR. But grub rebranding on install/upgrade should be aligned and consistent.

pkg/action/upgrade.go
davidcassany: they are different files, host vs deployed image. same as the comment above

davidcassany:
@mudler  as first round digest of the review gonna move envkeymap to constants and create a validate command over each config struct, this way we can easily grow in sanitize checks in a centralized fashion. Gonna run two additional commits for that, one for each.

cmd/config/config.go
mudler: my only concern here is that by limiting to 1 file only, there is no other way to override it in case of a downstream derivative. Let's say we have os2 and another derivative depending on it. At that point there is no way to source from both files, so the downstream derivative needs to overwrite the same file.

Anyhow, this is not a big issue to me, as technically speaking those settings should be provided anyhow by who is writing a derivative. It feels just a bit off that we force only the os-release file, as that might be completely up to who is writing a derivative if they want to touch it or left intact from the base image.

pkg/action/upgrade.go
mudler: sounds good here :+1:

pkg/action/upgrade.go
mudler: yup I think we should be consistent here, in the sense that we should fetch the configs in the same way across the steps. In this case it is sourcing the configuration from the upgrade source because the grub entry is specified in the image we are going to upgrade to. So if we upgrade to version X, we update the grub entry of X (and not the current installed version) which is annotated in the image. 

Anyhow feel free to ignore, this seems something for a follow-up, but keep in mind is used from the harvester team

mudler:
cc @kkaempf this changeset is still not compatible with os2, so after merging this PR new elemental builds won't work nicely with os2 until we sync them up

pkg/action/upgrade.go
davidcassany: Yes that's why I also didn't want to touch it. IMHO we should merge both behaviors, allow explicit string in config yaml and allow the value from the deployed `/etc/os-release` and agree on the prio. Then we can easily apply this criteria in both upgrade/install and it should be backward compatible.

davidcassany:
> cc @kkaempf this changeset is still not compatible with os2, so after merging this PR new elemental builds won't work nicely with os2 until we sync them up

Indeed, I'll check what needs to be changed in os2. I believe env variables and probably some small change in elemental command line if `--partitioning-layout` flag is in use.

mudler:
> > cc @kkaempf this changeset is still not compatible with os2, so after merging this PR new elemental builds won't work nicely with os2 until we sync them up
> 
> Indeed, I'll check what needs to be changed in os2. I believe env variables and probably some small change in elemental command line if `--partitioning-layout` flag is in use.

yup, maybe also https://github.com/rancher-sandbox/os2/blob/master/framework/files/etc/cos/config needs to be merged into os-release as well

mudler (approved):
Looks good here

Itxaka (commented):
still going trough it!

davidcassany:
> yup, maybe also https://github.com/rancher-sandbox/os2/blob/master/framework/files/etc/cos/config needs to be merged into os-release as well

just the grub config, however this should not prevent anything to boot, just change the grub naming. This could be part of /etc/os-release and if we make elemental install to consider it then we are done and it would cover install and upgrade processes.

davidcassany:
rebased and updated install, upgrade and reset command docs. In addition I also added the  `--squash-compression` flag to install command.

Itxaka (approved):
Looking good to me, will probably come back with questions once it's merged and I try to use  the flag/config lol

davidcassany:
merging since rancher-sandbox/cOS-Toolkit#1307 tests using this code are passing

PR #193, opened 2022-05-10
Try to get a decent luet tempdir

Itxaka:
Currently luet defaults to the default tmpdir which is usually /tmp

This is a problem for machines with low memory during upgrades as /tmp
is a tmpfs of 25% of the pgysical memory and can be exhausted easily.

Instead allow overriding the luet default tmpDirBase and reuse the
getTempDir util to try and get the persistent partition if its exists
and its mounted.

This should not affect other commands as it first tries to use the
TMPDIR if exists and then use the partition only if mounted, falling back
to the default /tmp dir.


pkg/utils/common.go
```diff
@@ -259,21 +261,26 @@ func LoadEnvFile(fs v1.FS, file string) (map[string]string, error) {
...
 }
 
-// GetUpgradeTempDir returns the dir for storing upgrade related temporal files
+// GetTempDir returns the dir for storing related temporal files
 // It will respect TMPDIR and use that if exists, fallback to try the persistent partition if its mounted
 // and finally the default /tmp/ dir
-func GetUpgradeTempDir(config *v1.RunConfig) string {
+func GetTempDir(config *v1.RunConfig) string {
 	// if we got a TMPDIR var, respect and use that
+	random.Seed(time.Now().UnixNano())
```
mudler: maybe https://github.com/rancher-sandbox/elemental/blob/main/pkg/utils/fs.go#L117 can be of any use here in this case?

mudler (approved):
Looks good!

pkg/utils/common.go
Itxaka: yes, was looking at that part and trying to rework that into this, like maybe make the fs return the tempdir somehow. But it needs to be linked with the runconfig for the labels so...a bit convoluted and dont want to mix both things

mudler (approved):

pkg/utils/common.go
```diff
@@ -259,21 +261,27 @@ func LoadEnvFile(fs v1.FS, file string) (map[string]string, error) {
...
 
-// GetUpgradeTempDir returns the dir for storing upgrade related temporal files
+// GetTempDir returns the dir for storing related temporal files
 // It will respect TMPDIR and use that if exists, fallback to try the persistent partition if its mounted
 // and finally the default /tmp/ dir
-func GetUpgradeTempDir(config *v1.RunConfig) string {
+// suffix is what is appended to the dir name elemental-suffix. If empty it will randomly generate a number
+func GetTempDir(config *v1.RunConfig, suffix string) string {
 	// if we got a TMPDIR var, respect and use that
+	if suffix == "" {
```
mudler, line 270: I just wonder, why not defaulting to random and avoid passing a suffix? Anyway, looks good to me

Itxaka: well, for using it on tests its necessary to prefill some stuff in order to trigger some tests.

Plus I think its nice in case we want to reuse it down the line, you can get a fixed tempdir but still get the proper place if needed, in case you want to reuse it somehow?

davidcassany (approved):

PR #195, opened 2022-05-10
Add job to auto create PR for updated docs

Itxaka:
Pretty sure we are gonna forget to update the docs if we ever change
anything in the cmds, so this job should auto create PRs with any
changes that happen from regenerating the docs after a merge into main
has occurred.


Itxaka:
test failure is due to the cache, has nothing to do with this change :=)

mudler (approved):

PR #196, opened 2022-05-10
Also set temp dir for luet pkgcache

Itxaka:
Seems like pkgs download are stored in there


mudler (approved):

mjura (approved):
LGTM

PR #197, opened 2022-05-10
Update go.mod to fix cache

Itxaka:
Seems like github cache is broken somehow, so we cannot run the test job

Updating the go.mod/go.sum should invalidate the cache


Itxaka:
Merging directly as this unblocks the test job

PR #203, opened 2022-05-17
Include other configs on config load

Itxaka:
Without it the embedded struct is ignored by the serialization libraries.

This fixes loading repos from the manifest file for example

Done by @davidcassany 


davidcassany (approved):

mjura (approved):
LGTM

PR #207, opened 2022-05-18
Make use of elemental API in build commands

davidcassany:

### DIFF
--- a/pkg/action/build-iso.go
+++ b/pkg/action/build-iso.go
@@ -18,182 +18,151 @@ package action
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/rancher-sandbox/elemental/pkg/constants"
-	"github.com/rancher-sandbox/elemental/pkg/partitioner"
+	"github.com/rancher-sandbox/elemental/pkg/elemental"
 	v1 "github.com/rancher-sandbox/elemental/pkg/types/v1"
 	"github.com/rancher-sandbox/elemental/pkg/utils"
 )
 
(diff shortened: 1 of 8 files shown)


Itxaka (approved):
nice!

pkg/action/build-iso.go
```diff
@@ -18,182 +18,151 @@ package action
...
+func NewBuildISOAction(cfg *v1.BuildConfig) *BuildISOAction {
+	return &BuildISOAction{
+		cfg: cfg,
+		e:   elemental.NewElemental(&cfg.Config),
+	}
+}
+
 // BuildISORun will install the system from a given configuration
-func BuildISORun(cfg *v1.BuildConfig) (err error) {
+func (b *BuildISOAction) ISORun() (err error) {
```
Itxaka, line 43: wait...what happened? I thought you didnt like this style and much preferred the old one of calling the method directly with the cfg?

davidcassany: indeed 🤣 good point!  yes you are right! Yes I have to admit it took me a while to get there 😅 

My reasoning is the following, I like the fact that in golang functions are first class citizens, so unless you need to operate and even modify some state full structures I do not see the need on creating objects that are not keeping any extra state. For instance I do not see the need of creating a struct that only includes a RunConfig struct, however this is no longer the case after #33 as RunConfig does not include the command specific state. Following with the #33 spirit my next PR is apply a similar criteria for BuildConfig, that is exclude the ISO struct and RawDisk struct from BuildConfig and make BuiIdSOAction to include the ISO and BuildConfig and the BuildDiskAction include RawDisk and BuildConfig.

PR #209, opened 2022-05-19
Add arch to config based on current arch

Itxaka:
No matter what, other commands can override this on demand by using the
WithArch option, so it should be safe to get the current arhc. Plus it
fix installing stuff under arm64


davidcassany (approved):
Looks good!

I wonder if we will need further translations such as `amd64` -> `x86_64`, but I guess we will discover it in trial and error fashion :)

PR #210, opened 2022-05-19
Use equivalent logic in run and build commands to load configuration

davidcassany:

### DIFF
--- a/cmd/build-disk.go
+++ b/cmd/build-disk.go
@@ -38,7 +38,6 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 		Short: "Build a raw recovery image",
 		Args:  cobra.NoArgs,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			_ = viper.BindPFlags(cmd.Flags())
 			if addCheckRoot {
 				return CheckRoot()
 			}
@@ -47,18 +46,13 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			mounter := &mountUtils.FakeMounter{}
 
-			// If configDir is empty try to get the manifest from current dir
-			configDir := viper.GetString("config-dir")
-			if configDir == "" {
-				configDir = "."
-			}
-
-			cfg, err := config.ReadConfigBuild(configDir, mounter)
+			flags := cmd.Flags()
+			cfg, err := config.ReadConfigBuild(viper.GetString("config-dir"), flags, mounter)
 			if err != nil {
 				return err
 			}
 
-			err = validateCosignFlags(cfg.Logger, cmd.Flags())
+			err = validateCosignFlags(cfg.Logger, flags)
 			if err != nil {
 				return err
 			}
@@ -66,22 +60,23 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			// Set this after parsing of the flags, so it fails on parsing and prints usage properly
 			cmd.SilenceUsage = true
 			cmd.SilenceErrors = true // Do not propagate errors down the line, we control them
-			imgType, _ := cmd.Flags().GetString("type")
-			output, _ := cmd.Flags().GetString("output")
-			oemLabel, _ := cmd.Flags().GetString("oem_label")
-			recoveryLabel, _ := cmd.Flags().GetString("recovery_label")
 
-			entry := cfg.RawDisk[cfg.Arch]
-			if entry == nil {
-				// We didnt load anything from the config file, create empty map
-				cfg.RawDisk = map[string]*v1.RawDiskArchEntry{
-					cfg.Arch: {Repositories: nil, Packages: nil},
-				}
+			spec, err := config.ReadBuildDisk(cfg, flags)
+			if err != nil {
+				cfg.Logger.Errorf("invalid install command setup %v", err)
+				return err
 			}
+
+			// TODO map these to buildconfig and rawdisk structs, so they
+			// are directly unmarshaled and there is no need handle them here
+			imgType, _ := flags.GetString("type")
+			output, _ := flags.GetString("output")
+			oemLabel, _ := flags.GetString("oem_label")
+			recoveryLabel, _ := flags.GetString("recovery_label")
 
 			// Set the repo depending on the arch we are building for
 			var repos []v1.Repository
-			for _, u := range cfg.RawDisk[cfg.Arch].Repositories {
+			for _, u := range (*spec)[cfg.Arch].Repositories {
 				repos = append(repos, v1.Repository{
 					URI:         u.URI,
 					Priority:    constants.LuetDefaultRepoPrio,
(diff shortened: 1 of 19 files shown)


davidcassany:
This PR applies the same principles in Build commands as they were applied for run time commands in #189

Related to #33

cmd/build-disk.go
```diff
@@ -47,74 +46,51 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
+			// are directly unmarshaled and there is no need handle them here
+			imgType, _ := flags.GetString("type")
+			output, _ := flags.GetString("output")
+			oemLabel, _ := flags.GetString("oem_label")
+			recoveryLabel, _ := flags.GetString("recovery_label")
+
 			// Set the repo depending on the arch we are building for
 			var repos []v1.Repository
-			for _, u := range cfg.RawDisk[cfg.Arch].Repositories {
+			for _, u := range (*spec)[cfg.Arch].Repositories {
```
Itxaka, line 79: this eneds to be updated if #212 gets merged

cmd/build-disk.go
```diff
@@ -47,74 +46,51 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
 				return fmt.Errorf("output file %s exists, refusing to continue", output)
 			}
 
-			// Set defaults if they are empty
-			if len(cfg.Config.Repos) == 0 {
-				repo := constants.LuetDefaultRepoURI
-				if cfg.Arch != "x86_64" {
-					repo = fmt.Sprintf("%s-%s", repo, cfg.Arch)
-				}
-				cfg.Logger.Infof("Repositories are empty, setting default value: %s", repo)
```
Itxaka, line 107: we miss this message which is really useful :D

cmd/config/config.go
```diff
@@ -259,6 +262,46 @@ func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet, keyRemap map[string]
...
+	return iso, err
+}
+
+func ReadBuildDisk(b *v1.BuildConfig, flags *pflag.FlagSet) (*v1.RawDisk, error) {
+	disk := config.NewRawDisk(b.Config)
+	vp := viper.Sub("raw_disk")
+	if vp == nil {
+		vp = viper.New()
+	}
+	// Bind build-iso cmd flags
```
Itxaka: buiild-disk

cmd/config/config.go
```diff
@@ -259,6 +262,46 @@ func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet, keyRemap map[string]
...
+	bindGivenFlags(vp, flags)
+	// Bind build-iso env vars
+	viperReadEnv(vp, "RAWDISK", constants.GetDiskKeyEnvMap())
```
Itxaka, line 294: is this even necessary? How is anyone gonna pass the complex config via env var? I would drop env vars from this command

davidcassany: Currently there are no env vars for this command `constants.GetDiskKeyEnvMap()` returns an empty map, so no environment variables are bound. I just set the placeholder in case we want to add some, also to make explicit that environment variables are defined in `constants`.

cmd/build-disk.go
davidcassany: Yes, we can add it in `config`, I am on it.

davidcassany: Oh no sorry, you are right, we lose this message with this change. We could add it in `config.NewBuildConfig`, but at this point the repo can be still overwritten by the manifest, so it could be misleading. Better avoid it...

davidcassany:
rebased to include litter addition

cmd/build-disk.go
davidcassany: Yes, lets merge #212 first, then I will rebase this.

cmd/build-disk.go
Itxaka: :(

cmd/build-disk.go
Itxaka: done!

davidcassany:
rebased again

cmd/build-disk.go
```diff
@@ -47,41 +46,37 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
-			entry := cfg.RawDisk[cfg.Arch]
-			if entry == nil {
-				// We didnt load anything from the config file, create empty map
-				cfg.RawDisk = map[string]*v1.RawDiskArchEntry{
-					cfg.Arch: {Repositories: nil, Packages: nil},
-				}
+
+			spec, err := config.ReadBuildDisk(cfg, flags)
+			if err != nil {
+				cfg.Logger.Errorf("invalid install command setup %v", err)
```
Itxaka, line 66: This is not clear to me. This is a config error, isnt it? also what is returned in the error that we need to print an `%v` ? Is not a string?

cmd/config/config.go
```diff
@@ -119,17 +118,20 @@ func ReadConfigBuild(configDir string, mounter mount.Interface) (*v1.BuildConfig
 	viper.SetConfigName("manifest.yaml")
 	// If a config file is found, read it in.
 	_ = viper.MergeInConfig()
-	viperReadEnv(viper.GetViper(), "", map[string]string{})
+
+	// Bind runconfig flags
+	bindGivenFlags(viper.GetViper(), flags)
+	// merge environment variables on top for rootCmd
+	viperReadEnv(viper.GetViper(), "BUILD", constants.GetBuildKeyEnvMap())
```
Itxaka, line 125: will this get the ELEMENTAL_BUILD or plain BUILD env vars?

cmd/config/config.go
```diff
@@ -260,6 +262,46 @@ func ReadUpgradeSpec(r *v1.RunConfig, flags *pflag.FlagSet, keyRemap map[string]
...
+	bindGivenFlags(vp, flags)
+	// Bind build-iso env vars
+	viperReadEnv(vp, "ISO", constants.GetISOKeyEnvMap())
+
+	err := vp.Unmarshal(iso, setDecoder, decodeHook)
+	if err != nil {
+		b.Logger.Warnf("error unmarshalling LiveISO: %s", err)
+	}
+	err = iso.Sanitize()
+	b.Logger.Debugf("Loaded LiveISO: %s", litter.Sdump(iso))
```
Itxaka, line 281: isnt this already printed via the ReadConfigBuild? maybe we are dumping too much? or is this called from a different place?

pkg/action/upgrade.go
```diff
@@ -92,10 +91,6 @@ func (u *UpgradeAction) Run() (err error) {
 		finalImageFile = filepath.Join(mountPart.MountPoint, "cOS", constants.ActiveImgFile)
 	}
 
-	if upgradeImg.Source.IsEmpty() {
-		return fmt.Errorf("undefined upgrade source")
```
Itxaka, line 96: this can never happen now?

Itxaka: oh I see, its been moved down there!

Itxaka (approved):
couple of questions but looks good

cmd/config/config.go
davidcassany: This will be `ELEMENTAL_BUILD_` for all the vars defined in `constants.GetBuildKeyEnvMap()`

cmd/config/config.go
davidcassany: not really see https://github.com/rancher-sandbox/elemental/blob/d93cf3f8f3a4f395787e4b3d5ba70a6cbc38ae93/pkg/types/v1/config.go#L366-L374

`BuildConfig` does no longer include the LiveISO struct, neither the RawDisk. Those are created only for the devoted command.

cmd/build-disk.go
davidcassany: well errors are no exactly strings, they are objects implementing the  `func Error() string` interface. So to get a string an `err.Error()` should be used. I thought of simply using this `%v` to let golang _guess_ a way to print it, just incase there is something lost in `Error()` implementation. To be honest I don't think it really makes any difference.

PR #214, opened 2022-05-20
Fix passing repositories to luet channel unpack

davidcassany:


Itxaka (approved):
confirmed working, nice!

PR #215, opened 2022-05-20
Introduce ReferenceId for Repository 

Itxaka:
Now that we are pushing different references, we can use those to
pin builds to a given reference

### DIFF
--- a/cmd/build-disk.go
+++ b/cmd/build-disk.go
@@ -82,7 +82,14 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			// Set the repo depending on the arch we are building for
 			var repos []v1.Repository
 			for _, u := range cfg.RawDisk[cfg.Arch].Repositories {
-				repos = append(repos, v1.Repository{URI: u.URI, Priority: constants.LuetDefaultRepoPrio})
+				repos = append(repos, v1.Repository{
+					URI:         u.URI,
+					Priority:    constants.LuetDefaultRepoPrio,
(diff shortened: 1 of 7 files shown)


davidcassany (approved):
Looks good to me, even I am not sure what reference IDs are for

Itxaka:
@davidcassany 

```
reference: (optional) A reference to a repository index file to use to retrieve the repository metadata instead of latest. This can be used to point to a different or an older repository index to act as a “wayback machine”. The client will consume the repository state from that snapshot instead of latest.
```

We are now pushing references to the repos instead of dumping every package in existence into the main repository.yaml

So for example we got: v0.8.10-1-repository.yaml which corresponds to the packages build and released with the v0.8.10-1 release, but we also got for normal master runs in the format GIT-SHA-repository.yaml. They are much smaller and self-contained, and they have "fixed" versions as to say, packages in that repository.yaml will never get upgraded with the latest package, they are pin to those versions. i.e. on v0.8.10-1 metadata only the version 0.10.1-6 for the yip package exists. As long as your reference is pointing to that repository file, you will never get an updated package from that repo. While using the normal repository file as soon as a package is pushed, its a candidate for upgrade as it only hosts the latest versions available.

This also kind of fixes the huge memory consumption when reading the repository files. The normal repo will only have latest versions, so its a much smaller file (4/5 Mb vs 300Mb).

And lastly, I heard that harvester team is hacking their way around to build isos with references, to have pinned packages, so this PR helps in providing that out of the box.

davidcassany:
Oh interesting :+1:

PR #216, opened 2022-05-20
Report rsync status

### LINKED ISSUE #194: Output of spawned commands not visible
Running `elemental build-iso` I see in `ps` output that `rsync` got spawned:

```
13324 ?        Ssl   54:17  \_ elemental --debug --config-dir . build-iso
13362 ?        S      0:13      \_ rsync --archive --acls --xattrs --partial --human-readable --progress --exclude=/mnt --exclude=/proc --exclude=/sys --exclude=/dev --exclude=/tmp --exclude=/host --exclude=/run /iso/overlay/ /tmp/elemental-iso619470678/rootfs/
13363 ?        S      0:02          \_ rsync --archive --acls --xattrs --partial --human-readable --progress --exclude=/mnt --exclude=/proc --exclude=/sys --exclude=/dev --exclude=/tmp --exclude=/host --exclude=/run /iso/overlay/ /tmp/elemental-iso619470678/rootfs/
13364 ?        S      0:21              \_ rsync --archive --acls --xattrs --partial --human-readable --progress --exclude=/mnt --exclude=/proc --exclude=/sys --exclude=/dev --exclude=/tmp --exclude=/host --exclude=/run /iso/overlay/ /tmp/elemental-iso619470678/rootfs/
```

rsync is running with `--progress` but there's no visible output although elemental is called with `--debug`

```
Step 15/26 : RUN cd /iso &&     elemental --debug --config-dir . build-iso
 ---> Running in 61b1b8ca4024
INFO[2022-05-10T09:23:22Z] Starting elemental version 0.0.14            
DEBU[2022-05-10T09:23:22Z] Full config loaded: &{ISO:0x400029f0e0 Date:false Name:output RawDisk:map[] OutDir: Config:{Logger:0x400047f7a0 Fs:0x1b21160 Mounter:0x400012f5d8 Runner:0x4000570450 Syscall:0x1b21738 CloudInitRunner:0x400028f650 Luet:0x400013a9b0 Client:0x4000136f70 Cosign:false CosignPubKey: Repos:[] Arch:x86_64 SquashFsCompressionConfig:[]}} 
INFO[2022-05-10T09:23:22Z] Preparing squashfs root...                   
DEBU[2022-05-10T09:23:22Z] Applying source /iso/overlay to target /tmp/elemental-iso619470678/rootfs 
```

Itxaka:
Fixes: https://github.com/rancher-sandbox/elemental/issues/194

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -136,7 +136,7 @@ func CreateDirStructure(fs v1.FS, target string) error {
 
 // SyncData rsync's source folder contents to a target folder content,
 // both are expected to exist before hand.
-func SyncData(fs v1.FS, source string, target string, excludes ...string) error {
+func SyncData(log v1.Logger, fs v1.FS, source string, target string, excludes ...string) error {
 	if fs != nil {
 		if s, err := fs.RawPath(source); err == nil {
 			source = s
@@ -165,6 +165,27 @@ func SyncData(fs v1.FS, source string, target string, excludes ...string) error 
 			Exclude: excludes,
 		},
 	)
+
+	if v1.IsDebugLevel(log) {
+		go func() {
+			for {
+				state := task.State()
+				// Only log status if there is still things to copy, otherwise we are finished
+				if state.Remain > 0 {
+					log.Debugf(
+						"progress rsync %s to %s: %.2f / rem. %d / tot. %d / sp. %s",
+						source,
+						target,
+						state.Progress,
+						state.Remain,
+						state.Total,
+						state.Speed,
+					)
+					<-time.After(5 * time.Second)
+				}
+			}
+		}()
+	}
 
 	err := task.Run()
 	if err != nil {
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,9 @@ module github.com/rancher-sandbox/elemental
 
 go 1.16
 
+// until https://github.com/zloylos/grsync/pull/20 is merged we need to use our fork
+replace github.com/zloylos/grsync v1.6.1 => github.com/rancher-sandbox/grsync v1.6.2-0.20220526080038-4032e9b0e97c
+
 require (
 	github.com/Masterminds/sprig/v3 v3.2.2 // indirect
 	github.com/ProtonMail/go-crypto v0.0.0-20220113124808-70ae35bab23f // indirect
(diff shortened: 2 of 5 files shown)


pkg/utils/common.go
```diff
@@ -166,6 +167,27 @@ func SyncData(fs v1.FS, source string, target string, excludes ...string) error
 		},
 	)
 
+	if log.GetLevel() == logrus.DebugLevel {
```
davidcassany: I think we can use the `IsDebugLevel` from `pkg/types/v1/logger.go` here, this way we don't need to import the `github.com/sirupsen/logrus` and get rid of the linter complaint

davidcassany (approved):
Cool! Looks good to me, just this little complaint of the linter

pkg/utils/common.go
Itxaka: How did I not know about that method 😱

Itxaka:
yeah, this keeps hitting the DATA RACE condition on tests, so Im not sure its a good idea to add it :/

davidcassany:
> yeah, this keeps hitting the DATA RACE condition on tests, so Im not sure its a good idea to add it :/

yes... 😞 I don't know what happened with the grsync maintainer 🤷🏽‍♂️ My first patch was merged and released in two days, but he follow up is still there waiting since months ... https://github.com/zloylos/grsync/pull/20 At some point can can consider forking if we get stuck, it is not such a very big tool.

pkg/utils/common.go
Itxaka: blargh, we cant becuase the interface does not implement that, in order to be compliant with the logrus logger ¬_¬

mjura: nice to remember

mjura (approved):
LGTM, I also reviewed https://github.com/zloylos/grsync/pull/20

pkg/utils/common.go
davidcassany: I think this should work

```golang
if v1.IsDebugLevel(log) {
...
}
```

Itxaka: Goddammit, I happen to have forgotten how to write golang apparently

pkg/utils/common.go
```diff
@@ -166,6 +167,27 @@ func SyncData(fs v1.FS, source string, target string, excludes ...string) error
...
+						state.Total,
+						state.Speed,
+					)
+					<-time.After(5 * time.Second)
+				}
+			}
+		}()
+	}
+
 	err := task.Run()
```
davidcassany, line 190: ~~Shouldn't we use `task.Start()` and `task.Wait()` instead of `task.Run()`. `task.Run()` is just kept to keep backward compatibility, but it should not be used. It is not thread safe this is what https://github.com/zloylos/grsync/pull/20 aims to fix. So I think the process should be~~

```
task.Start()
//process state until it is done
task.Wait()
```

~So I think there is no need for a goroutine~

Forget all this, I confused `Task.Run` with `Rsync.Run` 😞

PR #217, opened 2022-05-23
Adding support for additional environment variables to support ros-imstaller

davidcassany:
Related rancher-sandbox/cos-toolkit#1347


Itxaka (approved):
I dont think we need the reset ones yet right? Good to have them anyway.

davidcassany:
> I dont think we need the reset ones yet right? Good to have them anyway.

Right, I just added them for parity with the install command. Since they share most of the logic and config options I though it was fair to keep them in sync on that too. But yes, we do not need them.

PR #221, opened 2022-05-24
Deactivate unmounted lvm volumes before partitioning

### LINKED ISSUE #220: [installer] If LVM are active on the target disk, install will fail
Reported on https://github.com/harvester/harvester/issues/2230

Basically if LVs are active on the target install, elemental will fail. Once the LVs are tear down, install will complete as normal.

We should check for LVM in the target disk and tear it down if we found any active LVs

davidcassany:
Bug reported in https://github.com/harvester/harvester/issues/2230
Fixes #220 

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -82,6 +82,11 @@ func (i InstallAction) Run() (err error) {
 			return fmt.Errorf("use `force` flag to run an installation over the current running deployment")
 		}
 	} else {
+		// Deactivate any active volume on target
+		err = e.DeactivateDevices()
+		if err != nil {
+			return err
+		}
 		// Partition device
 		err = e.PartitionAndFormatDevice(i.spec)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


davidcassany:
I tried to figure out if there was a sane way of narrowing down the scope of devices deactivated and I could not manage to find reasonable way. `blkdeactivate` command without a target device argument attempts to all unmounted devices within the host, since it deactivates only unmounted devices I consider it acceptable, it is unlikely that it causes important issues if it deactivates some unrelated devices as it was not mounted.

I tried `blkdeactivate /dev/sda` hoping it could eventually deactivate any LVM volumes including `/dev/sda`, but this is not the case, it requires the logical volume as argument. Then checking if a physical block devices is used to conform LVM volumes is not that simple, the only way I found to discover that with a structured output is using `lsblk`...

Itxaka:
udev reports that stuff, I wonder if ghw reports lvm stuff...

Itxaka (approved):

mudler (approved):
Looks good here!

Issue #223, opened 2022-05-25
Can't specify default grub entry


mudler:
According to https://github.com/rancher-sandbox/elemental/blob/main/config.yaml.example and by following https://rancher-sandbox.github.io/cOS-toolkit/docs/customizing/general_configuration/  I've tried to run an install with the following elemental config file:
```yaml
cosign: false
verify: false
install:
  grub-entry-name: "c3os"
  system:
    size: 2000
  recovery-system:
    size: 2000
upgrade:
  grub-entry-name: "c3os"
  recovery-system:
    size: 2000
  system:
    size: 2000
reset:
  grub-entry-name: "c3os"
  system:
    size: 2000
```
After install, the grub entry name wasn't applied, and indeed I could confirm that the grub env populated is with the default values:

```bash
cat /run/initramfs/cos-state/grub_oem_env 
# GRUB Environment Block
# WARNING: Do not edit this file by tools other than grub2-editenv!!!
default_menu_entry=cOS
```

By running upgrade with `--debug` can be seen that no string was loaded in the config.

Reproducible @ https://github.com/c3os-io/c3os/tree/updates

mudler:
mmm I'm trying to do few experiments here with a unit test... but looks also `cosign` and `verify` aren't read at all? I'm a bit confused. Although the size of images seems to be read correctly

mudler:
Fixed in https://github.com/rancher-sandbox/elemental/pull/224

PR #225, opened 2022-05-26
Move fixtures to its own folder, enhance config tests

mudler:
The added test is introduced while trying to debug #223


cmd/config/config_test.go
```diff
@@ -49,6 +46,30 @@ var _ = Describe("Config", Label("config"), func() {
 	AfterEach(func() {
 		viper.Reset()
 	})
+
+	Context("From fixtures", func() {
+		Describe("read all specs", Label("install"), func() {
+			It("reads values correctly", func() {
+				cfg, err := ReadConfigRun("../../tests/fixtures/simple/", nil, mounter)
```
Itxaka, line 53: blarg, isnt there a way of getting the root of the project somehow? this is really ugly :/

Itxaka (approved):

cmd/config/config_test.go
mudler: why? This is completely fine with go, as paths are resolved from where the unit test is

Itxaka: `¯\_(ツ)_/¯` I just dont really like this relative path vs full paths

mudler: blargh I've tried to have a look but the only way would be to use `runtime.Caller`, but things then look even more uglier..

PR #229, opened 2022-05-30
Ensure recovery image file is set according to the filesystem

davidcassany:


mudler (approved):
Looking good!

PR #230, opened 2022-05-30
Relabel SELinux in a chroot env

davidcassany:

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -403,9 +403,9 @@ func (e *Elemental) SelinuxRelabel(rootDir string, raiseError bool) error {
 		var out []byte
 		var err error
 		if rootDir == "/" || rootDir == "" {
-			out, err = e.config.Runner.Run("setfiles", "-F", "-v", contextFile, rootDir)
+			out, err = e.config.Runner.Run("setfiles", "-F", contextFile, rootDir)
 		} else {
-			out, err = e.config.Runner.Run("setfiles", "-F", "-v", "-r", rootDir, contextFile, rootDir)
+			out, err = e.config.Runner.Run("setfiles", "-F", "-r", rootDir, contextFile, rootDir)
 		}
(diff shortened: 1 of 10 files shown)


pkg/elemental/elemental.go
```diff
@@ -396,23 +396,24 @@ func (e *Elemental) CopyCloudConfig(cloudInit string) (err error) {
...
-
 	contextFile := filepath.Join(rootDir, "/etc/selinux/targeted/contexts/files/file_contexts")
 
-	_, err = e.config.Fs.Stat(contextFile)
-	contextExists := err == nil
+	contextExists, _ := utils.Exists(e.config.Fs, contextFile)
 
-	if utils.CommandExists("setfiles") && contextExists {
-		_, err = e.config.Runner.Run("setfiles", "-r", rootDir, contextFile, rootDir)
+	if contextExists && utils.CommandExists("setfiles") {
```
mudler, line 402: mmm I do wonder, in this case if setfile is present in the rootfs we just relabel regardless, right? The issue I see with this approach is if that the rootfs ships the file but does not intend to have any SELinux capabilities, we might just fail hard

I'm thinking that it might have more sense to be explicit. that is we enable selinux support explicitly via config file, and at that point we treat this without any specific logic around the error. If we want selinux and we fail, than we just error out. that should simplify also the error logics around here

davidcassany: This is will only be executed if there is the selinux policy file installed (the context file) and setfiles utility present. I think if the too are there we are should be pretty confident that the selinux packages/capabilities are there. Anyway I agree that having an opt in configuration flag for selinux makes sense.

This PR is just to enforce relaying on the `setfiles` utility installed within the upgraded/installed image, I found this need while using `elemental install` over a loop device on my host. I do not have SELinux in my host, but I still should be capable to build a disk having SELinux support.

mudler (approved):
Looks good here!

PR #235, opened 2022-06-03
rawdisk: Consume repositories from top level YAML

mudler:


cmd/build-disk.go
```diff
@@ -76,13 +76,23 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 
 			// Set the repo depending on the arch we are building for
 			var repos []v1.Repository
-			for _, u := range (*spec)[cfg.Arch].Repositories {
+			for _, u := range cfg.Repos {
```
davidcassany, line 79: I guess this logic is to filter repositories, wouldn't it be cleaner to run this logic inside the Luet interface implemenetation? I am thinking across this lines:

https://github.com/rancher/elemental-cli/blob/1554b61fd1642611bb7517196b873156377171b5/pkg/luet/luet.go#L172-L214

the `initLuetRepository` method could skip repo initialization if the repository arch an the `v1.Config.Arch` do not match. This way we apply the criteria for all Luet operations.

pkg/types/v1/config.go
```diff
@@ -387,6 +387,8 @@ type BuildConfig struct {
 
 type RawDisk map[string]*RawDiskArchEntry
 
+type Repositories []Repository
```
davidcassany, line 390: We could use that in `v1.Config.Repos`, not that is really needed, but we leave it prepared just in case we want to add some extra logic/methods under `Repositories`.

davidcassany (commented):
Looks good, even I think we could target a more generic approach at Luet level.

pkg/types/v1/config.go
mudler: argh, I think it was a leftover as we later realized it was already there. it was meant to be used in `v1.Config.Repos`, it's not actually used at all.

cmd/build-disk.go
```diff
@@ -76,13 +76,23 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
...
+				golangArch, err := utils.ArchToGolangArch(cfg.Arch)
+				if err != nil {
+					cfg.Logger.Errorf("invalid arch provided for repository '%s': %s", u.Name, err.Error())
+					return err
+				}
+
+				if u.Arch != "" && u.Arch != golangArch {
```
Itxaka, line 86: should probably warn or debug here that we are ignoring the repo, for easy troubleshoot.

pkg/utils/common.go
```diff
@@ -486,3 +486,23 @@ func FindFileWithPrefix(fs v1.FS, path string, prefixes ...string) (string, erro
 	}
 	return "", fmt.Errorf("No file found with prefixes: %v", prefixes)
 }
+
+var errInvalidArch = fmt.Errorf("invalid arch")
+
+func ArchToGolangArch(arch string) (string, error) {
```
Itxaka, line 492: this could be used in the pkg/config/config.go file as well.

pkg/types/v1/config.go
frelon: Removed the type :+1:

cmd/build-disk.go
frelon: Added debug log!

pkg/utils/common.go
frelon: Nice catch, replaced some code in config.go as well!

pkg/constants/constants.go
```diff
@@ -120,6 +120,9 @@ const (
 
 	// Eject script
 	EjectScript = "#!/bin/sh\n/usr/bin/eject -rmF"
+
+	ArchAmd64 = "amd64"
```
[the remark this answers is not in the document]
frelon: Added arm64 as well! Do we have a list of supported architectures?

cmd/build-disk.go
frelon: Exactly it is used for filtering. By skipping initialization you mean returning an empty LuetRepository? Or an error we can check where this is used to skip the repo?

pkg/constants/constants.go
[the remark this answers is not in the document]
Itxaka: On the CMD part there is an option that restricts the arch choices. Currently only x86 and arm64!

pkg/config/config.go
```diff
@@ -98,13 +99,11 @@ func WithArch(arch string) func(r *v1.Config) error {
...
-	arch := func() string {
-		switch runtime.GOARCH {
-		case "amd64":
-			return "x86_64"
-		}
-		return runtime.GOARCH
-	}()
+	arch, err := utils.GolangArchToArch(runtime.GOARCH)
+	if err != nil {
+		return nil
```
Itxaka, line 104: should return error here?

cmd/build-disk.go
Itxaka: I think he means moving this into the generic `Luet.initLuetRepository` so we blindly add all repos here, we dont care about filtering them or anything and then when `Luet` inits the repos it will either ignore or disable repos in which the `repo.Arch` does not match the `Config.arch`. That way we can load the repos anywhere and Luet will autofilter based on that.

frelon: Looks like we can set Arch on the LuetRepository and luet will take care of the filtering instead :+1:

pkg/config/config.go
frelon: Hmm, it only returns a `*v1.Config` but I added an error log to not swallow the error entirely!

pkg/luet/luet.go
```diff
@@ -203,13 +204,19 @@ func (l Luet) initLuetRepository(repo v1.Repository) (luetTypes.LuetRepository,
...
+	}
+
 	return luetTypes.LuetRepository{
 		Name:        name,
 		Priority:    repo.Priority,
 		Enable:      true,
 		Urls:        []string{repo.URI},
 		Type:        repoType,
 		ReferenceID: repo.ReferenceID,
+		Arch:        arch,
```
Itxaka, line 219: I think this might bite us. AFAIK luet will use this and the runtime.GOARCH to check if a repo is enabled, so we may have issues cross-building. Let me have a check locally!

Itxaka:
umm, this doesnt work. 

running with `elemental --debug --config-dir . build-disk --arch arm64` shows that it picks up both repos, but it doesnt disable any of them, and the artifacts are picked from a random one, no matter what are we building for. It also shows that even with only the arm64 repo in there, running from a x86 arch, it doesnt filter the repositories so we probably not calling that `SystemRepositories` on luet which is the one that filters them out.

We probably need to add the filtering on our side, on our luet interface instead of relaying on the luet package to do it for ourselves, so we can filter according to our cfg.Arch

Seems like this is getting more complicated :D

Itxaka:
> for the luet runtime repo detection arch needs to be specified and enabled needs to be set to false. The Enable field takes precedence and enable the repos nevertheless.

Ok, still better, because I think we better do the filtering in our side to allow for cross-arch build

frelon:
Oops that password is no longer usable...

PR #239, opened 2022-06-08
Split OEM reset from Persistent reset

### LINKED ISSUE #238: Resetting OEM should have its own separate flag
https://github.com/rancher/elemental-cli/blob/04bb568dae24dc9b8799f8137193406aea2b2966/pkg/action/reset.go#L83

OEM shouldn't be considered as the persistent application data, but rather a configuration overlay. If we wipe that out, eventual oem data that are written during install could get lost.

We should have a separate argument to optionally wipe OEM data.

mudler:
Fixes: https://github.com/rancher/elemental-cli/issues/238


Itxaka:
crap, was also doing this....at least mine has tests for this :P

Itxaka (approved):

mudler:
care to push them here or in a follow-up...? :)

Itxaka:
follow up, no probs, Im also working in more testing for the cli so I will push it together

PR #241, opened 2022-06-08
increase reset test

Itxaka:


mudler (approved):

mjura (approved):
New test are always welcome, thank you

pkg/utils/runstage_test.go
```diff
@@ -40,7 +40,7 @@ func writeCmdline(s string, fs v1.FS) error {
 	return fs.WriteFile("/proc/cmdline", []byte(s), os.ModePerm)
 }
 
-var _ = Describe("run stage", Label("RunStage", "root"), func() {
+var _ = Describe("run stage", Label("RunStage"), func() {
```
davidcassany, line 43: cool, another test that does not require root 👍🏽

davidcassany (approved):
Looks good :+1:

pkg/action/build-disk.go
```diff
@@ -220,7 +220,10 @@ func Raw2Gce(source string, fs v1.FS, logger v1.Logger, keepOldImage bool) error
 	}
 	defer file.Close()
 	// Create gzip writer
-	gzipWriter := gzip.NewWriter(file)
+	gzipWriter, err := gzip.NewWriterLevel(file, gzip.BestSpeed)
```
Itxaka, line 223: @davidcassany @mudler @mjura @fgiudici @frelon this introduces a small change in the code.

While testing this I noticed it was a bit slow compared to other tests and got to check, most of the time was spent on compressing the final file due to the default compression from gzip being 9, i.e. highest compression, which is very time consuming. Testing locally with different values I found out that for our use case there is almost no difference betwee 1 (BestSpeed) and 9 (BestCompression) as it only saves us about 1 or 2Mb, but it drops the creation speed by a lot (from 14 seconds to 3 seconds in a beefy machine, from 400 seconds to 120 in a more restricted machine like in CI)

I think its worth

PR #245, opened 2022-06-09
Improve SELinux labeling commands

davidcassany:
Ensure setfiles utility makes use of the guest selinux policy. Setfiles
by default makes use of the policy currenlty loaded within the host, if
any. In order to workaround that provide the specific binary policy file
to be loaded instead.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -100,7 +100,11 @@ const (
 	BuildImgName           = "elemental"
 	UsrLocalPath           = "/usr/local"
 	OEMPath                = "/oem"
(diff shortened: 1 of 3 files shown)


davidcassany:
Yesterday eve by causality while checking KIWI updates I found this issue https://github.com/livecd-tools/livecd-tools/pull/236, this is to prevent inconsistencies while labeling files if the host already has SELinux enable and using an incompatible policy compared to the installed one. Note in our case this could even happen during upgrades if the image to upgrade includes changes to the SELinux policy.

Itxaka (approved):

PR #249, opened 2022-06-09
Fix default config for RawDisk

### LINKED ISSUE #248: build-disk panics if no config provided
```
$ sudo bin/elemental build-disk --debug                            
DEBU[2022-06-09T14:22:43+02:00] Starting elemental version v0.0.15 on commit 73aa161d1a5d82a2117e150030d60f1a16830824 
DEBU[2022-06-09T14:22:43+02:00] Full config loaded: &v1.BuildConfig{
  Date: false,
  Name: "elemental",
  OutDir: "",
  Config: v1.Config{
    Logger: &v1.logrusWrapper{ // p0
      Logger: &logrus.Logger{
        Out: &os.File{},
        Hooks: logrus.LevelHooks{},
        Formatter: &logrus.TextFormatter{
          ForceColors: true,
          DisableColors: false,
          ForceQuote: false,
          DisableQuote: false,
          EnvironmentOverrideColors: false,
          DisableTimestamp: false,
          FullTimestamp: true,
          TimestampFormat: "",
          DisableSorting: false,
          SortingFunc: ,
          DisableLevelTruncation: false,
          PadLevelText: false,
          QuoteEmptyFields: false,
          FieldMap: logrus.FieldMap(nil),
          CallerPrettyfier: ,
        },
        ReportCaller: false,
        Level: 5,
        ExitFunc: os.Exit,
      },
    },
    Fs: &vfs.osfs{}, // p1
    Mounter: &mount.FakeMounter{
      MountPoints: nil,
      MountCheckErrors: map[string]error(nil),
      UnmountFunc: ,
    },
    Runner: &v1.RealRunner{ // p2
      Logger: p0,
    },
    Syscall: &v1.RealSyscall{},
    CloudInitRunner: &cloudinit.YipCloudInitRunner{},
    Luet: &luet.Luet{
      VerifyImageUnpack: false,
      TmpDir: "",
    },
    Client: &http.Client{},
    Cosign: false,
    Verify: false,
    CosignPubKey: "",
    LocalImage: false,
    Repos: []v1.Repository{
      v1.Repository{
        Name: "cos",
        Priority: 90,
        URI: "quay.io/costoolkit/releases-green",
        Type: "docker",
        Arch: "",
        ReferenceID: "",
      },
    },
    Arch: "x86_64",
    SquashFsCompressionConfig: []string{
      "-comp",
      "xz",
      "-Xbcj",
      "x86",
    },
  },
} 
WARN[2022-06-09T14:22:43+02:00] error unmarshalling RawDisk: 1 error(s) decoding:

* '[debug]' expected a map, got 'bool' 
DEBU[2022-06-09T14:22:43+02:00] Loaded RawDisk: &v1.RawDisk{}                
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x11a7e4e]

goroutine 1 [running]:
github.com/rancher/elemental-cli/cmd.NewBuildDisk.func2(0xc00041bb80, {0xc000287550?, 0x1?, 0x1?})
	/home/itxaka/projects/elemental-cli/cmd/build-disk.go:79 +0x1ce
github.com/spf13/cobra.(*Command).execute(0xc00041bb80, {0xc000287540, 0x1, 0x1})
	/home/itxaka/go/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0xc00041b900)
	/home/itxaka/go/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:974 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
	/home/itxaka/go/pkg/mod/github.com/spf13/cobra@v1.3.0/command.go:902
github.com/rancher/elemental-cli/cmd.Execute()
	/home/itxaka/projects/elemental-cli/cmd/root.go:48 +0x25
main.main()
	/home/itxaka/projects/elemental-cli/main.go:21 +0x17

```

Itxaka:
Having the config.ZeroFields set to true works for other places but in
the case of build-disk, we have some default in an array of arches and
setting it to true makes the whole thing to dissapear. But if there is
no configuration provided that can supply those values again, we end up
with an empty RaDisk config which is good for nothing.

This removes setting those ZeroFields to true, so the defaults are
always in the RawDisk and if there is a config coming, it will overwrite
it.

Tested locally with no config and providing a config.

As a side effect of the testing, this also adds logging for errors that
didnt have anything, so we could error out and not inform of the why.

Fixes #248 

### DIFF
--- a/cmd/build-disk.go
+++ b/cmd/build-disk.go
@@ -67,6 +67,14 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 				return err
 			}
 
+			// Get the spec for the arch only
+			var specArch *v1.RawDiskArchEntry
+			if cfg.Arch == "x86_64" {
+				specArch = spec.X86_64
+			} else {
+				specArch = spec.Arm64
+			}
+
 			// TODO map these to buildconfig and rawdisk structs, so they
 			// are directly unmarshaled and there is no need handle them here
 			imgType, _ := flags.GetString("type")
@@ -76,7 +84,7 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 
 			// Set the repo depending on the arch we are building for
 			var repos []v1.Repository
-			for _, u := range (*spec)[cfg.Arch].Repositories {
+			for _, u := range specArch.Repositories {
 				repos = append(repos, v1.Repository{
 					URI:         u.URI,
 					Priority:    constants.LuetDefaultRepoPrio,
@@ -97,7 +105,7 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 				return fmt.Errorf("output file %s exists, refusing to continue", output)
 			}
 
-			err = action.BuildDiskRun(cfg, spec, imgType, oemLabel, recoveryLabel, output)
+			err = action.BuildDiskRun(cfg, specArch, imgType, oemLabel, recoveryLabel, output)
 			if err != nil {
 				return err
 			}
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -283,7 +283,7 @@ func ReadBuildISO(b *v1.BuildConfig, flags *pflag.FlagSet) (*v1.LiveISO, error) 
 }
 
 func ReadBuildDisk(b *v1.BuildConfig, flags *pflag.FlagSet) (*v1.RawDisk, error) {
-	disk := config.NewRawDisk(b.Config)
+	disk := config.NewRawDisk()
 	vp := viper.Sub("raw_disk")
 	if vp == nil {
 		vp = viper.New()
--- a/cmd/config/config_test.go
+++ b/cmd/config/config_test.go
@@ -164,10 +164,10 @@ var _ = Describe("Config", Label("config"), func() {
 				Expect(err).ShouldNot(HaveOccurred())
 
 				// From config file
-				Expect(len((*disk)["x86_64"].Packages)).To(Equal(1))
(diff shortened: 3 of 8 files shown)


davidcassany:
I see how this fixes the issue, however I have doubts about it.

The problem is that the viper treats `RawDisk` as a single value to overwrite. For similar issues I ended up creating the `v1.ElementalPartitions` which was formerly a map, I could not find a way to properly merge the values within the map. In viper/mapstructure (IIRC is more mapstructure limitation) in any key value pair is used in config, at least the Key-Value pair is completely overwritten, this means the whole `RawDiskArchEntry` gets recreated and any default value there gets lost. Probably in this particular case this is not a big deal as those structs are really simple and repositories will be gone, but I think we are just lucky not to struggle with this limitation.

IMHO the problem is that viper/mapstructure are not really good to unmarshal to maps or top level maps, as they treat the map as a whole, as a unit. What if we consider using something like 

```golang
type RawDisk struct {
    X86_64 RawDiskArchEntry `yaml:"x86_64,omitempty" mapstructure:"x86_64"`
    Arm64  RawDiskArchEntry `yaml:"arm64,omitempty" mapstructure:"arm64"`
    ...
}
```

It has the limitation that you need to explicitly list supported entries, but on the other side it plays way better with viper and mapstructure and we are explicit. Also having in mind that repositories per arch list will be gone we could consider dropping `RawDiskArchEntry` struct and slightly simplify these structs.

Itxaka:
I was trying to avoid that but it seems more reasonable. Sighs...

frelon (approved):
Looks good to me!

PR #253, opened 2022-06-09
Bump ghw and fix labels

### LINKED ISSUE #254: ghw does not fetch partition labels
**cos-toolkit version:**

current master version

**CPU architecture, OS, and Version:**

any

**Describe the bug**

There is a regression with the ghw library adoption in elemental, the library does not get the partition label only the filesystem label. The partition name used to be mapped to partition label, now it gets mapped to some device filename. It is not big deal as long as we use default partition labels and as long as we do not relay on the given partition name form the ghw library (I am not 100% sure this is actually the case in current head). The partition names in elemental are constant and human readable (`oem` , `recovery` , `state`,  `efi` , etc.) and partitioner uses that as the partition label.

Not a bug causing issue now, but I bet this can be problematic at some point. The partition label mapped to partition name was done in purpose to provide a canonical way to identify partitions beyond filesystem labels. In code it is not really used because we still support legacy MSDOS partition tables where there are no partition labels at all.

**To Reproduce**

make use of 

**Expected behavior**

The method `utils.GetAllParitions` should map `v1.Partition.Name` to the device partition label if present.

**Additional context**

I found that in the scope of rancher-sandbox/elemental#33 while thinking on how to discover partitions.

Itxaka:
ghw was fixed upstream moving the proper partition label to
Label and having the filesystem label in the proper place,
FilesystemLabel, which is the one that we are using.

This patch bumps ghw and reworks our code to use that FS label instead

Fixes https://github.com/rancher/elemental-cli/issues/254


frelon (approved):
Nice!

pkg/utils/getpartitions.go
```diff
@@ -30,7 +30,7 @@ import (
 // ghwPartitionToInternalPartition transforms a block.Partition from ghw lib to our v1.Partition type
 func ghwPartitionToInternalPartition(partition *block.Partition) *v1.Partition {
 	return &v1.Partition{
-		Label:      partition.Label,
+		Label:      partition.FilesystemLabel,
 		Size:       uint(partition.SizeBytes / (1024 * 1024)), // Converts B to MB
 		Name:       partition.Name,
```
davidcassany: This is the partition label, this is what used to be the `PLabel` field back in time. So now this can be properly set to `Name: partition.PartitionLabel` (I don't remember exactly how it is finally called the partition label field in ghw).

Indeed this is the field that made me notice we were not using the partition label.

davidcassany (approved):
Looks good to me! Just a little enhancement to actually make use of the partition label from ghw.

pkg/utils/getpartitions.go
Itxaka: ah good point!

pkg/types/v1/config.go
```diff
@@ -182,14 +182,14 @@ func (u *UpgradeSpec) Sanitize() error {
...
-	Name       string
-	Label      string   `yaml:"label,omitempty" mapstructure:"label"`
-	Size       uint     `yaml:"size,omitempty" mapstructure:"size"`
-	FS         string   `yaml:"fs,omitempty" mapstrcuture:"fs"`
-	Flags      []string `yaml:"flags,omitempty" mapstrcuture:"flags"`
-	MountPoint string
-	Path       string
-	Disk       string
+	Name            string
+	FilesystemLabel string   `yaml:"FilesystemLabel,omitempty" mapstructure:"FilesystemLabel"`
```
davidcassany: I'd rather keep the yaml untouched, since this is user facing and this is the only configurable label. Also the filesystem label is referred to `LABEL` in most system utilities, so from a user perspective I think it simpler to call it `label`.

Itxaka: umm, not sure about that. What if down the line we need to use the partition label? Wouldnt it be simpler to just call things by its name as close as possible?

Not against changing it, but I would not want to change this again down the line, we did already too many changes tot he config stuff :)

davidcassany: I see your point, however to the user I'd use the naming convention as in fstab, device mapper, lsblk, blkid and others. They essentially use `LABEL` and `PARTLABEL`. With that in config we could use `label` and `part-label`.

I am also fine with something like `fs-label` (or `filesystem-label`) and `part-label`, no strong opinion here.

What I dislike is `FilesystemLabel` for the yaml, as it is not consistent with the other tags, we are not using capitals anywhere in configs and mostly using dashes as words separator.

Itxaka: umm, I see your point and I think that label is the correct one, especially if its used across other tools for fs/disks, it makes sense to have the same nomenclature in here. Reverting it back to label!

PR #255, opened 2022-06-10
rawdisk: Consume repositories from top level YAML

### LINKED ISSUE #1346: Use a single repositories list in manifest.yaml
Currently for `build-disk` command the repositories can be set in two different places under each arch specific stanza and globally.
I think this is error prone, confusing and adds some extra logic in elemental. I believe we should have only a single repositories list and eventually filter them by arch, so each architecture sees only the matching repositories at build time.

This is a currently valid

```yaml
raw_disk:
  x86_64:
    # which packages to install and the target to install them at
    packages:
      - name: system/myos
        target: efi
  aarch64:
    repositories:
     - uri: quay.io/costoolkit/releases-green-arm64
    packages:
      - name: system/grub2-efi-image
        target: efi

repositories:
- uri: registry.org/my/repo
```

I suggest to change it to:

```yaml
raw_disk:
  x86_64:
    # which packages to install and the target to install them at
    packages:
      - name: system/myos
        target: efi
  aarch64:
    packages:
      - name: system/grub2-efi-image
        target: efi

repositories:
- uri: registry.org/my/repo
  arch: "x86_84"
- uri: "quay.io/costoolkit/releases-green-arm64"
  arch: "aarch64"
```

frelon:
Had to create a new PR for this, took some fiddling around trying to get `cfg.Arch` into the `luet.UnpackFromChannel`, hopefully this is reusable enough.

Fixes https://github.com/rancher/elemental-toolkit/issues/1346


mudler (commented):
It looks good here! although I'd prefer @Itxaka to have a look as he already reviewed them first time :sweat_smile:

pkg/luet/luet.go
```diff
@@ -203,9 +206,14 @@ func (l Luet) initLuetRepository(repo v1.Repository) (luetTypes.LuetRepository,
 		repo.ReferenceID = "repository.yaml"
 	}
 
+	priority := repo.Priority
+	if repo.Priority == 0 {
```
Itxaka, line 217: this can potentially break stuff on the repos that we add for building the iso...I would need to check it further, althougth it seems the correct thing to do... :thinking:

Itxaka (approved):
lets gooo

Itxaka:
FYI if you use `Fixes` and then the issue link, it will auto link this PR to the issue and close the issue once this is merged ;)

Itxaka:
Doesnt seem to work to ignore the repos as its trying to build the iso on x86 with the packages from arm64: https://github.com/rancher/elemental-toolkit/pull/1473

Itxaka:
oh yeah, it makes sense. Like we updated the package in cos-toolkit to test, but the elemental being used from make_deps is the ones from the repos. So for building the ISO we are using the latest released/pushed artifact, not the one bumped in the package.

I guess we should do like in other places and add the build repo before installing the deps?

Itxaka:
mmmh, why is this not picking codecov?

frelon:
> oh yeah, it makes sense. Like we updated the package in cos-toolkit to test, but the elemental being used from make_deps is the ones from the repos. So for building the ISO we are using the latest released/pushed artifact, not the one bumped in the package.
> 
> I guess we should do like in other places and add the build repo before installing the deps?

Nice find! Could you point me to where this is done? Or maybe I could make the change backwards compatible to work with both versions of config...

Itxaka:
@frelon sorry, I went ahead and added it already! I was wondering why was this not working as the code made sense and seemed to work as expected :=)

frelon:
@Itxaka Hehe, no worries!

pkg/luet/luet.go
frelon: Looks like priority is already set in `build-iso` command so this #should not break anything :sweat_smile:

Itxaka: famous last words :P

cmd/build-iso.go
```diff
@@ -51,6 +52,8 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			path, err := exec.LookPath("mount")
 			if err != nil {
+				logger := v1.NewLogger()
```
Itxaka: The only problem i see with this is that we could be logging to an unwanted location. I.e. Maybe we have set up the `-quiet` flag because its being called from somewhere else? That is why there is no logger before the config is ingested, so we dont log unwanted stuff :)

frelon: Good catch! Added it for debugging purposes trying to get my PR in elemental-toolkit green, but I don't think that is the issue so will remove this again!

Itxaka (approved):
x86 passed, arm64 issue is due to unrelated stuff but it works locally, good to me :shipit:

PR #256, opened 2022-06-13
Fix build-iso command when no arguments are provided

davidcassany:
In previous changes of build-iso command we omitted the case where all
the configurations are comming from the manifest and no rootfs image
is provided within the command line.


mudler (approved):
Looks good!

PR #257, opened 2022-06-14
Add flag to fully disable compression for squashfs

Itxaka:

### DIFF
--- a/pkg/action/build-iso.go
+++ b/pkg/action/build-iso.go
@@ -144,12 +144,7 @@ func (b BuildISOAction) prepareISORoot(isoDir string, rootDir string, uefiDir st
 	}
 
 	b.cfg.Logger.Info("Creating squashfs...")
-	squashOptions := constants.GetDefaultSquashfsOptions()
-	if len(b.cfg.SquashFsCompressionConfig) > 0 {
-		squashOptions = append(squashOptions, b.cfg.SquashFsCompressionConfig...)
-	} else {
-		squashOptions = append(squashOptions, constants.GetDefaultSquashfsCompressionOptions()...)
-	}
+	squashOptions := append(constants.GetDefaultSquashfsOptions(), b.cfg.SquashFsCompressionConfig...)
 	err = utils.CreateSquashFS(b.cfg.Runner, b.cfg.Logger, rootDir, filepath.Join(isoDir, constants.IsoRootFile), squashOptions)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


pkg/action/build-iso.go
```diff
@@ -145,11 +145,14 @@ func (b BuildISOAction) prepareISORoot(isoDir string, rootDir string, uefiDir st
 
 	b.cfg.Logger.Info("Creating squashfs...")
 	squashOptions := constants.GetDefaultSquashfsOptions()
```
davidcassany: I think we could simply

```golang
squashOptions := append(constants.GetDefaultSquashfsOptions(), cfg.SquashFsCompressionConfig...)
```

pkg/elemental/elemental.go
```diff
@@ -301,7 +301,10 @@ func (e *Elemental) DeployImage(img *v1.Image, leaveMounted bool) error {
 			return err
 		}
 		if img.FS == cnst.SquashFs {
-			opts := append(cnst.GetDefaultSquashfsOptions(), e.config.SquashFsCompressionConfig...)
+			opts := cnst.GetDefaultSquashfsOptions()
```
davidcassany: same here, we could slightly simplify it.

pkg/types/v1/config.go
```diff
@@ -53,6 +53,7 @@ type Config struct {
 	Repos                     []Repository `yaml:"repositories,omitempty" mapstructure:"repositories"`
 	Arch                      string       `yaml:"arch,omitempty" mapstructure:"arch"`
 	SquashFsCompressionConfig []string     `yaml:"squash-compression,omitempty" mapstructure:"squash-compression"`
+	SquashFsNoCompression     bool         `yaml:"squash-no-compression,omitempty" mapstructure:"squash-no-compression"`
 }
 
 // Sanitize checks the consistency of the struct, returns error
```
davidcassany, line 59: We could use the Sanitize method to ensure `SquashFsCompressionConfig` is an empty slice if `SquashFsNoCompression` is set. Then we can simply trust the config flags anywhere.

davidcassany (approved):
Nice, looks good :+1:

I think we can slightly simplify the former logic though. I realized the SquashFsCompressionConfig is already initialized in  `pkg/config/config.go` to the default. So we can omit the checks in `build-iso.go` and `elemental.go` and directly trust `v1.Config.SquashFsCompressionConfig` to include the custom values, defaults or nothing if `SquashFsNoCompression` was selected.

pkg/types/v1/config.go
Itxaka: ahh good idea!

PR #259, opened 2022-06-15
Fix buld-iso

fgiudici:
We forgot to specify the arch in the default Luet repo, causing build-iso to fail (as the repo was dropped because not matching the arch).
The new raw image build tests moreover caused "make test" to fail in my environment: the three tests, when run in parallel, quickly fill up my /tmp dir (7GB). The proposal here is to serialize those test to let one clean-up before the following starts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,8 @@ test_deps:
 	go install github.com/onsi/ginkgo/v2/ginkgo
 
 test: $(GINKGO)
-	ginkgo run --label-filter !root --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -p -r ${PKG}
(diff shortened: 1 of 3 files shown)


fgiudici:
Regarding the tests, we could keep the full parallelism and move the generated files from /tmp to somewhere else on the regular disk. It looks anyway better to me to be somehow cautious with the maximum disk space we fill up (also if in the end we free all of it).

davidcassany (approved):
Looks good to me :+1:

~Just curiosity, are having any special feature the tests labeled with `serial`? Are they using  a different serial console or something like that?~

Edit: Just saw this is already explained within the PR message 😅

davidcassany:
> The new raw image build tests moreover caused "make test" to fail in my environment: the three tests, when run in parallel, quickly fill up my /tmp dir (7GB).

Do we really need that amount of space in for the raw_image tests, couldn't we make it without actually dumping serveral GBs? I know mocking command runs sucks, but probably we have chance there, it should also make tests way faster. Anyway, this is something to discuss/implement in the context of some other PR.

This is good as is :+1:

PR #260, opened 2022-06-16
Ensure we properly propagate repository arch configuration in repos

### LINKED ISSUE #1519: build-iso CI step does not build against the local repository
**cos-toolkit version:**

0.8.14-18

**CPU architecture, OS, and Version:**

x86_64

**Describe the bug**

The build iso step is not including the local repository due to a missing arch

This is critical as it makes our CI not to really test the code included and built within the PR

**To Reproduce**

**Expected behavior**

CI makes use of all packages modified in a PR

**Logs**

The issue can be seen here in this log traces
```
/usr/bin/elemental --debug --config-dir /home/runner/work/elemental-toolkit/elemental-toolkit build-iso --name cOS-Seed-teal-0.8.14-18-ge64b51f2-x86_64 --arch x86_64 --date --repo /home/runner/work/elemental-toolkit/elemental-toolkit/build
DEBU[2022-06-15T15:47:32Z] Starting elemental version v0.0.15 on commit 84c414baf593efee27c2439f3d12a814c9501a85 
DEBU[2022-06-15T15:47:32Z] Full config loaded: &v1.BuildConfig{
  Date: true,
  Name: "cOS-Seed-teal-0.8.14-18-ge64b51f2-x86_64",
  OutDir: "",
  Config: v1.Config{
    Logger: &v1.logrusWrapper{ // p0
      Logger: &logrus.Logger{
        Out: &os.File{},
        Hooks: logrus.LevelHooks{},
        Formatter: &logrus.TextFormatter{
          ForceColors: true,
          DisableColors: false,
          ForceQuote: false,
          DisableQuote: false,
          EnvironmentOverrideColors: false,
          DisableTimestamp: false,
          FullTimestamp: true,
          TimestampFormat: "",
          DisableSorting: false,
          SortingFunc: ,
          DisableLevelTruncation: false,
          PadLevelText: false,
          QuoteEmptyFields: false,
          FieldMap: logrus.FieldMap(nil),
          CallerPrettyfier: ,
        },
        ReportCaller: false,
        Level: 5,
        ExitFunc: os.Exit,
      },
    },
    Fs: &vfs.osfs{}, // p1
    Mounter: &mount.Mounter{},
    Runner: &v1.RealRunner{ // p2
      Logger: p0,
    },
    Syscall: &v1.RealSyscall{},
    CloudInitRunner: &cloudinit.YipCloudInitRunner{},
    Luet: &luet.Luet{
      VerifyImageUnpack: false,
      TmpDir: "",
      Arch: "x86_64",
    },
    Client: &http.Client{},
    Cosign: false,
    Verify: false,
    CosignPubKey: "",
    LocalImage: false,
    Repos: []v1.Repository{
      v1.Repository{
        Name: "",
        Priority: 30,
        URI: "quay.io/costoolkit/releases-teal",
        Type: "",
        Arch: "x86_64",
        ReferenceID: "",
      },
      v1.Repository{
        Name: "",
        Priority: 0,
        URI: "quay.io/costoolkit/releases-teal-arm64",
        Type: "",
        Arch: "arm64",
        ReferenceID: "",
      },
    },
    Arch: "x86_64",
    SquashFsCompressionConfig: []string{
      "-comp",
      "xz",
      "-Xbcj",
      "x86",
    },
  },
} 
DEBU[2022-06-15T15:47:32Z] Loaded LiveISO: &v1.LiveISO{
  RootFS: []*v1.ImageSource{
    &v1.ImageSource{},
  },
  UEFI: []*v1.ImageSource{
    &v1.ImageSource{},
  },
  Image: []*v1.ImageSource{
    &v1.ImageSource{},
    &v1.ImageSource{},
  },
  Label: "COS_LIVE",
  BootCatalog: "/boot/x86_64/boot.catalog",
  BootFile: "/boot/x86_64/loader/eltorito.img",
  HybridMBR: "/boot/x86_64/loader/boot_hybrid.img",
} 
INFO[2022-06-15T15:47:32Z] Preparing squashfs root...                   
INFO[2022-06-15T15:47:32Z] Copying system/cos source...                 
DEBU[2022-06-15T15:47:32Z] skipping repository '' for arch 'arm64'      
DEBU[2022-06-15T15:47:32Z] skipping repository '' for arch '' 
```

**Additional context**

davidcassany:
This commit ensures repository architecture is properly propageted from
`--repo` flag down to luet repository configuration.

In addition a repository without any architecture is set to the current
target architecture in v1.Config.

Fixes rancher/elemental-toolkit#1519

### DIFF
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -128,7 +128,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			}
 
 			for _, u := range repoURIs {
-				cfg.Repos = append(cfg.Repos, v1.Repository{URI: u, Priority: constants.LuetRepoMaxPrio})
+				cfg.Repos = append(cfg.Repos, v1.Repository{URI: u, Priority: constants.LuetRepoMaxPrio, Arch: cfg.Arch})
 			}
 
 			buildISO := action.NewBuildISOAction(cfg, spec)
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -122,7 +122,7 @@ func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Inter
 	// If a config file is found, read it in.
 	_ = viper.MergeInConfig()
 
-	// Bind runconfig flags
+	// Bind buildconfig flags
 	bindGivenFlags(viper.GetViper(), flags)
 	// merge environment variables on top for rootCmd
 	viperReadEnv(viper.GetViper(), "BUILD", constants.GetBuildKeyEnvMap())
@@ -133,6 +133,7 @@ func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Inter
 		cfg.Logger.Warnf("error unmarshalling config: %s", err)
 	}
 
+	err = cfg.Sanitize()
 	cfg.Logger.Debugf("Full config loaded: %s", litter.Sdump(cfg))
 	return cfg, err
 }
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -133,8 +133,6 @@ func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Inter
 		cfg.Logger.Warnf("error unmarshalling config: %s", err)
 	}
 
-	l.Arch = cfg.Arch
-
 	cfg.Logger.Debugf("Full config loaded: %s", litter.Sdump(cfg))
 	return cfg, err
 }
--- a/pkg/luet/luet.go
+++ b/pkg/luet/luet.go
@@ -47,7 +47,7 @@ type Luet struct {
 	plugins           []string
 	VerifyImageUnpack bool
 	TmpDir            string
-	Arch              string
+	arch              string
 }
 
 type Options func(l *Luet) error
@@ -99,7 +99,7 @@ func WithLuetTempDir(tmpDir string) func(r *Luet) error {
 
 func WithArch(arch string) func(r *Luet) error {
 	return func(l *Luet) error {
-		l.Arch = arch
+		l.arch = arch
 		return nil
 	}
 }
@@ -133,6 +133,10 @@ func NewLuet(opts ...Options) *Luet {
 	}
 
 	return luet
+}
+
+func (l *Luet) SetArch(arch string) {
+	l.arch = arch
 }
 
 func (l *Luet) SetPlugins(plugins ...string) {
@@ -186,6 +190,8 @@ func (l Luet) Unpack(target string, image string, local bool) error {
 // 3. Repo type is docker if the URL is of type [<dommain>[:<port>]]/<path>
 // Returns error if the type does not match any of any criteria.
 func (l Luet) initLuetRepository(repo v1.Repository) (luetTypes.LuetRepository, error) {
+	var err error
+
 	if repo.URI == "" {
 		return luetTypes.LuetRepository{}, fmt.Errorf("Invalid repository, no URI is provided: %v", repo)
 	}
@@ -209,6 +215,14 @@ func (l Luet) initLuetRepository(repo v1.Repository) (luetTypes.LuetRepository, 
 		}
 	}
 
+	arch := l.arch
+	if repo.Arch != "" {
+		arch, err = utils.ArchToGolangArch(repo.Arch)
+		if err != nil {
+			return luetTypes.LuetRepository{}, err
+		}
+	}
+
 	if repo.ReferenceID == "" {
 		repo.ReferenceID = "repository.yaml"
 	}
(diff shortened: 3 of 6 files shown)


davidcassany:
Draft since I have not yet adapted the unit tests neither validated it against https://github.com/rancher/elemental-toolkit/pull/1518, as this fix is required to ensure CI actually tests the PR code.

pkg/luet/luet.go
```diff
@@ -225,6 +239,7 @@ func (l Luet) initLuetRepository(repo v1.Repository) (luetTypes.LuetRepository,
 		Urls:        []string{repo.URI},
 		Type:        repoType,
 		ReferenceID: repo.ReferenceID,
+		Arch:        arch,
```
davidcassany, line 242: @mudler @Itxaka I have doubts of the implications of this change, I thought to just give it a try and discover whats bets in a trial and error mode, but probably you can give some insights here.

mudler: I think we have been here already . we have just to decide if letting luet handle the arch  and repository selection or either we just define the ones actually used (but we need to enable them).

That being said, specifying a arch here sounds ok to me.

frelon:
I think we will have to modify the skip condition as well to not skip repos without arch, since the local repo won't have an arch configured (or add arch in `luet repo add`)...

pkg/luet/luet.go
davidcassany: Right, I also thought that by doing this we could eventually omit our skipping logic, anyways having it does not hurt.

cmd/build-iso.go
```diff
@@ -128,7 +128,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			}
 
 			for _, u := range repoURIs {
-				cfg.Repos = append(cfg.Repos, v1.Repository{URI: u, Priority: constants.LuetRepoMaxPrio})
+				cfg.Repos = append(cfg.Repos, v1.Repository{URI: u, Priority: constants.LuetRepoMaxPrio, Arch: cfg.Arch})
```
davidcassany, line 131: In the case of `elemental build-iso --repo <custom-repo>` this sets the architecture to match `v1.Config.Arch` which is the same as `luet.Luet.Arch`. This covers building the iso with a local repo use case.

pkg/luet/luet.go
```diff
@@ -240,7 +255,8 @@ func (l Luet) UnpackFromChannel(target string, pkg string, repositories ...v1.Re
 	if len(repositories) > 0 {
 		repos = luetTypes.LuetRepositories{}
 		for _, r := range repositories {
-			if l.Arch != r.Arch {
+			// If the repository has no arch assigned matches all
+			if r.Arch != "" && l.arch != r.Arch {
```
davidcassany, line 259: Is repository arch is unset, we are not skipping it. So repositories without architecture defined will not be skipped and the `luet.Luet.Arch` will be used.

davidcassany (commented):
> I think we will have to modify the skip condition as well to not skip repos without arch, since the local repo won't have an arch configured (or add arch in `luet repo add`)...

I'd say this is already the case, see inline comments, probably I am missing something

cmd/config/config.go
```diff
@@ -133,8 +133,6 @@ func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Inter
 		cfg.Logger.Warnf("error unmarshalling config: %s", err)
 	}
 
-	l.Arch = cfg.Arch
```
fgiudici, line 136: At the beginning of the function we retrieve and set the architecture from config.NewBuildConfig(), shouldn't we set luet arch here too ?
I wouldn't remove it, otherwise looks to me we set it too late (after comparing luet arch vs repo arch, so luet arch would be empty).

davidcassany: For these kind of dependencies we have the `Sanitize` methods. Mapstructure utility does not provide postdecode hooks to digest field codependencies (aka `v1.Config.Arch` and `luet.Luet.arch` must match, but only `v1.Config.Arch` is decoded while parsing flags and yamls).

So the change in `pkg/types/v1/config.go` (line 66) serve that purpose. Note this is executed almost at the same stage, it is just a matter how the logic is organized.

There is an feature request in upstream for mapstructure here https://github.com/mitchellh/mapstructure/issues/253 even I doubt there is any interest on that from the maintainer 😓 

Gathering all this sort of checks once defaults, yamls and flags have been processed in a per type method opens the door of making use of an interface at some point if we see fit or the code grows.

fgiudici: I see, thanks for explaining!  I think what we miss than is a call to `cfg.Sanitize()` in `ReadConfigBuild` as we do in `ReadConfigRun`... did I got it correctly?

frelon:
> > I think we will have to modify the skip condition as well to not skip repos without arch, since the local repo won't have an arch configured (or add arch in `luet repo add`)...
> 
> 
> 
> I'd say this is already the case, see inline comments, probably I am missing something

Nice! I missed the comment on first read.

frelon (approved):

cmd/config/config.go
davidcassany: Damn! You are absolutely right, this should have been there since already a while 😓

fgiudici (approved):

davidcassany:
Ok, merging as I confirmed in rancher/elemental-toolkit#1518 that it fixes rancher/elemental#1519

Issue #265, opened 2022-06-10
Explore SCM integration with OBS automatically with service file 


mudler:
Outcome from discussion in https://github.com/rancher/elemental-toolkit/issues/1349:

Explore integration with OBS, where the Dockerfile built in OBS is living here in the source repository among the other source code and keep everything in one place:
- Have a Dockerfile.obs ( https://build.opensuse.org/package/view_file/home:kwk:elemental-testing:images/elemental-node-image/Dockerfile?expand=1 ) here. Take in mind that it will build only in OBS as there are injected automatically repositories.
 - CI should hook master to devel and tags to the non-testing repository. Dockerfile.e2e will have to point to the output of OBS from the testing base image, which contains tools built from master. When building master, the CI should build instead from the OBS output ( if feasable). At tags here corresponds tag of framework packages in OBS only, the CLI binary content under test will be annotated in a separate file

davidcassany:
I had a quick look about hooking elemental-cli in OBS to OBS changes. I wanted to apply the procedure described [here](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.source_service.html#sec.obs.sserv.token_usage), however it turns out we can't remotely run the `go_modules` service, I guess this is not installed within the OBS servers, this is quite frustrating as it could eventually turn into a blocker if the [workflow](https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html) concept does not cover our needs, moreover _worflows_ are still a beta feature in OBS, I have no news if this is likely to change anytime soon.

I am contacting _autobuild_ to ask about alternatives and options.

davidcassany:
According to autobuild `go_modules` service is on security audit. I understand it means is being analyzed to verify if it suits security standards prior being introduced in OBS servers.

mudler:
As agreed upon discussion, we will ship go mod vendor files in `vendor/`  inside the elemental-cli codebase. 

we can go back at this choice anytime

Itxaka:
After adding the vendored dir, I created the workflows file in my fork and create a package on obs and added the proper webhooks:

- workflow: https://github.com/Itxaka/elemental-cli/blob/main/.obs/workflows.yml
- package: https://build.opensuse.org/package/show/home:itxaka/elemental

This resulted in the service and package rebuilding on each main push, updating the sources and rebuilding the package.

So this seems to work as expected. There is also the push_tag event + branch_package action, so it should be possible to:

[...]

PR #266, opened 2022-06-22
Add vendor dir with deps

Itxaka:
Part of #265


davidcassany (approved):
Looks good :+1:

Itxaka:
I think that now we can drop the cache from the workflows as it wont be useful, having everything in the vendor directory.

davidcassany:
> I think that now we can drop the cache from the workflows as it wont be useful, having everything in the vendor directory.

That's a good point indeed

PR #269, opened 2022-06-22
Set serial tests with ginkgo dsl

Itxaka:
No need to filter ourselves, the nice folks from ginkgo already thougth
of this :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,7 @@ test_deps:
 	go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo
 
 test: $(GINKGO)
-	ginkgo run --label-filter '!root && !serial' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -p -r ${PKG}
-	ginkgo run --label-filter '!root && serial' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -r ${PKG}
+	ginkgo run --label-filter '!root' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -p -r ${PKG}
 
 test_root: $(GINKGO)
 ifneq ($(shell id -u), 0)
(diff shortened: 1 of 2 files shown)


Itxaka:
Im thinking that this should be impossible unless we are picking a wrong temp dir. AFAIK /tmp in the runners is mounted from the main disk, which should have around 20/30Gb free. The images are rounded up, yes but that means it should be a maximun of 3 images at the same time which are comprised of:

 - parts: maximum of 2200Mb (fixed image sizes, no matter the content)
 - raw image: 2200Mb as its the contents of the parts
 - Transformed image (azure rounded to 1Mb, gce to 1Gb so a max of 3Gb)
 
That would make the dir contain a max of 8Gb for GCE test, 4,5Gb for azure and 4Gb for simply the raw image.

Maybe things are not being cleaned up instead?

Itxaka:
oh well, just saw @fgiudici PR, its explained in there that it comes from a small /tmp dir (7Gb). Yup, that makes sense, but still we could be aggressive and clean up shit as soon as we are done with it, we would remove several GB. Notice that the raw image tests are slow so running them in serial introduces some time lost on our CI runs.

fgiudici:
> oh well, just saw @fgiudici PR, its explained in there that it comes from a small /tmp dir (7Gb). Yup, that makes sense, but still we could be aggressive and clean up shit as soon as we are done with it, we would remove several GB. Notice that the raw image tests are slow so running them in serial introduces some time lost on our CI runs.

Yep, makes sense. Not a problem for me to get back to full parallel tests... I was just thinking anyway the "casual" contributor wanting to run the "make tests" locally may hit this. But if this slows down too much our CI pipeline that is a no go...

fgiudici:
> No need to filter ourselves, the nice folks from ginkgo already thougth of this :)

Cool! (Tried it locally)

fgiudici (approved):

Itxaka:
oh shit, we forgot to add the different codecov outputs when splitting the tests so it was reporting a really low coverage :D

PR #275, opened 2022-06-23
Add squash flags to upgrade cmd

Itxaka:
In case of recovery being squash we need the flags available


Itxaka:
dummy, this flag is already there ¬_¬

PR #278, opened 2022-06-23
Produce state.yaml files on install, upgrade and reset commands

### LINKED ISSUE #261: Keep metadata of installed|upgrade system
In the rancher/elemental-toolkit#1516 PR I did there was an error making the system unbootalbe, however this passed all tests and got published. So our latest published image was not bootable. Then the PR to fix it (rancher/elemental-toolkit#1518) was not capable to pass tests because our CI includes upgrade|install tests that relay on upgrading to latest publised image, which was broken. Ouch 😓 

This was an unfortunate chain of errors, since there was a previous regression causing our CI to install published artifacts instead of whatever is built within the PR itself. Because of that the err in rancher/elemental-toolkit#1516 was not caught in the CI.

My take on all this story is that it is not obvious to trace back sources once you are running an Elemental OS installation (iso, disk image, installation, etc.), hence we are not validating the installation matches the expected versions or the checks we have are pretty weak (based on `/etc/os-release` value).

I believe we are missing a way easily know how the current system was build. For instance, if the system was build from docker image I'd like to easily find somewhere logged things like:

* version
* image full reference
* image sha
* install|upgrade hooks (if any)
...

Having an easy way to asses what is currently installed, makes it easier include sanity checks in our CI to ensure tests as passing with the expected artifacts and prevent issues like in rancher/elemental-toolkit#1519.

I am imagining some state data generated and maintained by `elemental-cli` so we can easily know what sources were used during installation|upgrade. Probably even keeping track of OS changes for full history.

davidcassany:
Fixes #261


davidcassany:
This PR creates a `state.yaml` file in recovery and state root partitions including the same exact data. The file includes system wide installation metadata such as modification date and some metadata for the main partitions (oem, recovery, state and persistent). Essentially it includes the label of the partition and the images deployed in it if any. Each image includes the label, filesystem and image source URI. For each image source there is also the source-metadata which varies depending on the source type (basically for `docker` and `channel` types).

Note this is slightly different form the issue description where we thought about having a devoted yaml file per image. I changed that to a system wide file because this way we can easily get all state data at once, including details like partition labels. This facilitates acknowledging passive, active and recovery differences after a single marshal.

I essentially found this very handy while trying to transparently support squashfs images for active and passive during the hackweek.

I also added the same file in recovery and state partitions because this way, regardless which system are we booting to we can easily find the `state.yaml` file into a known location and provides relevant information for updating or resetting the system without having to probe stuff and code heuristic logic.

Below there is an example of the state.yaml file that was installed and later on upgraded (passive and active differ):

 ```yaml
 # Autogenerated file by elemental client, do not edit

date: "2022-07-04T21:24:04Z"
oem:
    label: COS_OEM
persistent:
    label: COS_PERSISTENT
recovery:
    label: COS_RECOVERY
    recovery:
        source: channel://system/cos
        source-metadata:
            category: system
            name: cos
            version: 0.8.14-24
            finger-print: cos-system-0.8.14-24
            repositories:
                - name: cos
                  priority: 1
                  uri: quay.io/costoolkit/releases-teal
                  type: docker
        label: COS_SYSTEM
        fs: ext2
state:
    label: COS_STATE
    active:
        source: oci://quay.io/costoolkit/releases-teal:cos-system-0.8.14-23
        source-metadata:
            digest: sha256:81e191b8f7f9343c40ac23aeb8994453ae3d45642df9d1351d07abe9ec203ff9
            size: 773751864
        label: COS_ACTIVE
        fs: ext2
    passive:
        source: channel://system/cos
        source-metadata:
            category: system
            name: cos
            version: 0.8.14-24
            finger-print: cos-system-0.8.14-24
            repositories:
                - name: cos
                  priority: 1
                  uri: quay.io/costoolkit/releases-teal
                  type: docker
        label: COS_ACTIVE
        fs: ext2
```

File: pkg/luet/luet.go [hunk 1]
```diff
@@ -287,8 +292,38 @@ func (l Luet) UnpackFromChannel(target string, pkg string, repositories ...v1.Re
...
+
+	var meta *v1.ChannelImageMeta
+	if len(pkgs) > 0 {
+		meta = &v1.ChannelImageMeta{
+			Category:    pkgs[0].GetCategory(),
+			Name:        pkgs[0].GetName(),
+			Version:     pkgs[0].GetVersion(),
+			FingerPrint: pkgs[0].GetFingerPrint(),
+		}
+		//TODO: ideally we should only include the repository being used
```
davidcassany, line 313: @mudler do you know if there is a way to gather the repository that was actually used for the installation, not the full list? Also is there any way to get package specific metadata dependent on the repository type (checksums, digests, download/pull URI, etc.)?


I guess not, but I did not deeply followed Luet logic, so probably this area can be improved.

Makefile
```diff
@@ -44,20 +44,20 @@ test_deps:
 	go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo
 
 test: $(GINKGO)
-	ginkgo run --label-filter '!root' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt -p -r ${PKG}
+	ginkgo run --label-filter '!root' --fail-fast --slow-spec-threshold 30s --race --covermode=atomic --coverprofile=coverage.txt --coverpkg=github.com/rancher/elemental-cli/... -p -r ${PKG}
```
davidcassany, line 47: This change is not scritctly related to the PR purpose, however in this PR I noted there were few lines marked as not covered in the coverage analysis while I was certain they were...

basically this produces a coverage report taking into the account code parts that are on part of the package being tested in some specific suite.

pkg/elemental/elemental.go
```diff
@@ -155,6 +155,26 @@ func (e Elemental) UnmountPartitions(parts v1.PartitionList) error {
 	return nil
 }
 
+// MountRWPartition mounts, or remounts if needed, a partition with RW permissions
+func (e Elemental) MountRWPartition(part *v1.Partition) (umount func() error, err error) {
```
Itxaka, line 159: ni-ni-ni-niceeeee

Itxaka (approved):
looking really good!

pkg/luet/luet.go
mudler: `Sync()` returns actually all the repository metadata https://github.com/mudler/luet/blob/843858e24122219d99307125fc35a4bd4f2b1bc3/pkg/installer/repository.go#L877, seems we don't call it here explicitly.. maybe I'm wrong, anyhow under the hood is called by Install. 

We could also call it safely before as the subsequent calls would be cached, so there is guarantee that the second time you are still consuming the same repository in that timeframe - luet doesn't re-fetch repo data unless forced to do so, or 24h has been passed and repository index got updated.

PR #285, opened 2022-07-18
elemental install: allow to get the EJECT_CD option from the env

fgiudici:


mudler (approved):

pkg/constants/constants.go
```diff
@@ -216,6 +216,7 @@ func GetRunKeyEnvMap() map[string]string {
 		"poweroff": "POWEROFF",
 		"reboot":   "REBOOT",
 		"strict":   "STRICT",
+		"eject-cd": "EJECT_CD",
```
davidcassany, line 219: Good catch :+1:, this variable was not mapped on elemental-cli side. My bad, I should have reviewed that when we chose to drop `ros-installer` in favor of `elemental-operator`. These variables list was created based on `ros-installer` code few weeks ago.

davidcassany (approved):
Looks good

PR #293, opened 2022-07-25
Unify cmd-line usage

frelon:
This commit unifies the long/short description of commands and capitalization.


Itxaka:
Missing running go mod vendor, although I don't understand why this change would require that?

Itxaka (approved):

frelon:
> Missing running go mod vendor, although I don't understand why this change would require that?

`go mod vendor` also did nothing.. but `make test_deps` pulled through :+1:

PR #296, opened 2022-07-26
Add pxe build command

frelon:
Adds a new elemental subcommand called build and adds build-disk,
build-iso and build-pxe to that.

Build-PXE is essentially a slimmed down version of build-iso that
outputs the kernel, initrd, rootfs.squashfs and ipxe-config to the
output-directory.


pkg/action/build-pxe.go
```diff
@@ -0,0 +1,160 @@
...
+
+func NewBuildPXEAction(cfg *v1.BuildConfig, spec *v1.PXEConfig) *BuildPXEAction {
+	return &BuildPXEAction{
+		cfg:  cfg,
+		e:    elemental.NewElemental(&cfg.Config),
+		spec: spec,
+	}
+}
+
+// BuildISORun will build an ISO from a given configuration
```
ldevulder, line 49: Comment is not adapted to PXE boot I think ;-)

ldevulder (approved):
Very good! Just a minor comment. Not a Go expert, but looks good.

I think that we can merge if it works, no need to wait for a full codecov coverage.

Itxaka:
This is now too old to keep around, closing it for now.

PR #298, opened 2022-07-29
Log error on upgrade

frelon:
Encountered this error and did not get a reason why it happend :shrug: 


Itxaka (approved):
Good catch, I hate when this happens.

PR #299, opened 2022-07-29
Allow overriding default grub entry or fallback to default

Itxaka:
The way this was done before it didnt allow a derivative to override the
default grub entry by setting the GRUB_ENTRY_NAME in the os-release file
as it will only look up that if the entry was empty, but we were always
initializing it to the defaul cOS value.

This patch makes the default grub empty on the config empty by default
and falls back to it in case we cant load it from the os-release file


frelon (approved):
Nice! :+1:

PR #300, opened 2022-07-29
Set the message of what we are setting as default properly

Itxaka:
I was using the constant instead of the actual value...


kkaempf (approved):
🤞🏻

PR #302, opened 2022-08-03
Add space after 'Downloading'

kkaempf:
Otherwise the log looks a bit strange:
```
Downloadingquay.io/costoolkit/releases-green:repository.yaml
```


Itxaka (changes requested):
yikes, nope. This is changing a vendor lib.

Itxaka:
The real problem with this is that we have different loggers for our stuff and the logger by default in Luet, which accepts a list of strings and adds spaces between them.

The fix is to change our logger to also deal with this type of input

kkaempf:
Will create an issue instead

PR #305, opened 2022-08-05
Exclude /etc/resolv.conf from syncing during upgrade

mjura:
/etc/resolv.conf shouldn't never be copied from source directory
to target OS image. It will be generated by network service during booting.

Fixes https://github.com/orgs/rancher/projects/9

Signed-off-by: Michal Jura <mjura@suse.com>

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -369,7 +369,7 @@ func (e *Elemental) DumpSource(target string, imgSrc *v1.ImageSource) (info inte
 			return nil, err
 		}
 	} else if imgSrc.IsDir() {
-		excludes := []string{"/mnt", "/proc", "/sys", "/dev", "/tmp", "/host", "/run"}
+		excludes := []string{"/etc/resolv.conf", "/mnt", "/proc", "/sys", "/dev", "/tmp", "/host", "/run"}
 		err = utils.SyncData(e.config.Logger, e.config.Fs, imgSrc.Value(), target, excludes...)
 		if err != nil {
 			return nil, err


pkg/elemental/elemental.go
```diff
@@ -369,7 +369,7 @@ func (e *Elemental) DumpSource(target string, imgSrc *v1.ImageSource) (info inte
 			return nil, err
 		}
 	} else if imgSrc.IsDir() {
-		excludes := []string{"/mnt", "/proc", "/sys", "/dev", "/tmp", "/host", "/run"}
+		excludes := []string{"/etc/resolv.conf", "/mnt", "/proc", "/sys", "/dev", "/tmp", "/host", "/run"}
```
mudler, line 372: Just a bit unsure here.. wouldn't break the installation, avoiding to populate the resolv.conf file initially?

Itxaka: I'm also not sure about this, if this was the issue, for sure it would have appeared before on the toolkit tests?

Plus on each reboot yip the network service would make sure to regenerate the resolv file when it's started no?

mjura: `/etc/resolv.conf` should not be populated during installation or even during upgrade process, because it will be updated by network service (wicked or NetworkManager) during system startup 
Right now during copying source filesystem from container context will copy also `/etc/resolv.conf` from container which is static file and make it read-only on host system. During system startup NetworkManager or wicked will check it and not update it leaving configuration from container image.

Itxaka: Im wondering about other derivatives that are based on alpine for example. Do they manage this resolv.conf file in a different way? maybe @mudler knows as he has an alpine based derivative...

mudler: What looks really off to me is that we would exclude a file from /etc. I don't think the installer should be aware of what to exclude or what not from a rootfs perspective, it should just copy what's there

mudler: I'm not a fan of this approach - because nothing would forbid to have a rootfs with an hard coded resolv.conf, or even ship some defaults. This approach is not obvious to me. I'd rather go on the other side of this and try to preserve it while applying the upgrade plan.

mjura: Any hardcoded values for NAMESERVERS in Linux should be set as one of following variable in /etc/sysconfig/network/config file, this is also telling us comment from /etc/resolv.conf
```
# Before you change this file manually, consider to define the
# static DNS configuration using the following variables in the
# /etc/sysconfig/network/config file:
#     NETCONFIG_DNS_STATIC_SEARCHLIST
#     NETCONFIG_DNS_STATIC_SERVERS
#     NETCONFIG_DNS_FORWARDER
```

System should be able to auto-generate this file `/etc/resolv.conf` based on network services configuration in our case wicked or NetworkManager. With current state in case of installation or upgrade this file can be set as read-only in source root filesystem.

For me fixing this in `elemental-cli` is still the best, but I have proposed also other solution
https://github.com/rancher/elemental/pull/261

I will look also possibility of fixing this in https://github.com/rancher/elemental/blob/master/framework/files/usr/sbin/suc-upgrade

mjura:
@mudler here is fix which we have talked about https://github.com/rancher/elemental/pull/265

mjura:
Patch https://github.com/rancher/elemental/pull/265 was merged

PR #306, opened 2022-08-08
Add extra partitions config on installation

### LINKED ISSUE #304: Partition function cannot create custom/additional partition for installation
In past elemental-cli supported create additional partitions.
elemental-cli 0.0.15 didn't have this feature.

Harvester will create a Longhorn partition in the OS disk in some conditions.
So we need to send config to elemental-cli, let it create partition for harvester to use.

Thanks a lot

ref: https://github.com/harvester/harvester/issues/2565

Itxaka:
This adds a new key to the install config that allows to pass a list of
extra partitions to add during install.

Same keys as the usual partitions exists, name, label, fs, mountpoint,
etc..

Also allows to override the sizes and reorders the partition creation to
take that into account, as the partition with size == 0 (which means use
all the free space available) has to go the last.

This allows creating extra partitions with size 0 and will reorder the
partitions properly to account for this. Also forbids having several
partitions with size == 0 or a persistent and extra partition with both
having size == 0

Fixes https://github.com/rancher/elemental-cli/issues/304

### DIFF
--- a/pkg/types/v1/config.go
+++ b/pkg/types/v1/config.go
@@ -385,7 +385,6 @@ func (ep ElementalPartitions) PartitionsByInstallOrder(extraPartitions []*Partit
 	if ep.Persistent != nil && !inExcludes(ep.Persistent, excludes...) {
 		// Check if we have to set this partition the latest due size == 0
 		if ep.Persistent.Size == 0 {
-			fmt.Printf("Setting persistent to latest")
 			lastPartition = ep.Persistent
 		} else {
 			partitions = append(partitions, ep.Persistent)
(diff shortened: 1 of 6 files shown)


pkg/types/v1/config.go
```diff
@@ -364,8 +383,34 @@ func (ep ElementalPartitions) PartitionsByInstallOrder(excludes ...*Partition) P
...
+	}
+	for _, p := range extraPartitions {
+		// Check if we have to set this partition the latest due size == 0
+		// Also check that we didn't set already the persistent to last in which case ignore this
+		// InstallConfig.Sanitize should have already taken care of failing if this is the case, so this is extra protection
+		if p.Size == 0 {
+			if lastPartition != nil {
+				// Ignore this part, we are not setting 2 parts to have 0 size!
+				continue
+			} else {
```
fgiudici: nit: you could drop the `else` here as the if branch will "continue"

fgiudici (approved):
Looks good here!

PR #310, opened 2022-08-09
Drop luet init from BuildConfig

### LINKED ISSUE #175: build-disk command leaves a `repos` dir where its launched from
master version

Probably missing setting luet tmpdir

To reproduce:
`elemental build-disk`

What happens:
 a repos directory is left in the current dir

What is expected:
No leftovers

Itxaka:
We are already calling NewConfig inside BuildConfig and that will make
sure that a luet instance is properly created.

This fixes the issue of having temp files in the dir running elemental
as the BuildConfig luet was misconfigured with no tempDir and no FS

Fixes #175 


Itxaka:
lint issues are fixed on https://github.com/rancher/elemental-cli/pull/308

kkaempf (approved):
Dropping things is always welcome !

PR #311, opened 2022-08-31
Fix image signing

Itxaka:
The meta outputs is more than a line and that breaks signing, lets use
the tags directly


davidcassany (approved):
LGTM

PR #316, opened 2022-09-13
Include bootloader arrangement in build-iso

davidcassany:
Related rancher/elemental#263

Note this PR also drops non EFI ISO generation support. With this all ISOs are expected to be grub2 based and hybrid, they boot on EFI, non EFI and as disk or cdrom. Legacy firmware only ISOs based on syslinux are not considered.

If `bootloader-in-rootfs` flag is not provided we keep previous behavior, however if provided, the new logic kicks in and then the main OS image is expected to include all required packages and binaries to set the bootloader.

### DIFF
--- a/cmd/build-iso.go
+++ b/cmd/build-iso.go
@@ -144,6 +144,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
 	c.Flags().StringArray("repo", []string{}, "A repository URI for luet. Can be repeated to add more than one source.")
+	c.Flags().Bool("bootloader-in-rootfs", false, "Fetch ISO bootloader binaries from the rootfs")
 	addArchFlags(c)
 	addCosignFlags(c)
 	addSquashFsCompressionFlags(c)
--- a/pkg/live/common.go
+++ b/pkg/live/common.go
@@ -1,0 +1,71 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package live
+
+import (
+	"github.com/rancher/elemental-cli/pkg/constants"
+)
+
+const (
+	efiBootPath           = "/EFI/BOOT"
+	isoLoaderPath         = "/boot/x86_64/loader"
+	grubArm64Path         = grubPrefixDir + "/arm64-efi"
+	grubEfiImagex86Dest   = efiBootPath + "/bootx64.efi"
+	grubEfiImageArm64Dest = efiBootPath + "/bootaa64.efi"
+	grubCfg               = "grub.cfg"
+	grubPrefixDir         = "/boot/grub2"
+	//TODO use some identifer known to be unique
+	grubEfiCfg = "search --no-floppy --file --set=root " + constants.IsoKernelPath +
+		"\nset prefix=($root)" + grubPrefixDir +
+		"\nconfigfile $prefix/" + grubCfg
+
+	// TODO not convinced having such a config here is the best idea
+	grubCfgTemplate = `search --no-floppy --file --set=root /boot/kernel                               
+	set default=0                                                                   
+	set timeout=10                                                                  
+	set timeout_style=menu                                                          
+	set linux=linux                                                                 
+	set initrd=initrd                                                               
+	if [ "${grub_cpu}" = "x86_64" -o "${grub_cpu}" = "i386" -o "${grub_cpu}" = "arm64" ];then
+		if [ "${grub_platform}" = "efi" ]; then                                     
+			if [ "${grub_cpu}" != "arm64" ]; then                                   
+				set linux=linuxefi                                                  
+				set initrd=initrdefi                                                
+			fi                                                                      
+		fi                                                                          
+	fi                                                                              
+	if [ "${grub_platform}" = "efi" ]; then                                         
+		echo "Please press 't' to show the boot menu on this console"               
+	fi                                                                              
+	set font=($root)/boot/${grub_cpu}/loader/grub2/fonts/unicode.pf2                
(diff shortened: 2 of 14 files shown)


Itxaka:
Note that for the efi stuff iso, we only need a grub.efi and a grub.cfg. We can generate the grub.cfg in the elemental-cli and the grub.efi can come from the rootfs directly, no need to use toolkit packages for that, we only need to copy it to /EFI/BOOT/bootx64.efi and drop the grub.cfg into the same place.

And for the boot modules, I think for the iso the default grub should contain evertyhing, so Im willing to say that it would be possible to drop them directly.

davidcassany:
> And for the boot modules, I think for the iso the default grub should contain evertyhing, so Im willing to say that it would be possible to drop them directly.

I am not sure what are you referring to, but note most of the logic we have in `packages/livecd/grub2` build recipe is related to support legacy bios firmware, not efi. I could not find a prebuild core.img for i386-pc.

But yes there several files we are including that are not needed in those packages.

For ISOs the boot matrix expands to 4 as we build hybrid images. In short all these should boot fine:

```
qemu-kvm -m 2048 -cdrom test.iso
qemu-kvm -m 2048 -hda test.iso
qemu-kvm -m 2048 -cdrom test.iso -bios /usr/share/qemu/ovmf-x86_64.bin
qemu-kvm -m 2048 -hda test.iso -bios /usr/share/qemu/ovmf-x86_64.bin
```

Then we just need to include two more for the secure boot 😅 , but one step at a time.

kkaempf:
@agracey 👆🏻 - our test/support matrix is expanding. (And will double again when we add ARM). Can we limit it somehow ?

davidcassany:
Finally this is something is consider good for a first round. This produces functional ISOs with the following command and setup:


```
elemental --config-dir . build-iso docker:registry.opensuse.org/home/dcassany/elemental/teal53/15.4/rancher/elemental-teal/5.3:15.1
```

Having a manifest.yaml file like that in current dir:
```yaml
iso:
  bootloader-in-rootfs: true
  grub-entry-name: Elemental
  label: "ELEM_LIVE"

name: "test"
```

I used a branched node image to include the syslinux package in it.

agracey:
I don't think we need to support legacy boot modes going forward

pkg/live/green.go
```diff
@@ -0,0 +1,178 @@
...
+	"path/filepath"
+
+	"strings"
+
+	"github.com/rancher/elemental-cli/pkg/constants"
+	v1 "github.com/rancher/elemental-cli/pkg/types/v1"
+	"github.com/rancher/elemental-cli/pkg/utils"
+)
+
+type GreenLiveBootLoader struct {
```
davidcassany, line 30: I named it green as it follows the SUSE procedure, there are a bunch constants related to paths of installed binaries that are SUSE specific. Making it cross distro would require heuristics to determine grub2, syslinux and tool names or eventually code recipes for each _flavor_ and define the flavor as a config flag.

We never had bootloader artifacts on ISOs based on any other distro beyond openSUSE since they were not taken from the OS container and they were the same for all flavors. I'll have a look to Fedora and Ubuntu's paths of these tools to check if I can manage to find some simple and robust system adapt it in a single logic to make compatible for all, however I fear it might not be worth.

pkg/live/common.go
```diff
@@ -0,0 +1,103 @@
...
+package live
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/rancher/elemental-cli/pkg/constants"
+)
+
+const (
```
davidcassany, line 26: These constants in common.go file are values that are arbitrary or defined by us. Those are not SUSE specific and I would expect them to be reusable for other flavors.

Itxaka:
> I don't think we need to support legacy boot modes going forward

:eyes: 

Does that mean dropping support for booting bios isos and systems?????????? That would make everything really easy lol

agracey:
I don't see any reason why to do the extra work. IMO, the only reason to keep it around would be for older systems but we would typically be included in rollouts of new hardware that should not be using a bios from the 25 years ago.

We may want to document how to install the UEFI firmware on raspberry pis (or link to a guide we trust)

Itxaka:
damn, then this changes things. For EFI removable devices It its as simple as:

  - dd if=/dev/zero of=isorootfs/EFI/BOOT/efiboot.img bs=1M count=10
  - mkfs.msdos -F 12 -n 'EFI_BOOT' isorootfs/EFI/BOOT/efiboot.img
  - mmd -i isorootfs/EFI/BOOT/efiboot.img ::EFI
  - mmd -i isorootfs/EFI/BOOT/efiboot.img ::EFI/BOOT
  - mcopy -i isorootfs/EFI/BOOT/efiboot.img grub.efi ::EFI/BOOT/BOOTX64.EFI
  - xorriso -as mkisofs -V 'MY_ISO' -e EFI/BOOT/efiboot.img -no-emul-boot -o OUTPUT_ISO isoroot/

That would simplify things. Also on the grub part, we could drop the non-efi install which would also make it easier.

davidcassany:
> I don't think we need to support legacy boot modes going forward

I agree it is unlikely to face the need to provide legacy firmware support for any production deployment these days, however note that BIOS support is handy for testing with hipervisors. In fact, some of our tests make use of VirtualBox to run VMs in github workers and those are running on legacy firmware because EFI support in VirtualBox is just buggy and not reliable.

Itxaka:
> I agree it is unlikely to face the need to provide legacy firmware support for any production deployment these days, however note that BIOS support is handy for testing with hipervisors. In fact, some of our tests make use of VirtualBox to run VMs in github workers and those are running on legacy firmware because EFI support in VirtualBox is just buggy and not reliable.

Maybe we can just move those to qemu? We only use those for the raw image + installer tests IIRC, so it should be fairly straightforward. In fact it will probably be easier than dealing with vBox lol

cmd/build-iso.go
```diff
@@ -144,6 +144,7 @@ func NewBuildISO(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 	c.Flags().String("overlay-iso", "", "Path of the overlayed iso data")
 	c.Flags().String("label", "", "Label of the ISO volume")
 	c.Flags().StringArray("repo", []string{}, "A repository URI for luet. Can be repeated to add more than one source.")
+	c.Flags().Bool("bootloader-in-rootfs", false, "Fetch ISO bootloader binaries from the rootfs")
```
kkaempf, line 147: 👍🏻 🥳

pkg/live/green.go
kkaempf: 👍🏻

kkaempf (approved):
What kind of documentation changes to we need ?

davidcassany:
> What kind of documentation changes to we need ?

From user PoV there are no changes, so probably nothing on `rancher/elemental`, but the new option should be documented somehow into the `rancher/elemental-toolkit` docs.

Itxaka (approved):

PR #317, opened 2022-09-14
Rework efi grub install

Itxaka:
- drop grub2-install on efi
 - copy efi artifacts to ESP (shim, mokmanager, grub)
 - generate grub.cfg that chainloads into state partition grub
 - run efibootmgr to create the entry for the shim
 - drop --removable from the grub install


Itxaka:
This basically changes the EFI state.

We used to have a "--removable" flag on the grub-install which caused the ESP to contain only one file at `/EFI/BOOT/BOOTX64.EFI`, which is the "fallback bootloader". That means that it was the last thing to be launched on efi boot and would not be launched if something else was in the boot manager. This filename was originally intended for use only on removable media, so that they could be booted to install an OS. (If not for that fixed filename, a factory-fresh computer wouldn't be able to boot anything!).

This also meant that we were not creating any entries on the bootloader for elemental. And were booting grub directly.

Also, Im not sure how the hell the grub.cfg was loading? Im guessing because we built the grrub.efi with the config, the config was somehow embedded or something, no idea.

This needed change. Now what we do is copy the signed shim.efi, the mokmanager.efi, and the grub.efi instead of generating one, and create an entry on the boot manager that points to the shim. Also create a grub.cfg that loads the real grub.cfg from the state partition.

This results in the shim launching the grub.efi automatically, which loads the grub.efi in the ESP partiton and that loads the real grub.cfg from the state partition.

This should solve a few problems like what to boot after restart. Seems like newly added boot entries have preference over dvd drives, so we may not even need the eject-cd stuff anymore.

It also means that once we have a signed grub we can substitute it and we should be able to boot under secureboot as the rest of the chain is already signed.

Also means that we are more in-line with the standard installation of other systems on EFI without breaking the whole ESP partition and taking over it, which means that we could even dual-boot between elemental and other OS (LMAO)

This should work for both normal and secureboot systems

Itxaka:
test and lint added

Itxaka:
https://github.com/rancher/elemental-cli/pull/317

Itxaka:
not sure why codecov is failing due to the patch diff, we dont have that configured...

pkg/utils/grub.go
```diff
@@ -120,13 +110,82 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool)
...
 	if err != nil {
 		return err
 	}
 
-	g.config.Logger.Infof("Grub install to device %s complete", target)
+	if efi {
+		// Copy required extra modules to boot dir under the state partition
+		// otherwise if we insmod it will fail to find them
+		// We no longer call grub-install here so the modules are not setup automatically in the state partition
+		// as they were before. We now use the bundled grub.efi provided by the shim package
```
davidcassany, line 123: I believe secure boot should be optional, this will be a nightmare to keep it consistent and up to date for all flavors (green, blue, orange and I think someone did alpine trials too).

Probably `grub2-install` wasn't that bad after all as this should exist more or less in a similar way in all major distros 🤔 Is there any way to prevent grub2-install to create its own image? Probably there is as `shim-install` helper script calls it as far as I remember.

Another alternative here could be to find/guess relevant files in the file system, in a similar fashion kernel and initrd are found for ISO build, there are already some utilities to find files based on patterns. Probably we can handle some generic process that matches at least opensuse, fedora and ubuntu.

davidcassany (commented):
Just concerned about keeping our code distro agnostic...

pkg/utils/grub.go
```diff
@@ -120,13 +110,82 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool)
...
+		for _, m := range []string{"loopback.mod", "squash4.mod"} {
+			fileReadName := filepath.Join(rootDir, fmt.Sprintf("/usr/share/grub2/%s-efi/%s", g.config.Arch, m))
```
Itxaka: should probably search for this artifacts instead of hardcoding the path? Not fully understanding the implications of this tbh.

Im guessing this means that any source for the install needs grub2  packages installed.

pkg/utils/grub.go
```diff
@@ -120,13 +110,82 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool)
...
+		if exists, _ := Exists(g.config.Fs, filepath.Join(rootDir, fmt.Sprintf("/usr/share/efi/%s/", g.config.Arch))); exists {
+			g.config.Logger.Infof("Generating grub files for efi on %s", target)
+			err = MkdirAll(g.config.Fs, filepath.Join(cnst.EfiDir, "EFI/elemental/"), cnst.DirPerm)
+			if err != nil {
+				g.config.Logger.Errorf("Error creating dirs: %s", err)
+				return err
+			}
+
+			// Copy needed files for efi boot
+			for _, f := range []string{"shim.efi", "MokManager.efi", "grub.efi"} {
```
Itxaka: should probably search for this artifacts instead of hardcoding the path? Not fully understanding the implications of this tbh.

Im guessing this means that any source for the install needs shim package installed.

pkg/utils/grub.go
```diff
@@ -120,13 +110,82 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool)
...
+			}
+			// Create entry in bootmanager
+			// -c create
+			// -d disk
+			// -p partition containing loader
+			// -w write unique sig to MBR if needed
+			// -L label
+			// -l loader (inverted slashes!)
+			efibootmgrArgs := []string{"-c", "-d", target, "-p", "1", "-w", "-L", "elemental", "-l", "\\EFI\\elemental\\shim.efi"}
+			out, err := g.config.Runner.Run("efibootmgr", efibootmgrArgs...)
```
Itxaka: this requires efibootmgr package installed :/

Itxaka: So this only works if you install from the same machine into local disk. If you install into a loop device (for testing?) then you get an entry in you local machine :/

pkg/utils/grub.go
Itxaka: secureboot is optional anyways, the shim does not force you to have secureboot, it just helps you by settings things up in case you enable secureboot.

Currently reworking this to auto find the artifacts instead of hardcoding them. Should work for all systems as long as they have the shim package in source rootfs.

Itxaka:
keep in mind that this is very experimental :)

unfortunately there is not a blocked tag in elemental-cli :D

pkg/utils/grub.go
Itxaka: Im thinking that grub2-install could work down the line to generate the grub image and the artifacts with a proper bootloader-id set to it.

but we still would need the shim package and the efibootmgr to create the entry pointing to the shim. Gonna do some tests to see if you can boot to a signed grub without using shim under secureboot (Its possible with no secureboot) but this would be future proof in we want to support it in the future (once grub bundles the required modules, that is the only missing thing)

Itxaka: files installed with just grub2-install do not seem to be signed as expected.
a signed grub also copied and booted directly works under non-secureboot but doesnt under secureboot.
a signed shim with a signed grub boots correctly on both secure and non secure boot systems.

Seems like shim -> grub is the only way to go to support both scenarios.

Itxaka: Regarding the flavors...we really dont support those, nor we test them (last time I tested a cos generated images for other flavors, most did not work...we dont even know if our elemental-cli changes are even compatible with them)

If someone is using those, they can come and contribute patches to the installer or cos themselves, but the only flavor we support and test against is teal. Does it make sense to take forward decisions based on those untested flavors? It would be cool if they work, and will try to make it as system agnostic as possible, but we should not dictate our direction based on those flavors IMHO. We need to support our current usecase, and for our maintained flavor, shim -> grub is the future-proof way of going.

davidcassany: It is not a matter of supporting other flavors or not, it is a matter of making our toolkit capable to produce those, so minimize the expectations from `elemental-cli` regarding the root tree setup. Up to now we claim we are only tied to dracut and systemd (we should add grub2, but I guess it is acceptable to omit since everyone supports it) and we should keep that. 

I did a check on how secure boot images are created in KIWI, just out of curiosity since it also supports most major distros. I realized it uses the same approach for disk and ISOs, it basicaly copies the shim.efi file as `bootx64.efi` includes the `grub.efi` and the mokmanager. grub2-install is only used for `i386-pc`. I assume this is to prevent having to deal with the efibootmngr and nvram entries. We have also been using this setup up to now and so I am wondering if this that bad, this suposed to be used as a fallback specially though for removable devices, but what's wrong on using it for the persistent disk?

Itxaka:
@davidcassany how about now? We only require that the artifacts are on the install source and that the dir in which they are has the arch, which should be the same in every distro as artifacts can be for different arches and the packages should not be tied to any arch (i.e. you can install the aarch64 artifacts in an x86 distro) and have them along like mostly everything on grub related stuff, like installing i386-grub artifacts/modules on aarch64 systems.

I mean, this is at least the minimal we need to have first a decent boot procedure, and second support secureboot once grub ships those modules. Plus it cleans a bit the grub install.

pkg/utils/grub.go
Itxaka: well, for one, if bootmanager entries are broken that is the name that will be used for the fallback. Second, there is nothing preventing something else from overwriting it as its a generic name for fallback, so we may not know but other utilities may overwrite it. Third, by setting it as removable it wont create an entry in bootmanager so the only way any system installed by elemental to boot its if there is nothing else in the bootmanager, which on some efi config wont be possible (I got on my an entry to upgrade the efi system added by the efi system)

Not only that but the efi dir is supposed to have subdirs with the respective systems setting their files in there instead of using the root dir for that, see my local system below:

```
 ls -ltra /boot/efi/EFI              
total 24
drwxr-xr-x 3 root root 16384 Jan  1  1970 ..
drwxr-xr-x 2 root root  2048 Aug 10 17:47 boot
drwxr-xr-x 3 root root  2048 Aug 10 20:43 opensuse
drwxr-xr-x 3 root root  2048 Aug 10 20:45 Dell
drwxr-xr-x 5 root root  2048 Aug 10 20:45 .
```

I mean, yeah we can use that and its much simpler but then forget about supporting secureboot ever and expect some issues to come from that as we are not using correctly. Mainly we are abusing a fallback mechanism which yeah, works. This is the issue the metal3 prototype hit, in which after an install they were unable to boot properly into elemental and it kept booting from the dvd

We can do it much better and avoid issues down the line and prepare for secureboot feature with this changes, not sure why we would not do it.

> I realized it uses the same approach for disk and ISOs, it basicaly copies the shim.efi file as bootx64.efi includes the grub.efi and the mokmanager. grub2-install is only used for i386-pc. I assume this is to prevent having to deal with the efibootmngr and nvram entries.

That is weird because its not the way of doing it. Opensuse does not do that, they create their dir and store the files in there and create an entry for it in the bootmanager, so it may be for cloud disks and isos(that makes sense, you dont want to relay into the bootmanager having entries as there are no installs in that case), but on the install they do the right thing:

```
$ ls -ltra /boot/efi/EFI/opensuse     
total 3172
-rwxr-xr-x 1 root root   62840 May  7 23:34 fwupdx64.efi
drwxr-xr-x 2 root root    2048 Aug 10 20:43 fw
drwxr-xr-x 3 root root    2048 Aug 10 20:43 .
drwxr-xr-x 5 root root    2048 Aug 10 20:45 ..
-rwxr-xr-x 1 root root  934680 Sep 13 14:08 shim.efi
-rwxr-xr-x 1 root root  846240 Sep 13 14:08 MokManager.efi
-rwxr-xr-x 1 root root  143360 Sep 13 14:08 grubx64.efi
-rwxr-xr-x 1 root root 1247232 Sep 13 14:08 grub.efi
-rwxr-xr-x 1 root root     150 Sep 13 14:08 grub.cfg
-rwxr-xr-x 1 root root      58 Sep 13 14:08 boot.csv

BootCurrent: 0000
Timeout: 1 seconds
BootOrder: 0002,0000,000A,0001
Boot0000* opensuse-secureboot	HD(1,GPT,7962bf87-25dc-4e13-8467-fb74a6e93b42,0x800,0x32000)/File(\EFI\opensuse\shim.efi)
```

And its the same on a freshly installed system :eyes: 

> Up to now we claim we are only tied to dracut and systemd (we should add grub2, but I guess it is acceptable to omit since everyone supports it) and we should keep that.

We can keep claiming that and add shim for efi systems. Also everybody supports it and uses the same code (after all the shim artifact is the only one signed by Microsoft, so everyone needs to use the same one and use the same workflow :D)



TL;DR: For cloud images/isos where there is no install it makes sense to have it. For normal installations it does not. Way of doing it is to add an entry to boot manager. Way to support secureboot in the future is to always use the shim. shim works in non-secureboot systems equally. future gains VS keeping it "wrong" and dealing with future issues.

davidcassany: > That is weird because its not the way of doing it. Opensuse does not do that, they create their dir and store the files in there and create an entry for it in the bootmanager, so it may be for cloud disks and isos(that makes sense, you dont want to relay into the bootmanager having entries as there are no installs in that case), but on the install they do the right thing:

Well, not that weird, consider the image builder is never executed over the system that will actually run the image. This is the problem we would have by running an installation over a loop device. The problem of setting efi entries is that this belongs to the host, to the current EFI firmware. I am not sure how secure boot is supported/used in public cloud, but I guess there aren't many other options beyond relaying on fallback. I know this is not much of a problem for our current specific use case, but still would be nice being capable to build disk images. I know we are not targeting cloud but being capable to build images for cloud would be nice, we already spent quite some time to ensure we were capable to do it a year ago 😅 

Would it make sense leaving the code ready for several grub installation styles? Probably we can start by making nvram updates by default and having a flag to deactivate that, probably `removable` flag to use the fallback mechanism?

I am not complaining of using efibootmngr and setting the appropriate efi entries, it makes complete sense for the edge bare metal case we are targeting. I am just worried about always applying this setup in all cases and breaking some of the use cases already documented in elemental-toolkit (loop device install, cloud image builds, etc.).

davidcassany:
> I mean, this is at least the minimal we need to have first a decent boot procedure, and second support secureboot once grub ships those modules. Plus it cleans a bit the grub install.

I absolutely agree on that. We just need to be careful to not break some of the features or use cases we considered in for the elemental-toolkit. Since this is part of the elemental-cli binary it gets buried pretty deep in the stack.

pkg/utils/grub.go
Itxaka: Im not sure what you mean with not capable of building disk images? This only affect to the install and reset commands which is the only part of the code that calls the `grub.Install`, the only thing that we would miss is the install to loop devices, which AFAIK is only done for testing and dev and can be easily fixed by not running the efibootmanager if the target is a loop device?


On cloud images and isos you NEED to use the fallback mechanism because you boot from the media directly, you havent had any time to configure the bootmanager, so the fallback mechanism is the only reason you can boot from those devices be it an iso or a qcow2 raw disk :)


BTW, fedora install does exactly the same as this patch: https://github.com/rhinstaller/anaconda/blob/f0df778cc43e2494fdd0fad131aa8643590510ac/pyanaconda/modules/storage/bootloader/efi.py#L71 gathers the files into their proper places and adds an entry(they are more intelligent than us as their efi packages drop all the files on the /boot/efi dir which is automounted to the EFI partition LOL) to the bootmanager.

davidcassany: > Im not sure what you mean with not capable of building disk images?

I mean a state less installation over a loop device.

> On cloud images and isos you NEED to use the fallback mechanism because you boot from the media directly, you havent had any time to configure the bootmanager, so the fallback mechanism is the only reason you can boot from those devices be it an iso or a qcow2 raw disk :)

Exactly, similar case to virtualized deployments. That's why I am wondering about coding both installation procedures. Also the build-disk command should use the same code, not something for now, but having it separated is tech debt, our move should keep that perspective or potentially reusing it for `build-iso`.

pkg/utils/grub.go
```diff
@@ -120,13 +110,137 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool)
...
+		}
+
+		err = MkdirAll(g.config.Fs, filepath.Join(cnst.EfiDir, fmt.Sprintf("EFI/%s/", bootloaderName)), cnst.DirPerm)
+		if err != nil {
+			g.config.Logger.Errorf("Error creating dirs: %s", err)
+			return err
+		}
+
+		// Copy needed files for efi boot
+		system, err := IdentifySourceSystem(rootDir)
```
Itxaka: @davidcassany here starts the shitty code to support the 3 existing flavors for the efi files :crying_cat_face:

pkg/utils/grub.go
Itxaka: Well, we have a request for secureboot support while we have zero requests for device images, so that for me makes the case :) .

In any case, the code should be adaptable to that case when (if?) we reach it. Simply check the target type and skip if needed. We still need an EFI partiton in that case. We can even adapt this to also copy the grub.efi into the /EFI/bootx64.efi so its still in there as fallback for cloud/isos, but no idea why we should worry now about that, the workflow for efi is simple enough as to be modified in the future (copy files, add boot entry)

Itxaka: oh, I get what you mean now, you want to not use efibootmanager AND drop everything into the /EFI/ dir so it **always** uses the fallback?

I mean...sure we can do that. I would not do it, as nobody does it, but yeah, we can do it. We still need the shim in there and the grub in there but yeah, that would work for all types of media. But I wonder, if nobody does it it must be for a reason...maybe some efi systems have an integrated fallback. Maybe the system already has entries as a leftover from other older installs. Maybe some systems wont allow booting into fallback. There has to be a reason for it and for installs not ever doing it...

Itxaka:
ok, updated. This should work for all systems as long as they have grub2+shim installed, no matter what as it searches in the recently installed system (rootfs) on the iso for the required modules needed (squash4+loopback) and the required efi files (shim, grub, mokmanager)

Then it drops them in the fallback/rescue place so it alwasy boot from there.

That is still an issue IMHO, but Im gonna add an extra commit that adds the bootmanager entry and allow setting that via a flag to the installer so it can be set via command line. Set as false by default to maintain the same behavior as before although I would think that we should make it the default so we create a proper entry in the boot manager and it boots from there on the next step.

davidcassany:
> That is still an issue IMHO, but Im gonna add an extra commit that adds the bootmanager entry and allow setting that via a flag to the installer so it can be set via command line. Set as false by default to maintain the same behavior as before although I would think that we should make it the default so we create a proper entry in the boot manager and it boots from there on the next step.

I agree that `elemental install` should set the bootmanager entry by default. I was thinking in some sort of `--no-nvram` flag, I think grub2 and shim-install have such a flag, so we also align with other bootloader utilities having similar defaults. Probably there is no need to maintain previous behavior by default, I'd say it is good enough if we can achieve it via flag.

pkg/utils/grub.go
```diff
@@ -40,74 +41,63 @@ func NewGrub(config *v1.Config) *Grub {
 }
 
 // Install installs grub into the device, copy the config file and add any extra TTY to grub
-func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool) (err error) { // nolint:gocyclo
+func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool, stateLabel string) (err error) { // nolint:gocyclo
 	var grubargs []string
 	var grubdir, finalContent string
+	// only install grub on non-efi systems
+	if !efi {
```
davidcassany, line 48: This is a behavior change, the result of this install will not be hybrid right? I mean, it will not be bootable on BIOS if EFI is on.

So options are:
1. only BIOS
2. only EFI
3. BIOS and EFI 

I think we were supporting 1 and 3 before. Any thoughts about what should we support? probably omitting 3 is an interesting option if we drop BIOS support some day.

pkg/utils/common.go
```diff
@@ -514,3 +516,69 @@ func CalcFileChecksum(fs v1.FS, fileName string) (string, error) {
 
 	return fmt.Sprintf("%x", h.Sum(nil)), nil
 }
+
+// IdentifySourceSystem tries to find the os-release file in a given dir and identify the system based on the data in there
+func IdentifySourceSystem(vfs v1.FS, path string) (string, error) {
```
davidcassany, line 540: @Itxaka An idea I had now, do you think it would make sense to use this logic to determine to use one interface or another. I am thinking about having some sort of `green`, `blue`, `orange` interface and use a similar logic to this to choose which one to use, so we simply hardcode paths and names in each specific implementation and it doesn't prevent us to imagine some sort of _generic_ procedure in case distro is not detected. Also to isolate the distro specific portions of the code at the time we make them easier to extend and modify independently.

In any case I'd encompass that probably in a follow up PR. I have also been thinking that probably we need a bootloader package, so instead of having the code in `utils` se a `bootloader` package and include there disk and live use cases.

davidcassany (approved):
Looks good to me. Just wondering if we should support EFI & BIOS boot on the same installation, any concern on allowing only EFI or BIOS installs but not both?

pkg/utils/grub.go
Itxaka: I think 1 and 2 are the way to go. If you got UEFI, you probably also can boot on bios, but there is no reason to do so. And if you got bios, its probably because you dont support UEFI. Keeping them separate rather than having an hybrid approach is much better, separation of concerns, yada yada yada. Dont want to mess with those efi things that fallback into bios and then we have a mess in our hands trying to know where the boot came from.

pkg/utils/common.go
Itxaka: yep, that sounds good IMO. Having a separate interface with their own paths and such its much nicer. Then we can just access interface.GrubEfiPath and that would always point to the proper place. :+1: 

I hate seeing the different paths in here.

PR #319, opened 2022-09-19
Add an OsConfig struct to track differences in base systems

Itxaka:
Currently we have no single struct to manage any differences between base systems when running elemental-cli. This produced some weird code in order to support different names or paths for some files like the shim efi files.

This patch introduces an OsSystem struct which can house paths or names for fiels across systems so we can access those values independtly of the base system and not complicate our code much more.

Currently supporting the different shim names for EFI systems.

This patch also introduces some extra constants to use under the grub package in order to centralize those paths and names.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -127,6 +127,12 @@ const (
 	Fedora = "fedora"
 	Ubuntu = "ubuntu"
 	Suse   = "suse"
+
+	ShimFallbackx86_64 = "bootx64.efi"                                                                                           // Name of the default fallback efi name on x86_64 systems
+	ShimFallbackarm64  = "bootaa64.efi"                                                                                          // Name of the default fallback efi name on aarch64 systems
+	EfiFallbackGrub    = "search --no-floppy --label --set=root %s\nset prefix=($root)/grub2\nconfigfile ($root)/grub2/grub.cfg" // default grub template for efi grub. Chainloads the grub.cfg under the state label partition
+	EfiGrubPath        = EfiDir + "/EFI/boot/"                                                                                   // Efi path for installs
+	EfiGrubPathCfg     = EfiGrubPath + "grub.cfg"                                                                                // Efi grub cfg for installs
 )
 
 func GetCloudInitPaths() []string {
@@ -141,7 +147,7 @@ func GetDefaultSquashfsOptions() []string {
 func GetDefaultSquashfsCompressionOptions() []string {
 	options := []string{"-comp", "xz", "-Xbcj"}
 	// Set the filter based on arch for best compression results
-	if runtime.GOARCH == "arm64" {
+	if runtime.GOARCH == ArchArm64 {
 		options = append(options, "arm")
 	} else {
 		options = append(options, "x86")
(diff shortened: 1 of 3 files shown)


pkg/utils/grub.go
```diff
@@ -150,59 +150,34 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
...
+		for _, f := range osConfig.ShimFiles() {
 			_ = WalkDirFs(g.config.Fs, rootDir, func(path string, d fs.DirEntry, err error) error {
 				if err != nil {
 					return err
 				}
 
 				if d.Name() == f {
-					// On suse systems check if the path contains the proper arch
-					if system == cnst.Suse && !strings.Contains(path, g.config.Arch) {
+					// Unfortunately on Suse the shim files are named the same no matter the arch
```
davidcassany, line 173: Is it actually needed? Shouldn't we assume the first match is sane enough? e.g. if one includes shim binaries for x86_64 and arm64 I bet this is an error, I can see the use case of having both.

Also I can't understand why is it returning nil instead of error if arch is not matching.

pkg/constants/constants.go
```diff
@@ -257,3 +263,63 @@ func GetDiskKeyEnvMap() map[string]string {
 	// None for the time being
 	return map[string]string{}
 }
+
+// OsConfig represents a configuration values for a system
+// So we can use the same methods on different base OS and have those keys point to the proper values
+type OsConfig struct {
```
davidcassany, line 269: What about making it an interface? I think it would be better this way we can handle distro specific logic too, if needed, not only static values.

pkg/constants/constants.go
```diff
@@ -127,6 +127,12 @@ const (
 	Fedora = "fedora"
 	Ubuntu = "ubuntu"
 	Suse   = "suse"
+
+	ShimFallbackx86_64 = "bootx64.efi"                                                                                           // Name of the default fallback efi name on x86_64 systems
```
davidcassany, line 131: What about keeping those constants private into bootloader package? So we do not expose those as part of the API.

davidcassany (commented):
I like the overwall concept and its logic, however I was wondering if it wouldn't be cleaner trying to move all grub logic into a `bootloader` package and then there include internal private constants and create a per flavor interface. In that case then I would also move the logic in current `live` package into the new package and extend the interfaces to also cover the ISO use cases (and `build-disk` cases in a yet another round). So all the logic and ugliness of managing grub is in a single package with a very simple and limited API to serve our use cases.

What do you think? I like the idea of isolating as much as possible all the bootloader code into its own package.

pkg/utils/grub.go
Itxaka: Yes, could be. I think I'm biased due to our iso/raw building  or my tests with weird custom rootfs.

pkg/constants/constants.go
Itxaka: Sure. I don't really expect this to grew that much as to need an interface no? In any case no hard feelings one way or the other

pkg/constants/constants.go
Itxaka: Yep, no problem there.

Itxaka: > What do you think? I like the idea of isolating as much as possible all the bootloader code into its own package.

Yeah, makes sense if we want to expand this.

Itxaka:
Closing this as it will be reworked into its own bootloader package

PR #321, opened 2022-09-19
Fix disk partitioning

davidcassany:


davidcassany:
Otherwise a nil pointer was added in case no partition is defined with size `0`. 

This is needed to fix rancher/elemental-toolkit#1607

Itxaka (approved):
Good catch, I knew there had to be issues :)

PR #322, opened 2022-09-21
Create ISO for EFI only or BIOS only, not both

davidcassany:
This commit adds a firmware flag in build-iso to choose BIOS or EFI builds. ISOs supporting both are not longer possible.

'firmware' is added as deprecated since BIOS support is deprecated leaving EFI as the only option.

Firmware is set to EFI by default

### DIFF
--- a/pkg/live/common.go
+++ b/pkg/live/common.go
@@ -18,22 +18,22 @@ package live
 
 import (
 	"fmt"
-	"path/filepath"
 
 	"github.com/rancher/elemental-cli/pkg/constants"
+	v1 "github.com/rancher/elemental-cli/pkg/types/v1"
 )
 
 const (
-	isoEFIPath            = "/boot/uefi.img"
 	efiBootPath           = "/EFI/BOOT"
-	isoLoaderPath         = "/boot/x86_64/loader"
+	isoLoaderPathX86      = "/boot/x86_64/loader"
+	isoLoaderPathArm64    = "/boot/arm64/loader"
 	grubArm64Path         = grubPrefixDir + "/arm64-efi"
 	grubEfiImagex86Dest   = efiBootPath + "/bootx64.efi"
 	grubEfiImageArm64Dest = efiBootPath + "/bootaa64.efi"
 	grubCfg               = "grub.cfg"
 	grubPrefixDir         = "/boot/grub2"
 
-	// TODO document any custom bootloader must match this setup as these are not configurable
+	// TODO document any custom BIOS bootloader must match this setup as these are not configurable
 	// and coupled with the xorriso call
 	isoHybridMBR   = "/boot/x86_64/loader/boot_hybrid.img"
 	isoBootCatalog = "/boot/x86_64/boot.catalog"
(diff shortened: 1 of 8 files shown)


davidcassany:
@Itxaka @kkaempf this will create pure UEFI ISOs by default. However ISOs for BIOS are still possible by using the flag `--firmware bios`. BIOS ISOs can't be built in non privileged containers. BIOS/UEFI ISOs are no longer possible.

I believe this is the sane way of fading out BIOS support, having completely separate paths and not considering support for both.

pkg/live/common.go
```diff
@@ -81,23 +82,35 @@ const (
...
-		"-append_partition", "2", "0xef", filepath.Join(root, isoEFIPath),
-		"-boot_image", "any", "next",
-		"-boot_image", "any", "efi_path=--interval:appended_partition_2:all::",
-		"-boot_image", "any", "platform_id=0xef",
-		"-boot_image", "any", "emul_type=no_emulation",
+func XorrisoBooloaderArgs(root, efiImg, firmware string) []string {
+	switch firmware {
+	case v1.EFI:
+		args := []string{
+			"-append_partition", "2", "0xef", efiImg,
```
Itxaka, line 83: do we still need all this flags for uefi? AFAIK in my tests a simple `xorriso -as mkisofs -V 'MY_ISO' -e EFI/BOOT/efiboot.img -no-emul-boot -o OUTPUT_ISO isoroot/` worked to create a uefi bootable iso?

Itxaka: where efiboot.img is the `efiImg`  and the rest of the parameters are already known. Im not sure we need all this, we could make this more straightforward.

pkg/live/green.go
```diff
@@ -87,37 +86,37 @@ func (g *GreenLiveBootLoader) PrepareISO(rootDir, imageDir string) error {
...
+			}
 		}
-		fontsDir := filepath.Join(loaderDir, "/grub2/fonts")
-		err = utils.MkdirAll(g.buildCfg.Fs, fontsDir, constants.DirPerm)
-		if err != nil {
-			return err
-		}
-		err = utils.CopyFile(g.buildCfg.Fs, filepath.Join(rootDir, grubFont), fontsDir)
+	case constants.ArchArm64:
+		err = g.copyGrubFonts(rootDir, imageDir, isoLoaderPathArm64)
```
Itxaka: any idea why we need to copy the fonts only on arm64? arent they bundled with the grub binary?

Itxaka: ah this is for both, sorry it was hidden by github.

pkg/live/common.go
davidcassany: I am not sure the result would still be bootable as disk, including the partition table and EFI fat image. Also note that `-as mkisofs` is just a kind of a preset in xorriso. It attempts to emulate mkisofs flags and convert it into xorriso params. I just opted to go straight to the use of `xorriso`.

pkg/live/green.go
```diff
@@ -87,37 +86,37 @@ func (g *GreenLiveBootLoader) PrepareISO(rootDir, imageDir string) error {
 
 	switch g.buildCfg.Arch {
 	case constants.ArchAmd64, constants.Archx86:
-		// Create eltorito image
-		eltorito, err := g.BuildEltoritoImg(rootDir)
+		err = g.copyGrubFonts(rootDir, imageDir, isoLoaderPathX86)
```
Itxaka: I dont think we need fonts for the grub.efi image.

pkg/live/green.go
davidcassany: This is the same we do for x86_64 on EFI. We need the fonts to have a proper view of the grub menu, otherwise some weird artifacts appear in the screen.

pkg/live/green.go
davidcassany: Go give it a try, but I think we do. I do not remember exactly in which case we needed these fonts, probably it was only for BIOS, worth a try. Gonna check.

pkg/live/green.go
```diff
@@ -132,8 +131,29 @@ func (g *GreenLiveBootLoader) PrepareISO(rootDir, imageDir string) error {
 		return err
 	}
 
-	// Include EFI contents in iso root too
-	return g.PrepareEFI(rootDir, imageDir)
+	if g.spec.Firmware == v1.EFI {
+		// Include EFI contents in iso root too
```
Itxaka, line 126: Is this really needed? If we create an ESP partition on the iso it should boot from there? what are those files in the iso for?

pkg/live/common.go
Itxaka: Sure, but Im saying that for the following flags needed for a UEFI boot iso, under `mkisofs` they probabyl dont translate to those 100 flags under `xorriso` no? And if the do...do we really care? Make its much simpler...

Itxaka: anyway, something we can fix down the line, Im just scared of this 100 flags which I have no idea what they do and checking the xorriso man pages...its daunting.

Itxaka (approved):
awaiting the font stuff, but again that can be dropped later. lgtm

Itxaka:
We musnt forget to add the shim on the iso efi generation once this gets in!

pkg/live/green.go
davidcassany: With the current setup this is required to boot as ISO, as a cdrom. To boot as a block storage it reads the EFI partition. Without I can only boot as disk reading the ESP partition.

davidcassany:
Right just removed the fonts copy, there were some grub.cfg parts loading it. Apparently removing all this does not affect the result, I guess this is already the default.

In addition I also removed some xorriso parameters that were already set to the default value.

pkg/live/green.go
davidcassany: Your were right, just removed it.

PR #323, opened 2022-09-21
Fix image tests to use an up to date system for testing the install

Itxaka:
Latest versions were not being pushed to quay.io due to the old cos systems not supporting the new grub artifacts


davidcassany (approved):
👍🏽

PR #324, opened 2022-09-21
Keep working dir after a chroot call

davidcassany:

### DIFF
--- a/pkg/utils/chroot.go
+++ b/pkg/utils/chroot.go
@@ -131,17 +131,34 @@ func (c *Chroot) Close() error {
 
 // RunCallback runs the given callback in a chroot environment
 func (c *Chroot) RunCallback(callback func() error) (err error) {
-	// Store current root
-	oldRootF, err := os.Open("/") // Can't use afero here because doesn't support chdir done below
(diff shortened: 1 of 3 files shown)


davidcassany:
@Itxaka this fixes the need of using absolute paths build-iso calls, we were not keeping the working directory after undoing the chroot.

pkg/utils/chroot.go
```diff
@@ -131,17 +131,34 @@ func (c *Chroot) Close() error {
...
+			err = tmpErr
+		}
+	}()
+
 	// Store current root
-	oldRootF, err := os.Open("/") // Can't use afero here because doesn't support chdir done below
+	oldRootF, err = c.config.Fs.Open("/")
 	if err != nil {
-		c.config.Logger.Errorf("Cant open /")
+		c.config.Logger.Errorf("Can't open current rooy")
```
Itxaka: Rooy? 🤣

Itxaka (approved):

pkg/utils/chroot.go
davidcassany: 😱

PR #326, opened 2022-09-22
Add CreateBootEntry to EFI install/reset

Itxaka:
We now have a flag on install/reset commands to create an entry in the
EFI bootmanager pointing to the signed shim.efi that we copy on
grub.Install

This allows to have the boot manager create the entry directly instead of
relaying in the fallback mechanism to boot an elemental install.

The behaviour is unchanged because the option defaults to false


Itxaka:
@davidcassany what do you think? Currently defaults to false but I would really like to enable it by default

Itxaka:
Also, vendoring sucks, jsut look at commit [`823c9a8` (#326)](https://github.com/rancher/elemental-cli/pull/326/commits/823c9a8fe28da22a1619c0928d80a9e9b073afcb) which has the actual change

pkg/utils/grub.go
```diff
@@ -254,11 +251,42 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
...
+			}
+			g.config.Logger.Info("Entry created for elemental-shim in the EFI boot manager")
+		}
 	}
 
 	return nil
 }
 
+func CreateBootEntry(shimName string, relativeTo string, efiVariables efibootmgr.EFIVariables) error {
+	bm, err := efibootmgr.NewBootManagerForVariables(efiVariables)
```
davidcassany, line 325: Is this pure go-lang? i guess it is not, does it require `efibootmgr` package in the host?

davidcassany (dismissed):
Looks good to me :+1:

> Currently defaults to false but I would really like to enable it by default

I fully agree. Probably we should name the flag different then, something like `--no-boot-entry`?

pkg/utils/grub.go
Itxaka: pure golang babyyyyy

in modern linux systems, efivars are always mounted on the system, and they are just files with some data on them, kind of like the /sys/dev/block stuff, so you dont need any stinking programs to access them. You could even manually manage them if you understood the data

`efivarfs on /sys/firmware/efi/efivars type efivarfs (rw,nosuid,nodev,noexec,relatime)`

Itxaka:
Updated, now its enabled by default, the flag disables it. I wonder if we should set the mutually exclusive values with the firmware to check that its on efi? I mean, nothing will happen if the flag is enabled but we are on iso so.... not really an issue, but UI/UX wise makes sense.

Itxaka:
@davidcassany dismissing the review as I changed a few things while testing this.

 - Now we also copy the artifacts into /EFI/elemental like good citizens
 - Entry in the bootmanager points to the path above, as it should be. Fallback its still in place so that behaviour should be the same
 - On install there is a cleaning done to remove any entries on the boot manager that are called the same. This is due to the entries on bootmanager referencing the partition UUID, which on an install changes, as the partitions are recreated. This only happens if you reinstall the same machine several times. Expected to be much improved in the future, but currently a rough approach is done to not leave broken entries around.

Itxaka:
This has been tested with a clean install and no issues found. Also with several reinstalls and no issues found.

davidcassany (approved):
Nice!

Itxaka:
Also tested on reset, works as expected

Itxaka:
cc @mudler some changes here that will reverb to kairos.

Unfortunately entry cant be named by now so you would get an `elemental-shim`entry name. Can opt out of this by passing the `--no--not-entry` flag to the install/reset. Hopefully this cna be configured soon-ish

PR #327, opened 2022-09-22
Add secure boot for EFI ISO images

davidcassany:

### DIFF
--- a/pkg/live/common.go
+++ b/pkg/live/common.go
@@ -24,12 +24,12 @@ import (
 )
 
 const (
-	efiBootPath           = "/EFI/BOOT"
-	grubEfiImagex86Dest   = efiBootPath + "/bootx64.efi"
-	grubEfiImageArm64Dest = efiBootPath + "/bootaa64.efi"
-	grubCfg               = "grub.cfg"
-	grubPrefixDir         = "/boot/grub2"
-	isoBootCatalog        = "/boot/boot.catalog"
+	efiBootPath    = "/EFI/BOOT"
+	efiImgX86      = "bootx64.efi"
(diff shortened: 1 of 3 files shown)


kkaempf (approved):
lgtm

pkg/live/common.go
```diff
@@ -24,12 +24,12 @@ import (
...
 
 const (
-	efiBootPath           = "/EFI/BOOT"
-	grubEfiImagex86Dest   = efiBootPath + "/bootx64.efi"
-	grubEfiImageArm64Dest = efiBootPath + "/bootaa64.efi"
-	grubCfg               = "grub.cfg"
-	grubPrefixDir         = "/boot/grub2"
-	isoBootCatalog        = "/boot/boot.catalog"
+	efiBootPath    = "/EFI/BOOT"
+	efiImgX86      = "bootx64.efi"
```
Itxaka, line 28: should we reuse the generic constants to share the same filename across here and the utils.grub?

Itxaka (approved):
looking good!

pkg/live/common.go
davidcassany: Good point! What I'd really like is to have all the bootloader code in a `pkg/bootloader` so move this `pkg/live` to `pkg/bootloader` and include all the grub related code from `pkg/utils` in there too.

So we can keep most of this logic isolated and some of the logic and constants private, so make the bootloader API minimal to serve our needs.

Just as a measure to ensure all the this unpleasant logic related to the bootloader stays silent and quite in a single place 😅 

But I'd do that once we have the new logic settled in a specific PR.

PR #330, opened 2022-09-23
Rework SetDefaultGrubEntry

### LINKED ISSUE #329: On reset, grub entry name is lost
Using: latest main
Example system based on Elemental Teal latest main https://github.com/rancher/elemental

## How to reproduce

 - Get an elemental iso
 - Install the system
 - Reboot
 - **You will see the `Elemental` entry in grub
 - Boot to recovery
 - Run `elemental reset`
 - Reboot
 - Grub entries have reverted to `cOS`

## What was expected

 - Entries in grub to keep the same name as before

## Extra information

 - Only seems like the entries were affected, the rest of the system is fine
 - /etc/os-release still contains the proper GRUB_ENTRY_NAME

Just had a look at SetDefaultGrubEntry and it seems the workflow is wrong. If a `defaultEntry` is passed it will always use that but we want to get the entry from the os-release first and then fall back into the default :/

Itxaka:
The current workflow is wrong as it would skip looking for an entry name in the os-release file and alwasy use the defaultEntry

Now we first check to see if there is an os-release entry and fallback into the defaultEntry otherwise

Fixes #329


fgiudici (approved):
looks good here!

PR #331, opened 2022-09-23
Remove sanity check for SUSE based systems on efi installation

davidcassany:
Shim images are under /usr/share/efi/aarch64/shim.efi in arm64 hence the architecture string is not included within the path. Same happens with the grub.efi image.

### DIFF
--- a/pkg/utils/grub.go
+++ b/pkg/utils/grub.go
@@ -208,11 +208,6 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool, 
 				}
 
 				if d.Name() == f {
-					// On suse systems check if the path contains the proper arch
-					if system == cnst.Suse && !strings.Contains(path, g.config.Arch) {
-						return nil
-					}
-
 					fileContent, err := g.config.Fs.ReadFile(path)
 					if err != nil {
 						return fmt.Errorf("error reading %s: %s", path, err)


davidcassany:
@Itxaka I am not sure why this sanity check was here and if it is necessary, I guess we might  need to constrain a bit more the search if other unrelated files are called the same. Probably we should look only under `/usr` and ensure the path includes `shim` or `grub` rather the arch.

pkg/utils/grub.go
```diff
@@ -208,11 +208,6 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
 				}
 
 				if d.Name() == f {
-					// On suse systems check if the path contains the proper arch
-					if system == cnst.Suse && !strings.Contains(path, g.config.Arch) {
-						return nil
-					}
-
 					fileContent, err := g.config.Fs.ReadFile(path)
```
davidcassany, line 211: by the way just noted this reads and writes the whole file, we should better use the `utils.CopyFile(fs v1.FS, source string, target string)` as this prevents loading the whole file to memory. I guess not really relevant for this specific use case, but still.

Itxaka:
>@Itxaka I am not sure why this sanity check was here and if it is necessary, I guess we might need to constrain a bit more the search if other unrelated files are called the same. Probably we should look only under /usr and ensure the path includes shim or grub rather the arch.

This check was in there because somehow I ended up with 2 different arch packages in my iso rootfs somehow??

It should be impossible to have both of them installed at the same time I think, shim package is arch dependant...wtf.

Yeah looking good :+1:

Itxaka (approved):

pkg/utils/grub.go
Itxaka: Good point. The problem with this is that the `CopyFile` is not under the fs file so I never found about it lol

fgiudici (approved):
lgtm

Itxaka:
@davidcassany there is another check for the modules but I think that one has to stay, as you can install cross-arch modules cant you? Maybe we will need to transform constants.ArchArm64 into "aarch64" for that check

Itxaka:
ah no, that one passes correctly due to the proper arch:

```
$ ls /usr/share/grub2         
arm64-efi  ascii.pf2  euro.pf2  grub-mkconfig_lib  i386-pc  themes  unicode.pf2  x86_64-efi
```

Itxaka:
no need to update branch, changes merged were the docs update

PR #332, opened 2022-09-23
Stop testing on image push/release

Itxaka:
Gating should happen at the PR level no here. Otherwise we are not pushing the images and have no visibility if they failed to push due to broken tests


davidcassany (approved):
lgtm

PR #334, opened 2022-09-23
Drop binary signature for now

Itxaka:


Itxaka:
Merging directly, is blocking the release

PR #335, opened 2022-09-23
Add binary signatures

Itxaka:
Seems like a bad mix of an older cosign + newer goreleaser was causing the issues. Tested in a different repo and all artifacts were properly signed with no issues


fgiudici (approved):
This basically reverts #334 and fixes the issue that forced to drop cosign signatures.
Looks good here!

PR #336, opened 2022-09-28
Do not add partitions in EFI ISOs

davidcassany:
There is no need of adding partition headers in EFI ISOs as they can still be dded to a USB and boot from there.

related to rancher/elemental#365

### DIFF
--- a/pkg/action/build-iso.go
+++ b/pkg/action/build-iso.go
@@ -140,22 +140,14 @@ func (b *BuildISOAction) ISORun() (err error) {
 		return err
 	}
 
-	err = b.prepareISORoot(isoDir, rootDir)
+	err = b.prepareISORoot(isoDir, rootDir, uefiDir)
 	if err != nil {
 		b.cfg.Logger.Errorf("Failed preparing ISO's root tree: %v", err)
 		return err
 	}
 
-	if b.spec.Firmware == v1.EFI {
-		b.cfg.Logger.Info("Creating EFI image...")
-		err = b.createEFI(uefiDir, filepath.Join(isoTmpDir, constants.IsoEFIImg))
-		if err != nil {
-			return err
-		}
-	}
-
 	b.cfg.Logger.Infof("Creating ISO image...")
-	err = b.burnISO(isoDir, filepath.Join(isoTmpDir, constants.IsoEFIImg))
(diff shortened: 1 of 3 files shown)


davidcassany:
In relation to rancher/elemental#365 the xorriso maintainer noted it is not possible to add an additional session to the iso image without breaking the partition metadata. Then I tried to remove the partition data to keep it simple and also realized I was capable to boot the iso as cdrom and as disk with qemu. I still have to experiment with some real hardware.

The relevant thing with this setup is that it is possible to add the yaml registration file into the iso without having to rebuild it. That means we could eventually release an ISO and then call

```
xorriso -boot_image any keep -dev <path_to_released_iso> -map <path_to_registration_yaml> /livecd-cloud-config.yaml
```

and in less than a second the ISO is ready (requires xorriso >= v1.5, Leap ships v1.4.6... 😞 ).

I believe this is valuable if we ever consider adding the ISO _build_ into the elemental-operator.

agracey:
This would be amazing!

scdbackup:
Hi,

as i wrote in
  https://github.com/rancher/elemental/issues/365#issuecomment-1261852353
this change will probably keep some EFI implementations from booting the
ISO from USB stick or other hard-disk-like storage devices.

The remaining El Torito EFI boot image is supposed only to work from
optical media (including virtual ones).
OVMF is more tolerant in this aspect than the specs and than other
real iron EFI implementations allow.

Have a nice day :)

Thomas

davidcassany:
> Hi,
> 
> as i wrote in [rancher/elemental#365 (comment)](https://github.com/rancher/elemental/issues/365#issuecomment-1261852353) this change will probably keep some EFI implementations from booting the ISO from USB stick or other hard-disk-like storage devices.
> 
> The remaining El Torito EFI boot image is supposed only to work from optical media (including virtual ones). OVMF is more tolerant in this aspect than the specs and than other real iron EFI implementations allow.
> 
> Have a nice day :)
> 
> Thomas

Thanks for the insights and the recommendation. This is all valuable information.

davidcassany:
Closing since there is no need to adapt anything. The changes here could lead, for instance, to unbootable USB sticks.

In fact the real issue on trying to modify or customize the build by elemental-cli is using an up to date xorriso.

PR #339, opened 2022-10-15
Fill persistent mountpoint on upgrade

Itxaka:
Upgrade tries to mount the persistent partition during the upgrade in order to use it as tmp dir to store and decompress images. Unfortunately as the mountpoint is empty, this fails and it always falls back to the /tmp dir which is a tmpfs dir with 25% of the RAM as space. This means that in a 4Gb device we get a /tmp dir of 1Gb, which is not enough to download and decompress a single image.

This used to work before so we probably introduced this regression with some changes and never notice due our workers having 10Gb of RAM

This patch fills the persistent mountpoint on upgrade config load so we can temp mount it during upgrade and maintains the same fallback

Signed-off-by: itxaka <igarcia@suse.com>

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -164,9 +164,7 @@ func (u *UpgradeAction) Run() (err error) {
 			tmpdir := utils.GetTempDir(&u.config.Config, "")
 			u.config.Luet.SetTempDir(tmpdir)
 			// Remove the tmpdir before unmounting
-			cleanup.Push(func() error {
-				return u.config.Fs.RemoveAll(filepath.Join(persistentPart.MountPoint, "tmp"))
-			})
+			cleanup.Push(func() error { return u.config.Fs.RemoveAll(tmpdir) })
 			cleanup.Push(umount)
 		}
 	}
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -150,15 +150,21 @@ func (u *UpgradeAction) Run() (err error) {
 	// Cleanup transition image file before leaving
 	cleanup.Push(func() error { return u.remove(upgradeImg.File) })
 
-	// Recovery does not mount persistent, so try to mount it. Ignore errors, as its not mandatory.
+	// Recovery does not mount persistent, so try to mount it. Ignore errors, as it's not mandatory.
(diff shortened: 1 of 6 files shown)


pkg/action/upgrade.go
```diff
@@ -150,15 +150,24 @@ func (u *UpgradeAction) Run() (err error) {
...
-			u.config.Logger.Warn("could not mount persistent partition")
+			u.config.Logger.Warn("could not mount persistent partition: %s", err.Error())
 		} else {
-			cleanup.Push(func() error { return e.UnmountPartition(persistentPart) })
+			// Set luet tempdir
+			tmpdir := utils.GetTempDir(&u.config.Config, "")
+			u.config.Luet.SetTempDir(tmpdir)
+			// Remove the tmpdir before unmounting
+			cleanup.Push(func() error {
+				return u.config.Fs.RemoveAll(filepath.Join(persistentPart.MountPoint, "tmp"))
```
kkaempf: This looks inconsistent. We set `tmpdir` in line 165 but cleanup `"tmp"` here 🤔

kkaempf (approved):
👍🏻 esp. on the much improved logging

pkg/action/upgrade.go
davidcassany: Inded this "tmp" literal string here is confusing. I believe we should have a constants for `tmp` and make use of it here and in `utils.GetTempDir`, as this code is dependent on `utils.GetTempDir`.

davidcassany (approved):
Looks good to me, just one comment on the use of `tmp` literal, it should be a constants shared with the relevant utils method, as both pieces of code are coupled.

Also it is missing the mock implementation for the luet interface, that's why tests are failing.

Itxaka:
yep, sorry about this, I sent this yesterday in a hurry as its part of fixing the https://github.com/rancher/elemental-toolkit/pull/1624 and didnt really work it trougth properly, just did it as a workaround to see if it fixed the issue on the toolkit during the upgrade tests. 

Let me rework this and fix the tests!

pkg/action/upgrade.go
Itxaka: well its kind of difficult to have a constant as it can have different base dirs, so just a `/tmp` constant wont work as it can be under persistent for example.

PR #341, opened 2022-10-17
Align install, reset and upgrade hooks

davidcassany:
This PR puts the hooks for the `upgrade`, `install` and `reset` under the same criteria.

* `before`, right before applying the OS image, but after the partitions are set and mounted. This allows to preload data to partitions if needed (e.g. drop some driver binaries into the persistent partition)
* `after-chroot`, right after applying the OS image and the OS image is still RW mounted.
* `after`, right after setting all images (active, passive and/or recovery) and the deployed image is no longer mounted

Related to rancher/elemental#334


davidcassany:
This allows to copy some data into the persistent storage before running upgrade, install or reset, so that this data can eventually be used in a after chroot hook (e.g. install some drivers that where present in installation media).

Itxaka (approved):

PR #349, opened 2022-10-31
upgrade: Skip the device which only has the same label name

### LINKED ISSUE #348: If more than one partition has a COS_* label, upgrade wont work
Reported to harvester folks: https://github.com/harvester/harvester/issues/3070

FC storage without multipath, so the disk is seen 4 times.

This seems to confuse elemental-cli as it will just try to mount the fist found partition for COS_STATE. We should make sure that we mount the ACTIVE COS_STATE (On upgrade it should be mounted RO and we remount it RW)

Patch that fixes it: https://github.com/Vicente-Cheng/elemental-cli/commit/5ccfbbc22f3870ab667408a9377af1649f770f8d

This patch could be backported to an older version to unblock the harvester upgrade path.

Harvester is using v0.0.14+gedc412

Itxaka:
- when upgrading, we need find the related label name and
      upgrade with it. If we have many devices have the same
      label name, it would cause failure. To avoid this case,
      we need add more checking about the target disk should
      be the active cOS disk.

Fixes  #348 

Signed-off-by: Vicente Cheng <vicente.cheng@suse.com>

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -167,7 +167,7 @@ func (u *UpgradeAction) Run() (err error) { // nolint:gocyclo
 			upgradeStateDir = constants.RecoveryDir
 		} else { // We are updating active
 			// Remount state partition RW
-			statePart, _ := utils.GetFullDeviceByLabel(u.Config.Runner, u.Config.StateLabel, 2)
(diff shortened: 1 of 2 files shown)


Itxaka:
closing in favour of https://github.com/rancher/elemental-cli/pull/350

PR #352, opened 2022-11-02
Prioritize mounted device if multiple devices match label

davidcassany:
Related to #348


davidcassany:
This is how I would solve #348 in our current main. This affects the logic of getting devices for upgrade and reset and fetches first mounted device matching label. If none is mounted last device matching label wins.

Itxaka (approved):

PR #354, opened 2022-11-07
[POC] Exit codes

Itxaka:
This is an example on how to tag and propagate the exit codes from different parts of the code into the final cmd and how to exit properly from different parts of the code.

 - Introduces an enhanced Error that also brings the code inside it so we can propagate the errors along the code for functions that support it, i.e. if CreateImage has several parts where it could fail we have to return an ElementalError so we can cleanup and then os.exit with the proper error code.
 - introduces exit_codes list under the constants dir. Did not bother to make specific numbers associated with them and just used an iota. Not sure about this as when the list grows the codes migth change if we change the order, maybe it would be nicer to have some specific codes attached to specific errors and they are frozen for life?

So putting it out there for consideration. I tried several approaches with this, but unfortunately os.exit DOES NOT call any defer calls, so in our case its very important that on failure we try to cleanup as much as possible, removing temp dirs, remounting/unmounting partitions, etc... so we cannot blindly just call that. And making everything use our elementalError is a bit of pita as there is a lot of places we pass the errors from one function to other before we can really exit due to the cleanup....

Not sure if a better way is possible? Maybe we could use go channels to have a channel that runs the cleanup and exits with an exit code and we then can call that from anywhere? No idea, seems more complex.

Anyway, putting this up for comments, ideas, new approaches, etc..

Signed-off-by: itxaka <igarcia@suse.com>

### DIFF
--- a/cmd/build-disk.go
+++ b/cmd/build-disk.go
@@ -17,7 +17,8 @@ limitations under the License.
 package cmd
 
 import (
-	"fmt"
+	"github.com/rancher/elemental-cli/pkg/constants"
+	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -54,7 +55,7 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 
 			err = validateCosignFlags(cfg.Logger, flags)
 			if err != nil {
-				return err
+				os.Exit(constants.ExitCosignError)
 			}
 
 			// Set this after parsing of the flags, so it fails on parsing and prints usage properly
@@ -63,8 +64,7 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 
 			spec, err := config.ReadBuildDisk(cfg, flags)
 			if err != nil {
-				cfg.Logger.Errorf("invalid install command setup %v", err)
-				return err
+				os.Exit(constants.ExitConfigError)
 			}
 
 			// Get the spec for the arch only
@@ -89,15 +89,10 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 
 			if exists, _ := utils.Exists(cfg.Fs, output); exists {
 				cfg.Logger.Errorf("Output file %s exists, refusing to continue", output)
-				return fmt.Errorf("output file %s exists, refusing to continue", output)
+				os.Exit(constants.ExitFileExists)
 			}
 
-			err = action.BuildDiskRun(cfg, specArch, imgType, oemLabel, recoveryLabel, output)
-			if err != nil {
-				return err
-			}
-
-			return nil
+			return action.BuildDiskRun(cfg, specArch, imgType, oemLabel, recoveryLabel, output)
 		},
 	}
 	root.AddCommand(c)
--- a/pkg/constants/exit_codes.go
+++ b/pkg/constants/exit_codes.go
@@ -1,0 +1,23 @@
+package constants
+
+const (
+	ExitUnknownError = iota + 10
+	ExitCopyError
+	ExitTempdirError
+	ExitNoPackagesArchBuildDisk
+	ExitNoRepoConfiguredBuildDisk
+	ExitCreateDirError
+	ExitCreateFileError
+	ExitTruncateFileError
+	ExitFailedFileCloseError
+	ExitOpenFileError
+	ExitStatFileError
+	ExitTarError
+	ExitGzipError
+	ExitParseSourceError
+	ExitCosignError
+	ExitConfigError
+	ExitFileExists
+	ExitFailedCratingPart
+	ExitFailedRunningCmd
+)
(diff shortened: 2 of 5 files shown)


cmd/build-disk.go
```diff
@@ -54,7 +55,7 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 
 			err = validateCosignFlags(cfg.Logger, flags)
 			if err != nil {
-				return err
+				os.Exit(constants.ExitCosignError)
```
davidcassany, line 58: I know it is a PITA, but I'd probably try to keep the `os.Exit` as it was, only at the very top level and simply propagate the error up. Having many `os.Exit` in code is usually not nice to debug.

pkg/error/elementalError.go
```diff
@@ -0,0 +1,22 @@
+package error
+
+// ElementalError is our enhanced error that brings an error string + exit code
+type ElementalError struct {
```
davidcassany, line 4: I like it, probably we could also extend it to cover the use cases we have for `github.com/hashicorp/go-multierror` and probably some other (I am not sure if we use some other errors library now).

pkg/constants/exit_codes.go
```diff
@@ -0,0 +1,23 @@
+package constants
+
+const (
```
davidcassany, line 3: I am fine having some sort of `iota` for the error numbers, we should be strict about only appending errors at the bottom, probably a comment here at the top would be enough.

My only concern is that it would be really nice having a literal string human readable to make the error number understandable. So the output message in logs should include the propagated error string literal in addition to the staged error code plus a meaningful literal tied to the error code.

What if we use the error code number for some sort of an error factory?

Something like

```golang
func New(err error, code int) *ElementalError {
  switch code {
    case ExitCopy:
       return &ElementalError{
        s:    fmt.printf("CopyError %d: %s", code, err.Error()),
        code: code,
	}
    ...
    default:
      return &ElementalError{
        s:    fmt.Printf("UnknownError %d", code, err.Error()),
        code: code,
	}
    }
```

I guess the problem of this approach or similar ones having to maintain separately the error text and error number.

Another option would be creating a new type per error (with static error num), but then nothing prevents from having to different types with the same error number. 🤷🏽‍♂️

davidcassany (commented):
I like it, the only minor comment I have is that I would propagate the error until the very top as it used to be.

Also I like the idea taking the approach of getting it done cmd by cmd. I think we could limit the scope not beyond the `pkg/actions`, so start using the new error interface from `pkg/actions` up to the cmds and main. This would be easy to split into multiple independent cards to work in parallel and also the result would be quite a decent error tracking to understand at which stage failed.

Itxaka:
The main problem with this is that we cant use ginkgo as it is to test the CMD package, as we are not able to catch the os.exit cleanly from ginkgo. We would need to actually run the commands, at least for failures, to get those, which is a pain in the ass.

Probably affects non-cmd stuff :/

cmd/build-disk.go
Itxaka: yes, this would be nice but its a huge work to make it work.

It kinds of change the signate of the cobra CMD stuff so we have to create our own with overrides and crap liek that. golang is seriously not made for nice clean inheritance and not everybody uses interfaces :cry: 

I tried that already and it was terrible, let me re-try it again tonight to see if I can make it more easy on myself lol

pkg/constants/exit_codes.go
Itxaka: Well, log errors and exit codes are different no? I would rather not mix them to be honest, if you cant access the log somehow (no idea why) you can always use the simple exit code for things that retries (i.e error 123 is download error so I will auto-retry on my application) but on the log...its much more rich than that, there is not really that necessity for error codes when we have verbose errors+attached errors (i.e. download error: ssl connection failure)

cmd/build-disk.go
davidcassany: mmm I see, the limits are the cobra method signatures... I don't know how but we are facing cobra or viper limitations pretty often... 😅

pkg/constants/exit_codes.go
davidcassany: Well my point is that if we get an ExitCopy error (lets imagine it is number `13`) I'd like to quickly see this this error code means `ExitCopy`, so I don't have to download the github code to guess it. Usually I would expect this sort of stuff to appear in man pages (see `man zypper`) or similar command line accessible docs but I don't think we will be shipping such a thing. We could also list them in  `help` message, but I think it will quickly turn to be annoying if its a long list.

No strong opinion here but I'd say the user should be capable to easily relate the number with the name we gave to it to make it really useful.

Itxaka:
ok, ignore this, Im dumb or something but there is a much easire way of doing this, keeping our stuff and propagating the exit code.

pkg/constants/exit_codes.go
Itxaka: I still not convinced. Error codes are for machines mainly, while the logs are for people (Used to be, not sure anymore with all the structured log messages lol). I have only used exit codes for automating things like retries, not from an user perspective, as that has always being moref focused on the log/error messages.

I like the possibility of having the error codes printed on a command, should be pretty easy and its greapable so it could make sense to have them in a special command to print all exit codes, as we dont ship a man file.

Or we could ship a man file LOL

PR #355, opened 2022-11-07
Introduce ElementalError and exit codes

Itxaka:
- ElementalError is an enhaced error interface that brings along the exit code so we can exit properly from the main cmd command
 - Exit codes are stored along the error package
 - Build-disk command and related, but confined functions are all using the new ElementalError to propagate the exit code
 - Exit codes docs are autogenerated via the build_docs target, which is also triggered automatically on main merger and crates PRs if the docs/exit codes are not updated


Itxaka:
@davidcassany version 2, this works in parallel as well, we can go one by one, maybe starting with upgrade as it seems the one that its more likely to be automated

cmd/build-disk.go
```diff
@@ -49,12 +49,12 @@ func NewBuildDisk(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			flags := cmd.Flags()
 			cfg, err := config.ReadConfigBuild(viper.GetString("config-dir"), flags, mounter)
 			if err != nil {
-				return err
+				return eleError.NewFromError(err, eleError.ReadingBuildConfig)
```
Itxaka, line 52: Good thing here is, when(if?) we move `config.ReadConfigBuild` to return an elemental error, we can just return it directly instead of transforming it here. The we can just use the transform for things that are out of our control like external libs that just return an error interface.

cmd/root.go
```diff
@@ -47,6 +48,12 @@ var rootCmd = NewRootCmd()
 func Execute() {
 	err := rootCmd.Execute()
 	if err != nil {
-		os.Exit(1)
+		switch t := err.(type) {
+		case *eleError.ElementalError:
+			os.Exit(t.ExitCode())
+		default:
```
davidcassany, line 54: would it make sense adding a warning here? Mostly for us so whenever we see some traces going this path we can evaluate if some specific error code was missing or not. Not so sure we would make much use of it.

cmd/build-disk.go
davidcassany: Yes I see, this is good :+1:

davidcassany (approved):
I like it, I am just wondering how to document or make user aware of the different error codes.

cmd/root.go
Itxaka: I guess we would see on the logs the error that resulted in this no? Then we could follow it up?

Itxaka:
> I like it, I am just wondering how to document or make user aware of the different error codes.

Yes, this is a pain in the ass. Not even sure if it would be possible to autogenerate them and put them under docs somehow.

Itxaka:
@davidcassany How about this? Adding comments to the constants should result in the go packages doc site creating the proper docs for the constants: https://pkg.go.dev/github.com/rancher/elemental-cli/cmd#pkg-constants so we could link to there. That should be automatic and the only thing needed would be to add comments to any new constants.

frelon:
Looks nice! I would name the module _pkg/error_ -> _pkg/errors_ so it's possible to import without aliasing and use consistent separators for the new files instead of camelCase :+1:

Itxaka:
> Looks nice! I would name the module _pkg/error_ -> _pkg/errors_ so it's possible to import without aliasing and use consistent separators for the new files instead of camelCase +1

welp, why didnt I think of that, I kept suffering on how to name the import properly as to not collide  ¬_¬

Itxaka:
> > Looks nice! I would name the module _pkg/error_ -> _pkg/errors_ so it's possible to import without aliasing and use consistent separators for the new files instead of camelCase +1
> 
> welp, why didnt I think of that, I kept suffering on how to name the import properly as to not collide ¬_¬

wops, there is an errors package in the standard lib :D

frelon:
> > > Looks nice! I would name the module _pkg/error_ -> _pkg/errors_ so it's possible to import without aliasing and use consistent separators for the new files instead of camelCase +1
> > 
> > 
> > welp, why didnt I think of that, I kept suffering on how to name the import properly as to not collide ¬_¬
> 
> wops, there is an errors package in the standard lib :D

Hehe yep, it gets tricky when using both standard-lib and the custom one :sweat_smile:

davidcassany:
> @davidcassany How about this? Adding comments to the constants should result in the go packages doc site creating the proper docs for the constants: https://pkg.go.dev/github.com/rancher/elemental-cli/cmd#pkg-constants so we could link to there. That should be automatic and the only thing needed would be to add comments to any new constants.

Yes, quite a good option. That could be an easy way to produce something without having explicitly maintain it in different locations.

Itxaka:
> > @davidcassany How about this? Adding comments to the constants should result in the go packages doc site creating the proper docs for the constants: [pkg.go.dev/github.com/rancher/elemental-cli/cmd#pkg-constants](https://pkg.go.dev/github.com/rancher/elemental-cli/cmd#pkg-constants) so we could link to there. That should be automatic and the only thing needed would be to add comments to any new constants.
> 
> Yes, quite a good option. That could be an easy way to produce something without having explicitly maintain it in different locations.

I think that for that we probably need to set fixed numbers to the errors otherwise they wont be interpreted. Unfortunately I cannot make godoc work locally to test that ¬_¬

davidcassany:
> I think that for that we probably need to set fixed numbers to the errors otherwise they wont be interpreted.

Yes that's likely to be the case. But this is not a big issue I think, using fixed numbers is not that bad.

Itxaka:
updated, looks like this with godoc and extra comments (Could be removed the extra comments so they only point tot he error explanation:

image

Itxaka:
I fucking hate that they are out of order, not sure why  ¬_¬

davidcassany:
> I fucking hate that they are out of order, not sure why ¬_¬

Really? aren't they in the same order as the code? Probably this happens because at some point internally this is mapped to a yaml or a map container, they are not sorted.

Itxaka:
oh, they are in constant alphabetical order not value grrrrr

Itxaka:
I ordered them but this sucks becuase then they wil be displayed on the docs out of order as soon as we add a new one. either we add them all at the same time, or we keep changing values under the main branch until all are done and then we release our official version with all the exit codes in place and ordered

Itxaka:
ok, third and final @davidcassany 

This reaches a nice compromise between maintainability, ordering of the error codes, expansion in the future and auto updating the public docs with all error codes.

Mainly using the AST parser to parse constants in the exit-codes file and extract those to a md file that we can publish along our normal docs for ease of use and ease of check, no need to clone anything, just have a look at our public docs.

Also as part of the generate docs job, any extra errors added will trigger a new automated PR with the given changes, so thats mostly zero maintenance and zero possibility of adding errors and not being reflected in the docs :D

Itxaka:
Also this doesnt prevent the autogenerated docs at the https://pkg.go.dev/github.com/rancher/elemental-cli/ site for existing, they will live along so there is two places to check them.

mjura (approved):
LGTM

davidcassany:
LGTM :+1:

PR #371, opened 2022-11-14
cli: Add exit-codes to upgrade cmd

### LINKED ISSUE #370: Add exit codes to upgrade
Add exit codes

frelon:
Uses the new elemental error to add exit-codes for error paths in the upgrade command and upgrade action.

Fixes #370


cmd/config/config.go
```diff
@@ -189,6 +195,7 @@ func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interfa
 	err := viper.Unmarshal(cfg, setDecoder, decodeHook)
 	if err != nil {
 		cfg.Logger.Warnf("error unmarshalling RunConfig: %s", err)
+		return cfg, err
```
Itxaka: Im not sure we even need to return an error here, something to discuss for sure. I mean, the config is not strictly necessary, and we are already warning the user that we could not load it....but probably is good UX to rejext continuing if we cant load the user config....

pkg/action/upgrade.go
```diff
@@ -138,12 +139,12 @@ func (u *UpgradeAction) Run() (err error) {
 
 	umount, err := e.MountRWPartition(u.spec.Partitions.State)
 	if err != nil {
-		return err
+		return elementalError.New(err.Error(), elementalError.MountStatePartition)
```
Itxaka: Why are you not using the `NewFromError` ? Should we just drop it?

Itxaka (commented):
LGTM, only one small issue

pkg/action/upgrade.go
frelon: Oh, did not see that one! Will refactor to use it :+1:

cmd/config/config.go
frelon: Yea, was thinking it's safer to error out if the config is broken instead of potentially running something unwanted..

frelon: I removed the return in latest commit, we could have a discussion later if we should continue running commands if we fail to load configuration.

Itxaka: :+1: Although I think this is the correct approach, we may be kind of breaking the workflow here that we didnt before..

Itxaka (approved):

PR #374, opened 2022-11-15
cli: Add exit-codes to install command

### LINKED ISSUE #361: Add error codes to elemental-cli
Add errors to following parts and its dependencies:

 - [x] #362
 - [x] #363
 - [ ] #364
 - [x] #365
 - [x] #366
 - [x] #367
 - [x] #368
 - [x] #369
 - [x] #370

Example Exit codes PR: https://github.com/rancher/elemental-cli/pull/355

### LINKED ISSUE #365: Add exit codes to install

frelon:
Fixes #365

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cmd
 
 import (
-	"errors"
 	"os/exec"
 
 	"github.com/spf13/cobra"
@@ -26,6 +25,7 @@ import (
 
 	"github.com/rancher/elemental-cli/cmd/config"
 	"github.com/rancher/elemental-cli/pkg/action"
+	elementalError "github.com/rancher/elemental-cli/pkg/error"
 	v1 "github.com/rancher/elemental-cli/pkg/types/v1"
 )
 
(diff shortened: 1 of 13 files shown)


cmd/install.go
```diff
@@ -70,24 +72,20 @@ func NewInstallCmd(root *cobra.Command, addCheckRoot bool) *cobra.Command {
 			spec, err := config.ReadInstallSpec(cfg, cmd.Flags())
 			if err != nil {
 				cfg.Logger.Errorf("invalid install command setup %v", err)
-				return err
+				return elementalError.NewFromError(err, elementalError.ReadingSpecConfig)
```
Itxaka, line 75: uh, interesting that in the install we DO return an error if we cant load the config... :thinking:

pkg/action/install.go
```diff
@@ -269,13 +270,5 @@ func (i InstallAction) Run() (err error) {
...
-	// Reboot, poweroff or nothing
-	if i.cfg.Reboot {
-		i.cfg.Logger.Infof("Rebooting in 5 seconds")
-		return utils.Reboot(i.cfg.Runner, 5)
-	} else if i.cfg.PowerOff {
-		i.cfg.Logger.Infof("Shutting down in 5 seconds")
-		return utils.Shutdown(i.cfg.Runner, 5)
-	}
-	return err
+	return PowerAction(i.cfg)
```
Itxaka, line 273: nice

Itxaka (approved):
very nice!

pkg/action/install.go
frelon: :muscle:

cmd/install.go
frelon: aha! but it looks like it always does that in the cmd/* and then not returns in ReadRunConfig/ReadUpdateSpec.. I will remove the error-return I added in the last PR just to be consistent!

PR #375, opened 2022-11-15
cli: Add exit-codes for reset command

### LINKED ISSUE #368: Add exit codes to reset

frelon:
Fixes #368

### DIFF
--- a/pkg/error/elemental-error.go
+++ b/pkg/error/elemental-error.go
@@ -30,10 +30,6 @@ func (e *ElementalError) ExitCode() int {
 	return e.code
 }
 
-var (
-	ErrCommandRequiresRoot = New("this command requires root privileges", RequiresRoot)
-)
-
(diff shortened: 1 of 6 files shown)


pkg/error/elemental-error.go
```diff
@@ -30,6 +30,10 @@ func (e *ElementalError) ExitCode() int {
 	return e.code
 }
 
+var (
+	ErrCommandRequiresRoot = New("this command requires root privileges", RequiresRoot)
```
Itxaka: IIRC this is overridable no? So I can override this var...not sure how I feel about that....

Itxaka (approved):
damn boy, you are in a streak with this exit codes :fire:

pkg/error/elemental-error.go
frelon: Yep!  would be nice to have const-errors! I removed it since it was only used in the one place.

PR #379, opened 2022-11-18
Use state in upgrade

davidcassany:
This PR is to make use of the installation state yaml file in ugrade/reset procedures, mainly for three reasons:

* This facilitates the change of configurable default values (e.g. partitition labels), as with the change, in upgrade/reset procedure the configuration is applied as (right max prio, stacked from left to right):
  `defaults <- install state <- configuration files <- environment variables`
  so we could change the default partition labels and in an upgrade elemental-cli would still use whatever it was installed on the system. This is also key for changes like the ones discussed in rancher/elemental-toolkit#1643
* This enables us to remove some heuristics to check former installation stats (e.g. checks around squashfs)
* Reading the state file is simple and, IMHO, safer than attempting to detect configuration based on heuristics

In addition it also changes the `BootedFrom` method calls to parse the image file path instead of the image label, again this is safer as the image path is not configurable, but the label is.

### DIFF
--- a/cmd/config/config_test.go
+++ b/cmd/config/config_test.go
@@ -301,7 +301,7 @@ var _ = Describe("Config", Label("config"), func() {
 			var ghwTest v1mock.GhwMock
 
 			BeforeEach(func() {
-				bootedFrom = constants.SystemLabel
+				bootedFrom = constants.RecoveryImgFile
 				flags = pflag.NewFlagSet("testflags", 1)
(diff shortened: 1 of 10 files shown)


davidcassany:
While discussing the grub rancher/elemental-toolkit#1644 and rancher/elemental-toolkit#1643 I thought this is indeed relevant and not difficult to achieve.

Itxaka (approved):
nice!

PR #381, opened 2022-11-21
Fix config regression

### LINKED ISSUE #380: Regressions raising after bumping elemental-cli in elemental-toolkit

frelon:
Checks if config-file exists before trying to merge it into running configuration.

Fixes #380 


cmd/config/config.go
```diff
@@ -195,6 +196,7 @@ func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interfa
 	err := viper.Unmarshal(cfg, setDecoder, decodeHook)
 	if err != nil {
 		cfg.Logger.Warnf("error unmarshalling RunConfig: %s", err)
+		return cfg, err
```
davidcassany, line 199: In fact I prefer to keep it. I liked the fact of being explicit on errors, we just need to handle them better.

The changes you did on tracking proper error messages uncovered poor previous logic

davidcassany (commented):
I have been also working on fix 😅 I was about to submit it. In fact I am fighting with another corner case I saw. Unit tests for this part are a pretty annoying to code and I realized there are some glitches.

frelon:
Duplicate of #382

PR #382, opened 2022-11-21
Fix configuration loading

### LINKED ISSUE #380: Regressions raising after bumping elemental-cli in elemental-toolkit

### LINKED ISSUE #385: Duplicate "default" in help output for "install"
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Run
```
elemental install -h
```
outputs
```
...
Global Flags:
      --config-dir string   Set config dir (default is /etc/elemental) (default "/etc/elemental")
...
```

(note the duplication of "default")

**What did you expect to happen:**

```
...
Global Flags:
      --config-dir string   Set config dir (default is "/etc/elemental")
...
```

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Elemental-cli version 0.0.15+g605346d

- Elemental release version (use `cat /etc/os-release`):
- Rancher version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:

davidcassany:
This commit ensures viper only tries to merge files that already exist.

In addition this commit properly sets default configuration directory for build commands.

Fixes #380

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -100,9 +100,6 @@ func bindGivenFlags(vp *viper.Viper, flagSet *pflag.FlagSet) {
 
 func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Interface) (*v1.BuildConfig, error) {
 	logger := v1.NewLogger()
-	if configDir == "" {
-		configDir = "."
-	}
 
 	cfg := config.NewBuildConfig(
 		config.WithLogger(logger),
@@ -110,6 +107,12 @@ func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Inter
 	)
 
 	configLogger(cfg.Logger, cfg.Fs)
+	if configDir == "" {
+		configDir = "."
+		cfg.Logger.Info("reading configuration from current directory")
+	} else {
+		cfg.Logger.Info("reading configuration form '%s'", configDir)
+	}
 
 	viper.AddConfigPath(configDir)
 	viper.SetConfigType("yaml")
@@ -138,8 +141,11 @@ func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interfa
 		config.WithLogger(v1.NewLogger()),
 		config.WithMounter(mounter),
 	)
-
 	configLogger(cfg.Logger, cfg.Fs)
+	if configDir == "" {
+		configDir = constants.ConfigDir
+	}
+	cfg.Logger.Info("reading configuration form '%s'", configDir)
 
 	const cfgDefault = "/etc/os-release"
 	if exists, _ := utils.Exists(cfg.Fs, cfgDefault); exists {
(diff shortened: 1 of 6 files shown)


cmd/config/config.go
```diff
@@ -100,16 +100,19 @@ func bindGivenFlags(vp *viper.Viper, flagSet *pflag.FlagSet) {
...
 		config.WithMounter(mounter),
 	)
 
 	configLogger(cfg.Logger, cfg.Fs)
+	if configDir == "" {
+		configDir = "."
+		cfg.Logger.Info("reading configuration from current directory")
+	} else {
+		cfg.Logger.Info("reading configuration form '%s'", configDir)
+	}
```
davidcassany, line 115: Log the directory we are loading configuration from

cmd/config/config.go
```diff
@@ -138,8 +141,11 @@ func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interfa
 		config.WithLogger(v1.NewLogger()),
 		config.WithMounter(mounter),
 	)
-
 	configLogger(cfg.Logger, cfg.Fs)
+	if configDir == "" {
+		configDir = constants.ConfigDir
+	}
+	cfg.Logger.Info("reading configuration form '%s'", configDir)
```
davidcassany, line 148: Set here the configuration default directory, not at flags level, otherwise we can't distinguish between build and runtime commands.

cmd/config/config.go
```diff
@@ -148,40 +154,40 @@ func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interfa
 
 		err := viper.MergeInConfig()
 		if err != nil {
-			cfg.Logger.Warnf("error merging os-release file: %s", err)
+			cfg.Logger.Error("error merging os-release file: %s", err)
```
davidcassany, line 157: We return error, hence they should be error messages.

cmd/config/config.go
```diff
@@ -148,40 +154,40 @@ func ReadConfigRun(configDir string, flags *pflag.FlagSet, mounter mount.Interfa
...
 			return cfg, err
 		}
 	}
 
 	// merge yaml config files on top of default runconfig
-	if exists, _ := utils.Exists(cfg.Fs, configDir); exists {
+	if exists, _ := utils.Exists(cfg.Fs, filepath.Join(configDir, "config.yaml")); exists {
```
davidcassany, line 163: Make sure we tests the actual file, not the directory

cmd/root.go
```diff
@@ -30,7 +30,7 @@ func NewRootCmd() *cobra.Command {
 		Short: "Elemental",
 	}
 	cmd.PersistentFlags().Bool("debug", false, "Enable debug output")
-	cmd.PersistentFlags().String("config-dir", "/etc/elemental", "Set config dir (default is /etc/elemental)")
+	cmd.PersistentFlags().String("config-dir", "", "Set config dir")
```
davidcassany, line 33: This default was making the build configuration load to never look at the current directory. Better set the default on build config load and on run config load.

davidcassany (commented):
@frelon See this PR is slightly different from what you did, the error fix is the same: test the actual file to load exists.

However in this PR I added few additional logging comments, a couple of tests for bad formatted yamls and fix the default configuration directory for build commands.

frelon:
Very nice! I will close my PR in favor of this one! :+1:

frelon (approved):
LGTM!

PR #384, opened 2022-11-21
Ensure the manifest.yaml exist before attempting to read it

davidcassany:
This commit instead of swallowing viper errors when loading config files it reports back the error for build commands. This is the same criteria as it is for the run time commands.


frelon (approved):
Very nice! :+1:

PR #387, opened 2022-11-22
Set garbage collector temp dir for luet

### LINKED ISSUE #380: Regressions raising after bumping elemental-cli in elemental-toolkit

davidcassany:
This commit sets the temp dir for luet on start up.

It also fixes a bug on partitions detection, where the device name was used instead of the partition label. In elemental-cli partition names are partition labels, they are hardcoded and not configurable.

Finally fixes few log messages

Fixes #380

### DIFF
--- a/pkg/luet/luet.go
+++ b/pkg/luet/luet.go
@@ -128,6 +128,9 @@ func NewLuet(opts ...Options) *Luet {
 	if luet.context == nil {
 		luetConfig := luet.createLuetConfig()
 		luet.context = context.NewContext(context.WithConfig(luetConfig), context.WithLogger(luet.log))
+		if luet.TmpDir != "" {
+			luet.context.GarbageCollector = gc.GarbageCollector(filepath.Join(luet.TmpDir, "tmpluet"))
+		}
 	}
 
 	if luet.auth == nil {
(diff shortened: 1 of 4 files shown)


pkg/luet/luet.go
```diff
@@ -128,6 +128,9 @@ func NewLuet(opts ...Options) *Luet {
 	if luet.context == nil {
 		luetConfig := luet.createLuetConfig()
 		luet.context = context.NewContext(context.WithConfig(luetConfig), context.WithLogger(luet.log))
+		if luet.TmpDir != "" {
+			luet.context.GarbageCollector = gc.GarbageCollector(filepath.Join(luet.TmpDir, "tmpluet"))
```
davidcassany, line 132: I believe this was the real issue I was seeing... setting those temp dirs is a PITA.

Moreover I believe some of those are just left behind after executing luet... but this should be tackled in a separate PR

pkg/luet/luet.go
```diff
@@ -392,8 +395,12 @@ func (l Luet) createLuetConfig() *luetTypes.LuetConfig {
...
-	l.context.Config.System.PkgsCachePath = l.TmpDir
-	l.context.Config.System.DatabasePath = l.TmpDir
-	l.context.GarbageCollector = gc.GarbageCollector(filepath.Join(tmpdir, "tmpluet"))
+	if l.context != nil {
+		if l.context.Config != nil {
+			l.context.Config.System.TmpDirBase = l.TmpDir
+			l.context.Config.System.PkgsCachePath = l.TmpDir
+			l.context.Config.System.DatabasePath = l.TmpDir
+		}
+		l.context.GarbageCollector = gc.GarbageCollector(filepath.Join(tmpdir, "tmpluet"))
```
davidcassany, line 404: This is just the result of myself being paranoid trying to trace all the luet configuration 😅 In any case I think this is better having this sanity checks as if `SetTempDir` is executed in a non properly configured Luet this would panic

pkg/utils/getpartitions.go
```diff
@@ -32,7 +32,7 @@ func ghwPartitionToInternalPartition(partition *block.Partition) *v1.Partition {
 	return &v1.Partition{
 		FilesystemLabel: partition.FilesystemLabel,
 		Size:            uint(partition.SizeBytes / (1024 * 1024)), // Converts B to MB
-		Name:            partition.Name,
+		Name:            partition.Label,
```
davidcassany, line 35: Yes, I know, this is confusing, we discussed already several times. Partially my fault, probably I could have chosen better names... on my defense all the command line tools around (`blkid` and friends) are also misleading and confusing regarding the filesystem and partitions labels 😓

pkg/luet/luet.go
Itxaka: goddammit, this tempdir stuff is killing me, we keep finding issues about it ¬_¬

Itxaka (approved):

PR #390, opened 2022-11-30
Add docker auth flags and luet init

### LINKED ISSUE #389: Unable to pull an image that requires authentication
Build ISO

```
 elemental build-iso  --config-dir /root/derivative --overlay-iso /root/derivative/overlay/iso --debug -o /root/derivative  172.16.1.208/library/example:v10.1
```
```
ERRO[2022-11-30T13:28:28+08:00] Failed installing OS packages: GET http://172.16.1.208/v2/library/example/manifests/v10.1: UNAUTHORIZED: unauthorized to access repository: library/example, action: pull: unauthorized to access repository: library/example, action: pull 
ERRO[2022-11-30T13:28:28+08:00] 1 error occurred:
        * GET http://172.16.1.208/v2/library/example/manifests/v10.1: UNAUTHORIZED: unauthorized to access repository: library/example, action: pull: unauthorized to access repository: library/example, action: pull
```
Unable to pull an image that requires authentication. The public repository is normal

Itxaka:
Otherwise if anything calling the luet with a docker source needs auth, we cannot pass it

Fixes #389


pkg/config/config.go
```diff
@@ -146,7 +154,10 @@ func NewConfig(opts ...GenericOptions) *v1.Config {
 
 	if c.Luet == nil {
 		tmpDir := utils.GetTempDir(c, "")
-		c.Luet = luet.NewLuet(luet.WithFs(c.Fs), luet.WithLogger(c.Logger), luet.WithLuetTempDir(tmpDir))
+		if c.DockerAuth == nil {
```
frelon, line 157: This check looks like it's done in _pkg/luet/luet.go:136_ as well. Is this needed here?

frelon (approved):
Small question, but looks good!

pkg/types/v1/config.go
```diff
@@ -56,6 +57,7 @@ type Config struct {
 	Arch                      string       `yaml:"arch,omitempty" mapstructure:"arch"`
 	SquashFsCompressionConfig []string     `yaml:"squash-compression,omitempty" mapstructure:"squash-compression"`
 	SquashFsNoCompression     bool         `yaml:"squash-no-compression,omitempty" mapstructure:"squash-no-compression"`
+	DockerAuth                *types.AuthConfig
```
davidcassany, line 60: I am wondering if we should enable these values to be configured by environment variables or by the config yaml file, probably we don't.

Importing directly the struct from the docker api prevents setting up our own serialization so they can't be easily serialized and deserialized together with other flags, env vars or config variables.

Wouldn't it be worth having our own struct for that matter? Also if this is needed for upgrades probably makes sense having it in config.yaml.

pkg/types/v1/config.go
```diff
@@ -24,6 +24,7 @@ import (
 	"gopkg.in/yaml.v3"
 	"k8s.io/mount-utils"
 
+	"github.com/docker/docker/api/types"
```
davidcassany, line 27: nit: this appears to be deprecated/moved here

https://github.com/moby/moby/blob/master/api/types/auth.go

Probably better importing the actual definition from

https://github.com/moby/moby/blob/master/api/types/registry/authconfig.go

cmd/config/config.go
```diff
@@ -144,9 +162,26 @@ func ReadConfigBuild(configDir string, flags *pflag.FlagSet, mounter mount.Inter
...
+
+	dockerAuth := &types.AuthConfig{
+		Username:      user,
+		Password:      pass,
+		ServerAddress: server,
+		Auth:          authType,
+		IdentityToken: identity,
+		RegistryToken: registryToken,
+	}
+
```
davidcassany, line 180: All this is unneeded if the data is serialized and deserialized as other values. It would just get in by `Viper.MergeInConfig` if set in config yamls and from `bindGivenFlags(viper.GetViper(), flags)` if set infalgs. However for that to happen they need to be both in the same map structure, this is the counter part.

davidcassany (commented):
Looks good, i am just wondering if only having it in flags is what we want. If so this is all fine, if we consider this to be part of the yaml then we should be careful, viper plus mastructure is pretty annoying in that regard as the flags should perfectly match yaml keys IIRC.

pkg/config/config.go
Itxaka: I think you are rigth here, we can safely ignore this as luet will be in charge of using the auth or not

pkg/types/v1/config.go
Itxaka: Im also not clear on that. It was fine to add it as flags but...it leaks the secrets for a long running process (i.e. someone doing a ps will see your user/password).

Also not nice in env vars (althougth useful and more anti-leak...kind of) and on config files....

PR #394, opened 2022-12-14
Add OBS build files into the repository

davidcassany:
Part of rancher/elemental#577


davidcassany:
@Itxaka  looks like the secboot version we were using in go.mod is no longer available... I did an attempt to manually recreate the vendor folder without using the secboot fork but I miserably failed... I am worried we might be forced to keep around legacy code no longer available upstream, I am missing some background on this issue, do you have any idea about how should we proceed or workaround it?

kkaempf (approved):
Having the builder image Dockerfile here is a bit strange but I don't know a better place either :-/

Itxaka:
> @Itxaka looks like the secboot version we were using in go.mod is no longer available... I did an attempt to manually recreate the vendor folder without using the secboot fork but I miserably failed... I am worried we might be forced to keep around legacy code no longer available upstream, I am missing some background on this issue, do you have any idea about how should we proceed or workaround it?

yikes...seems like they removed the fork that nullboot is using...weird, seeing that nullboot is still using that fork. Bit of a mess actually, because we dont want to recreate the whole thing ourselves....argh.

Itxaka:
opened an issue upstream as it also happens in the nullboot repo: https://github.com/canonical/nullboot/issues/28

davidcassany:
> Having the builder image Dockerfile here is a bit strange but I don't know a better place either :-/

We could keep it entirely within the OBS only, like the the `build-iso` image. However I do believe it makes sense it here. We call it the builder image, however it could also be called the `elemental-cli` image as this is what it actually includes and it makes sense for any `elemental build` command.

Itxaka:
yikes, its a bit of a mess.

nullboot requires secboot which in turn requres go-tpm2. But all of those versions are a bit mangled and old. Updating them requires a chain of updates to the code in all three repos....we migth need to just implement  the parts that we are using from nullboot (list entry vars, remove entry, add entry) and call it a day, otherwise...I dont see a clear path in which we update 3 different libraries to be in sync, especially with all this TPM stuff in the middle, which we dont even care about here. Ugh.

Issue #393, opened 2022-12-08
Don't try to write efivars on Raspberry


kkaempf:
`elemental install` fails on Raspberry since there's no efivars storage.

The log shows
```
2022/12/08 17:39:07 Could not read BootOrder variable, populating with default, error was: variable does not exist
ERRO[2022-12-08T17:39:07Z] error creating boot entry: open /sys/firmware/efi/efivars/Boot0000-8be4df61-93ca-11d2-aa0d-00e098032b8c: read-only file system
...
Error: 1 error occurred:
        * open /sys/firmware/efi/efivars/Boot0000-8be4df61-93ca-11d2-aa0d-00e098032b8c: read-only file system
time="2022-12-08T17:41:34Z" level=fatal msg="failed calling elemental client: exit status 1"
```

kkaempf:
I would've thought that setting
```
spec:
  config:
     elemental:
        install:
          disable-boot-entry: true
```
would prevent writing of EFI variables.

However, when editing `machineRegistration` in Rancher (operator 1.1.0) this setting isn't accepted.

kkaempf:
Hmm, `elemental-cli`'s set of [install variables](https://github.com/rancher/elemental-cli/blob/5ae32204b0f2160365c8d64997068b2be1a25133/pkg/types/v1/config.go#L134) is different from [`elemental-operator`'s set](https://github.com/rancher/elemental-operator/blob/af83eea27fbc5b2149c48bf4fef367aa72b77035/api/v1beta1/types.go#L21) 🤔

kkaempf:
Hmm, latest `elemental-operator` now accepts `disable-boot-entry: true` in `install:` but `elemental-cli` still tries to write an EFI variable.

kkaempf:
Install log

`disable-boot-entry` is `true`
```
...
Dec 19 10:11:26 rancher-5628 elemental[1042]: time="2022-12-19T10:11:26Z" level=debug msg="Fetched configuration from manager cluster:\nelemental:\n  install:\n    device: /dev/mmcblk0\n    debug: true\n    reboot: true\n    disable-boot-entry: true\n  registration:\n    url: https://192.168.0.28/elemental/registration/549rjd99gxjfmblrmfjmfl26zkzh7fsbssnqx49j6wd4b4g4dsc446\n    
```

but this doesn't get passed properly:

```
...
[...]

PR #395, opened 2022-12-14
Copy ARM firmware files to boot partition

### LINKED ISSUE #391: ARM firmware files missing on installed disk
While installing Elemental works nicely on Raspberry Pi, it's missing the firmware files to boot.

When booting in live mode, this can be fixed by

```
# mount /dev/disk/by-label/COS_GRUB /mnt
# cp -a /boot/vc/* /mnt
# umount /mnt
```

This step should be integrated into `elemental install` when running on `aarch64`

kkaempf:
Fixes #391

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -83,8 +83,17 @@ func GetFullDeviceByLabel(runner v1.Runner, label string, attempts int) (*v1.Par
 
 // CopyFile Copies source file to target file using Fs interface. If target
 // is  directory source is copied into that directory using source name file.
-func CopyFile(fs v1.FS, source string, target string) (err error) {
-	return ConcatFiles(fs, []string{source}, target)
+// File mode is preserved
+func CopyFile(fs v1.FS, source string, target string) error {
+	fInf, err := fs.Stat(source)
+	if err != nil {
+		return err
+	}
+	err = ConcatFiles(fs, []string{source}, target)
+	if err != nil {
+		return err
+	}
+	return fs.Chmod(target, fInf.Mode())
 }
 
 // ConcatFiles Copies source files to target file using Fs interface.
@@ -128,6 +137,58 @@ func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
 	}
 
 	return err
+}
+
+// CopyDirectory copies all files from a source directory to destination directory.
+// Destination directory is created if it doesn't exist.
+func CopyDirectory(fsys v1.FS, srcDir, dstDir string) error {
+
+	fInf, err := fsys.Lstat(srcDir)
+	if err != nil {
+		return err
+	}
+	if !fInf.IsDir() {
+		return fmt.Errorf("srcDir '%s' is not a directory", srcDir)
+	}
+
+	if exists, _ := Exists(fsys, dstDir); exists {
+		if dir, err := IsDir(fsys, dstDir); !dir {
+			return fmt.Errorf("dstDir '%s' is not a directory: %v", dstDir, err)
+		}
+	} else {
+		err = MkdirAll(fsys, dstDir, fInf.Mode())
+		if err != nil {
+			return err
+		}
+	}
+
+	return WalkDirFs(fsys, srcDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		var fInf fs.FileInfo
+		var link string
+
+		outpath := filepath.Join(dstDir, strings.TrimPrefix(path, srcDir))
+		fInf, err = d.Info()
+		if err != nil {
+			return err
+		}
+
+		switch fInf.Mode().Type() & os.ModeType {
+		case os.ModeDir:
+			return MkdirAll(fsys, outpath, fInf.Mode())
+		case os.ModeSymlink:
+			link, err = fsys.Readlink(path)
+			if err != nil {
+				return err
+			}
+			return fsys.Symlink(link, outpath)
+		default:
+			return CopyFile(fsys, path, outpath)
+		}
+	})
 }
 
 // Copies source file to target file using Fs interface
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -85,27 +85,23 @@ func GetFullDeviceByLabel(runner v1.Runner, label string, attempts int) (*v1.Par
 // is  directory source is copied into that directory using source name file.
 // File mode is preserved
 func CopyFile(fs v1.FS, source string, target string) error {
-	fInf, err := fs.Stat(source)
-	if err != nil {
-		return err
-	}
-	err = ConcatFiles(fs, []string{source}, target)
-	if err != nil {
-		return err
-	}
-	return fs.Chmod(target, fInf.Mode())
+	return ConcatFiles(fs, []string{source}, target)
 }
 
 // ConcatFiles Copies source files to target file using Fs interface.
 // Source files are concatenated into target file in the given order.
 // If target is a directory source is copied into that directory using
-// 1st source name file.
+// 1st source name file. The result keeps the file mode of the 1st source.
 func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
 	if len(sources) == 0 {
 		return fmt.Errorf("Empty sources list")
 	}
 	if dir, _ := IsDir(fs, target); dir {
 		target = filepath.Join(target, filepath.Base(sources[0]))
+	}
+	fInf, err := fs.Stat(sources[0])
+	if err != nil {
+		return err
 	}
 
 	targetFile, err := fs.Create(target)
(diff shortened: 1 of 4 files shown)


pkg/utils/grub.go
```diff
@@ -122,6 +122,17 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
 		return err
 	}
 
+	// copy ARM firmware files
+	if g.config.Arch == cnst.ArchArm64 {
+		g.config.Logger.Infof("Copying ARM firmware files")
+		// run via shell for "*" extension
```
davidcassany: @kkaempf thanks much for the fix.

I'd rather create an utility for copying all files in a folder (we could even include recursive flag). Probably it will be also be used in other places and it could be easily tested in unit tests.

davidcassany (changes requested):
I'd move the copy all files logic to an utility function to test it properly in unit tests. Then I'd also try to code a grub unit test covering this arm path.

@kkaempf let us know if you prefer to us to code such a little extension on top of your branch

kkaempf:
I'm a Go illiterate (I can copy & paste though 😝 ) and would very much prefer help here 😉

davidcassany:
Rebased and refactored to make use of new CopyDirectory utility method.

pkg/utils/grub.go
```diff
@@ -32,6 +32,7 @@ import (
 )
 
 const bootEntryName = "elemental-shim"
+const armFirmwarePath = "/boot/vc"
```
davidcassany, line 35: @kkaempf it is unclear where this path comes from. I doubt we can assume this is always valid, even in non SUSE systems. Gonna try to investigate it, but I am pretty ignorant on the device tree topic.

pkg/utils/common.go
```diff
@@ -83,8 +83,17 @@ func GetFullDeviceByLabel(runner v1.Runner, label string, attempts int) (*v1.Par
...
+func CopyFile(fs v1.FS, source string, target string) error {
+	fInf, err := fs.Stat(source)
+	if err != nil {
+		return err
+	}
+	err = ConcatFiles(fs, []string{source}, target)
+	if err != nil {
+		return err
+	}
+	return fs.Chmod(target, fInf.Mode())
```
davidcassany: I realized our former CopyFile method was not keeping former file mode, I think it is better to simply keep it, if this raises some issue we can later evolve it to apply custom file modes, but I'd rather keep it simple for now and see.

pkg/utils/grub.go
kkaempf: This path comes from the `raspberrypi-firmware*` rpms

pkg/utils/common.go
```diff
@@ -130,6 +139,58 @@ func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
 	return err
 }
 
+// CopyDirectory copies all files from a source directory to destination directory.
+// Destination directory is created if it doesn't exist.
+func CopyDirectory(fsys v1.FS, srcDir, dstDir string) error {
```
kkaempf, line 140: Does this do a recursive copy ? It's not immediately obvious to me.

kkaempf (commented):
I'd happily give this a try at https://build.opensuse.org/package/show/home:kwk:osimage/elemental-cli 😉

pkg/utils/common.go
davidcassany: Yes it does

pkg/utils/common.go
```diff
@@ -115,19 +120,64 @@ func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
...
+
+	fInf, err := fsys.Lstat(srcDir)
+	if err != nil {
+		return err
+	}
+	if !fInf.IsDir() {
+		return fmt.Errorf("srcDir '%s' is not a directory", srcDir)
+	}
+
+	return WalkDirFs(fsys, srcDir, func(path string, d fs.DirEntry, err error) error {
```
davidcassany, line 150: This is the recursive method. `WalkDirFs` calls the function provided as a parameter to all files it finds in the tree.

pkg/utils/grub.go
davidcassany: Yes I realized that and I believe this is openSUSE & raspberry specific, hence I am not really in favor of adding this stuff here... See the arm openSUSE images built with KIWI, it is quite frustrating to see an if clause for each board where some firmware board specific magic is applied. Here is the [script](https://build.opensuse.org/package/view_file/openSUSE:Factory:ARM:ToTest/JeOS/uboot-image-raspberrypi-install?expand=1) that patches boot for a rpi (and others) image.

davidcassany: I believe this logic should be applied somewhere else, u-boot related stuff is board specific. A couple (or so) of years ago we declined adding some specific u-boot support to KIWI because is was just not possible to set something generic that could be functional for `arm` in general.

davidcassany (commented):
I don't think this PR can be merged as is. Could be used for testing and validating the logic, but we should follow some other approach. Adding this patch in grub installation logic is RPI & openSUSE specific, we should avoid that.

davidcassany:
I believe the logic to add this firmware files in EFI partitions should somehow be part of https://build.opensuse.org/package/view_file/isv:Rancher:Elemental:Dev/elemental-post-build-generate-rpi-image/generate-rpi-image.sh?expand=1

pkg/utils/grub.go
kkaempf: I more and more believe we should differentiate between `architecture` (like ARM) and `platform` (like RPi) to cover all these variants.

kkaempf: That said, there's huge interest in Elemental + RPi from various sides.

Btw, this PR (plus fixing #393) makes Rancher+Elemental fully functional on Raspberry. 😉

kkaempf:
> I believe the logic to add this firmware files in EFI partitions should somehow be part of https://build.opensuse.org/package/view_file/isv:Rancher:Elemental:Dev/elemental-post-build-generate-rpi-image/generate-rpi-image.sh?expand=1

It's already there (otherwise the `.raw` image wouldn't even boot 😉 ) - for the ~installation~seed image.

This PR, however, is for the `elemental install` part.

pkg/utils/grub.go
davidcassany: > I more and more believe we should differentiate between `architecture` (like ARM) and `platform` (like RPi) to cover all these variants.

I definitely agree on that statement. However I am a bit resistant to start adding these variants in elemental-cli, it can easily grow into a monster.

What if we make the platform a user provided value? it could be a value in out elemental-cli config. So we do not have to code heuristics to determine the platform, we just assume we know which one is.

davidcassany:
> > I believe the logic to add this firmware files in EFI partitions should somehow be part of https://build.opensuse.org/package/view_file/isv:Rancher:Elemental:Dev/elemental-post-build-generate-rpi-image/generate-rpi-image.sh?expand=1
> 
> It's already there (otherwise the `.raw` image wouldn't even boot wink ) - for the ~installation~seed image.
> 
> This PR, however, is for the `elemental install` part.

Yes I know there is interest in RPi and it would be really  nice being capable to provide demos using RPi. I am really in to get this sorted out and provide a functional solution, I do have concerns about not approaching it as we should though. 

I am also wondering why are we considering an `elemental install` step for RPi if we are already building a disk image, there is already a detection `elemental register` command not to install a system over a previous installation. As far as I understand this requires a couple of block devices in RPi, one as the installation media and another one where to install the system. These two are removable devices, it feels odd to me. One should be enough.

Still thinking about this story...

pkg/utils/grub.go
kkaempf: I'd absolutely agree to that !

davidcassany:
For now I'd vote to manage this in an after-install hook. I created an elemental-cli PR to execute `after-install` hooks earlier, just right after `after-install-chroot`, this way both are applied in the same step with the only difference of being in a chroot or not, which is also what the name suggests. This allows to run after-install hooks while the OS image being deployed is loop mounted and all the relevant partitions are also mounted, all in rw mode.

https://github.com/rancher/elemental-cli/pull/399

With this change we can easily consider adding a simpler after-install hook to do the job within the installation media, see https://build.opensuse.org/package/show/home:dcassany:Elemental:Dev/elemental-post-build-generate-rpi-image

Need validate this approach, but, for now, I think this should be the way to go and move use case specific logic to hooks.

davidcassany:
@kkaempf I am building a RPi image in https://build.opensuse.org/project/show/home:dcassany:Elemental:Dev:Teal53 that should be ready to be used. Do you think you can get it a trial? I'd be more in favor of using the hooks approach rather than coding it into the elemental-cli grub install.

Just need to make sure to include the `config-dir` within the registration yaml:

```yaml
...
install:
  config-dir: "/run/initramfs/live/elemental"
  ...
```

I wonder if it makes sense having this config dir as a default, so we can simply provide a custom `config.yaml` for elemental-cli by dropping it in `/elemental` on the ISO filesystem.

davidcassany:
Closing as this has been fixed by adding an after-install hook in installation media.

PR #397, opened 2022-12-15
Drop bootmanager lib use from grub

Itxaka:
Instead just bring what we are using which amounts to a few lines and use the efilib library as much as possible.

It could even be shirnk down further, but at least this patch drops dependencies that have proven to be problematic


davidcassany:
@Itxaka I managed to get rid of the dependency errors by upgrading ginkgo version. See last commit on my PR https://github.com/rancher/elemental-cli/pull/394/commits/cf43e4e513dba922aeac2450d3ff61e998dcfc1f to me it looks like the  `make test_deps` step is causing issues.

davidcassany:
In fact if you want we can merge #394 and then if you rebase from there the diff should clearly show the changes and improvements you are working on

Itxaka:
> @Itxaka I managed to get rid of the dependency errors by upgrading ginkgo version. See last commit on my PR [cf43e4e](https://github.com/rancher/elemental-cli/commit/cf43e4e513dba922aeac2450d3ff61e998dcfc1f) to me it looks like the `make test_deps` step is causing issues.

Still, we are at the whim of an external party maintaining a fork. Should probably drop that as soon as possible or at least have our own fork so we know its always there. But I rather simplify and drop that asap

frelon (approved):
Very nice, finally able to drop those deps!

PR #399, opened 2022-12-20
Add post-* hooks

davidcassany:
This commit adds post-* hooks where the after hooks were originally defined. After hooks are moved to happen straight in sequence after the after-*-chroot hooks. So now `after-install-chroot` happens first and straight after `after-install` is executed, the only difference is that the first is executed in a chroot.

The main motivation for such a change is having a hook that includes the partition, the running system and the deployed image all mounted and easily accessible.


davidcassany:
> The main motivation for such a change is having a hook that includes the partition, the running system and the deployed image all mounted and easily accessible.

For instance, this turns to be relevant to run an `after-install` hook to copy RPi firmware to EFI partition. Another example, could be having an `after-upgrade` hook that arranges grub depending on the `grub.cfg` in state partition vs the `grub.cfg` in the new OS.

davidcassany:
This PR is being tested within elemental-toolkit in https://github.com/rancher/elemental-toolkit/actions/runs/3738250925

kkaempf (approved):
Missing docs and examples (:wink:) but otherwise :+1:

davidcassany:
> Missing docs and examples (wink) but otherwise +1

Yes, docs will be added in a follow up PR in elemental-toolkit.

PR #404, opened 2023-01-11
Refactor elemental

### LINKED ISSUE #172: Guess image size from source
https://github.com/rancher-sandbox/elemental/blob/80db95e02be93b47372ac25c092c13cc8d228d47/pkg/types/v1/config.go#L80

If we could guess the size of the image needed from the source - either analyzing it or fetching a manifest - we don't need anymore to specify a default image size. 

Also, while we can have a large size here to cover most use cases as we already do, in case where we want to reduce disk footprint usage this is sub-optimal.

davidcassany:
This commit refactors elemental image deploy procedure to always create and prepare the root tree in a temporary folder before synching it to a filesystem image most likely as a mounted loop device.

This change has a couple of immediate benefits:

1. We can pre-compute the root tree size before creating the filesystem  image. Hence image sizes can be adjusted to root-tree size. Currently in this PR the default size of the images is set to `0` which means it will use as much as the image requires plus and arbitrary overhead of 256MB (we could consider making it configurable if needed). The use of an static size is still possible by setting the partition size to any value greater than `0`.

2. There is no path differentiation between filesystems. The root tree is prepared following the same logic independently of the target filesystem. Squashfs is no longer an exception, building the image follows the same logic as it was any other writable filesystem.

We also can argue this is a simple, robust and flexible logic compared with the previous code. 

The counter part is having to prepare the root-tree to later on copy it to the final image, this causes the root-tree to be written twice, one to prepare it and another one to sync it to the target location.

Fixes #172

### DIFF
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -378,7 +378,7 @@ func (e *Elemental) CreateImgFromTree(root string, img *v1.Image, cleaner func()
 	return err
 }
 
-// CopyFileImg copies the files targed as the source of this image. It also applies the img label over the copied image.
+// CopyFileImg copies the files target as the source of this image. It also applies the img label over the copied image.
 func (e *Elemental) CopyFileImg(img *v1.Image) error {
 	if !img.Source.IsFile() {
 		return fmt.Errorf("Copying a file image requires an image source of file type")
--- a/pkg/elemental/elemental.go
+++ b/pkg/elemental/elemental.go
@@ -248,7 +248,7 @@ func (e Elemental) UnmountImage(img *v1.Image) error {
 	return err
 }
 
-// CreateFileSystemImage creates the image file for config.target
+// CreateFileSystemImage creates the image file for the given image
 func (e Elemental) CreateFileSystemImage(img *v1.Image) error {
 	e.config.Logger.Infof("Creating file system image %s", img.File)
 	err := utils.MkdirAll(e.config.Fs, filepath.Dir(img.File), cnst.DirPerm)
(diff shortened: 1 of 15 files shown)


kkaempf (changes requested):
I'd recommend to put the copyright year changes in a separate PR

pkg/elemental/elemental.go
```diff
@@ -281,69 +281,140 @@ func (e Elemental) CreateFileSystemImage(img *v1.Image) error {
...
+		err = utils.SyncData(e.config.Logger, e.config.Fs, root, img.MountPoint)
 		if err != nil {
-			return nil, err
+			return err
 		}
 	}
+	return err
+}
+
+// CopyFileImg copies the files targed as the source of this image. It also applies the img label over the copied image.
```
frelon: Nit: targed

pkg/utils/common.go
```diff
@@ -83,21 +83,26 @@ func GetFullDeviceByLabel(runner v1.Runner, label string, attempts int) (*v1.Par
...
+// File mode is preserved
+func CopyFile(fs v1.FS, source string, target string) error {
 	return ConcatFiles(fs, []string{source}, target)
 }
 
 // ConcatFiles Copies source files to target file using Fs interface.
 // Source files are concatenated into target file in the given order.
 // If target is a directory source is copied into that directory using
-// 1st source name file.
+// 1st source name file. The result keeps the file mode of the 1st source.
```
frelon, line 94: Just curious, why only the 1st source?

davidcassany: Yes I know this might look weird. This method concatenates several files, if only used with a single file this is a simple copy.

So imagine a call (pseudo code) like:

```
concat my/file1 my/file2 my/file3 my/folder/
```

The result will be a `my/folder/file1` with `file2` and `file3` appended to it. So permissions and file name are just borrowed from the first one. So if there is only a single file this is indeed a regular copy. If there was a nice way of adding a default file mode with some sort of optional argument this is what I'd have done.

frelon: Aha, gotcha!

kkaempf (approved):
lgtm 🤞🏻

frelon (approved):
Works like a charm!

fgiudici (approved):

PR #405, opened 2023-01-11
Make lint happy

davidcassany:
Updates copyright header and stops using a deprecated package form the golang standard library to make lint happy

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,7 +17,7 @@ linters-settings:
     min-complexity: 30
   goheader:
     template: |-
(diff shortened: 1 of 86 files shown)


kkaempf:
The copyright year should be a range and include the first year of copyright. So that would be `2022 - 2023` in this case.

davidcassany:
> The copyright year should be a range and include the first year of copyright. So that would be `2022 - 2023` in this case.

Right, I am adapting it

kkaempf (approved):
🤩

PR #413, opened 2023-01-27
Ignores any defined but not existing cloud-init paths

davidcassany:
Part of rancher/elemental#614

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -118,11 +118,12 @@ const (
 	SELinuxTargetedPolicyPath  = SELinuxTargetedPath + "/policy"
 
 	// Kernel and initrd paths are arbitrary and coupled to grub.cfg
-	IsoKernelPath = "/boot/kernel"
-	IsoInitrdPath = "/boot/initrd"
-	IsoRootFile   = "rootfs.squashfs"
-	IsoEFIImg     = "uefi.img"
-	ISOLabel      = "COS_LIVE"
+	IsoKernelPath    = "/boot/kernel"
+	IsoInitrdPath    = "/boot/initrd"
+	IsoRootFile      = "rootfs.squashfs"
+	IsoEFIImg        = "uefi.img"
+	ISOLabel         = "COS_LIVE"
+	ISOCloudInitPath = LiveDir + "/iso-config"
 
 	// Default directory and file fileModes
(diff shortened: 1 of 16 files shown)


davidcassany:
just to not fail if we ever configure a non existing directory over a RO filesystem (ISO filesystem for instance)

davidcassany:
Note gomega is updated in a separated commit, tests were failing so as it was a small not really noisy version bump I just applied it in this same PR.

davidcassany:
Finally added a default path to cloud-init files when booting from ISO. I did not remember at first there was a `cos.setup` parameter parsed in the cmdline to include an extra cloud-init URI. In addition the changes in this PR also allow setting the cloud-init paths in the config file to remote locations, files (not directories) and whatever yip supports.

davidcassany:
Still have to run a manual test, but the expectation is that ISOs built with this elemental-cli version will load cloud-init files included in an `/iso-config` folder with the ISO filesystem root. That folder could include cloud-init files for the ISO boot itself and it could also include installation hooks. Any `elemental run-stage` command executed from the ISO parses the `/proc/cmdline`  and this includes `cos.setup=/run/initramfs/live/iso-config` so this would be parsed at boot and at install time.

pkg/constants/constants.go
```diff
@@ -118,11 +118,12 @@ const (
...
-	IsoInitrdPath = "/boot/initrd"
-	IsoRootFile   = "rootfs.squashfs"
-	IsoEFIImg     = "uefi.img"
-	ISOLabel      = "COS_LIVE"
+	IsoKernelPath    = "/boot/kernel"
+	IsoInitrdPath    = "/boot/initrd"
+	IsoRootFile      = "rootfs.squashfs"
+	IsoEFIImg        = "uefi.img"
+	ISOLabel         = "COS_LIVE"
+	ISOCloudInitPath = LiveDir + "/iso-config"
```
frelon, line 126: Nit: mixing of ISO and Iso in variable name

frelon (approved):
LGTM

pkg/constants/constants.go
davidcassany: You are right indeed, probably this is the right time to fix it. Gonna make follow up PR if you don't mind :+1: Thanks for the notice though

PR #414, opened 2023-01-27
Set labels used by grub to find system disks

### LINKED ISSUE #632: Upgrading from Stable Elemental Teal to Dev Elemental Teal fails
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

The OS upgrade fails. Mostly because the upgrade hooks are applied with a wrong environment

**What did you expect to happen:**

A successful upgrade

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

The issue is related with the suc-upgrade script and with its runtime environment.

**Environment:**

- Elemental release version (use `cat /etc/os-release`):
- Rancher version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:

frelon:
This commit sets GRUB persistent variables for filesystem labels after install/upgrade/reset. These labels are used in the new GRUB configuration when booting the system.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -208,12 +208,13 @@ func (i InstallAction) Run() (err error) {
 	err = grub.SetPersistentVariables(
 		filepath.Join(i.spec.Partitions.State.MountPoint, constants.GrubOEMEnv),
 		map[string]string{
-			"state_label":    i.spec.Partitions.State.FilesystemLabel,
-			"active_label":   i.spec.Active.Label,
-			"passive_label":  i.spec.Passive.Label,
-			"recovery_label": i.spec.Recovery.Label,
-			"system_label":   i.spec.Partitions.Recovery.FilesystemLabel,
-			"oem_label":      i.spec.Partitions.OEM.FilesystemLabel,
+			"state_label":      i.spec.Partitions.State.FilesystemLabel,
+			"active_label":     i.spec.Active.Label,
+			"passive_label":    i.spec.Passive.Label,
+			"recovery_label":   i.spec.Recovery.Label,
+			"system_label":     i.spec.Partitions.Recovery.FilesystemLabel,
+			"oem_label":        i.spec.Partitions.OEM.FilesystemLabel,
+			"persistent_label": i.spec.Partitions.Persistent.FilesystemLabel,
 		},
 	)
 	if err != nil {
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -205,22 +205,10 @@ func (i InstallAction) Run() (err error) {
 		return elementalError.NewFromError(err, elementalError.HookAfterInstall)
 	}
 
-	grubEnv := map[string]string{
-		"state_label":    i.spec.Partitions.State.FilesystemLabel,
-		"active_label":   i.spec.Active.Label,
-		"passive_label":  i.spec.Passive.Label,
-		"recovery_label": i.spec.Recovery.Label,
-		"system_label":   i.spec.Partitions.Recovery.FilesystemLabel,
-		"oem_label":      i.spec.Partitions.OEM.FilesystemLabel,
-	}
-
-	if i.spec.Partitions.Persistent != nil {
-		grubEnv["persistent_label"] = i.spec.Partitions.Persistent.FilesystemLabel
-	}
-
+	grubVars := i.spec.GetGrubLabels()
 	err = grub.SetPersistentVariables(
 		filepath.Join(i.spec.Partitions.State.MountPoint, constants.GrubOEMEnv),
-		grubEnv,
+		grubVars,
 	)
 	if err != nil {
 		i.cfg.Logger.Error("Error setting GRUB labels: %s", err)
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/rancher/elemental-cli/pkg/constants"
 	cnst "github.com/rancher/elemental-cli/pkg/constants"
 	"github.com/rancher/elemental-cli/pkg/elemental"
 	elementalError "github.com/rancher/elemental-cli/pkg/error"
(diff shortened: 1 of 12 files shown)


pkg/action/install.go
```diff
@@ -205,6 +206,22 @@ func (i InstallAction) Run() (err error) {
 		return elementalError.NewFromError(err, elementalError.HookAfterInstall)
 	}
 
+	err = grub.SetPersistentVariables(
+		filepath.Join(i.spec.Partitions.State.MountPoint, constants.GrubOEMEnv),
+		map[string]string{
+			"state_label":    i.spec.Partitions.State.FilesystemLabel,
+			"active_label":   i.spec.Active.Label,
+			"passive_label":  i.spec.Passive.Label,
+			"recovery_label": i.spec.Partitions.Recovery.FilesystemLabel,
```
davidcassany: Edited: ~I know this might be confusing, but this should be `i.spec.Recovery.Label`~ the comment is wrong, code was already fine, sorry 🙏🏽

pkg/action/install.go
```diff
@@ -205,6 +206,22 @@ func (i InstallAction) Run() (err error) {
...
+			"system_label":   constants.SystemLabel,
```
davidcassany: Edited: ~Then this one should be `i.spec.Partitions.Recovery.FilesystemLabel`~ this is also a wrong comment, sorry 🙏🏽 It should have been `i.spec.Partitions.Recovery.Images[constants.RecoveryImgName].Label`

pkg/action/reset.go
```diff
@@ -213,6 +213,22 @@ func (r ResetAction) Run() (err error) {
...
 	}
 
+	err = grub.SetPersistentVariables(
+		filepath.Join(r.spec.Partitions.State.MountPoint, cnst.GrubOEMEnv),
+		map[string]string{
+			"state_label":    r.spec.Partitions.State.FilesystemLabel,
+			"active_label":   r.spec.Active.Label,
+			"passive_label":  r.spec.Passive.Label,
+			"recovery_label": r.spec.Partitions.Recovery.FilesystemLabel,
+			"system_label":   cnst.SystemLabel,
```
davidcassany: Here applies the same as in install. But here is tricky, because there is no such a `r.spec.Recovery`, however such a value can be computed form the install state. So something like `r.spec.State.Partitions[cnst.RecoveryPartName].FSLabel`

pkg/action/install.go
```diff
@@ -205,6 +206,22 @@ func (i InstallAction) Run() (err error) {
...
+			"oem_label":      i.spec.Partitions.OEM.FilesystemLabel,
```
davidcassany: Persistent partition is missing here, should be added too. You have not really missed it, I noted this was also missing in the former hook.

pkg/action/upgrade.go
```diff
@@ -237,6 +237,22 @@ func (u *UpgradeAction) Run() (err error) {
 		return elementalError.NewFromError(err, elementalError.HookAfterUpgrade)
 	}
 
+	err = utils.NewGrub(&u.config.Config).SetPersistentVariables(
```
davidcassany, line 241: I have doubts about adding this as part of the upgrade... also has the upgrade spec all the values we need? We need to be careful here.

davidcassany (commented):
Beyond some details regarding specific recovery label values (yes, having recovery partitions and recovery images both with labels make it confusing), I  think this is good :+1:

Something I have been wondering about is if we should create some sort of interface with common methods for `InstallSpec`, `UpgradeSpec`, `ResetSpec` so we could, for instance, create a method that returns the labels map and just make use of with the grub object.

Also I am wondering if the setting of these persistent labels should be part of the installation method of grub and also have it as a devoted method in grub object. My idea is that within the actions logic this should be transparent or nearly transparent and keep required bootloader steps in there as a single block.

I am think about all this as I'd like to refactor elemental-cli grub logic at mid term to make it more as a library as a `bootloader` package and include in there the logic for live ISO, install/upgrade/reset and build-disks. So the more contained we have the logic of grub the better I'd say.

pkg/action/upgrade.go
frelon: Hmm, good point! I did run the upgrade on a elemental-teal install and the labels were set, but I did not try changing them from the defaults.

pkg/action/install.go
frelon: It felt like there was something missing! Will add that one as well!

pkg/action/upgrade.go
davidcassany: After a further thought I think this is fine running again the labels setting during the upgrade. Makes sense to keep it aligned. Only image labels are exchangeable during the upgrade, so updating the lables makes sense and I believe setting all labels together is just safer and simpler than trying to just apply the ones that actually change.

pkg/types/v1/config.go
```diff
@@ -369,7 +369,7 @@ func NewElementalPartitionsFromList(pl PartitionList, state *InstallState) Eleme
 	ep.OEM = pl.GetByNameOrLabel(constants.OEMPartName, lm[constants.OEMPartName])
 	ep.Recovery = pl.GetByNameOrLabel(constants.RecoveryPartName, lm[constants.RecoveryPartName])
 	ep.State = pl.GetByNameOrLabel(constants.StatePartName, lm[constants.StatePartName])
-	ep.Persistent = pl.GetByNameOrLabel(constants.PersistentPartName, lm[constants.PersistentLabel])
+	ep.Persistent = pl.GetByNameOrLabel(constants.PersistentPartName, lm[constants.PersistentPartName])
```
davidcassany, line 372: 😱 thanks!

pkg/action/install.go
```diff
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/rancher/elemental-cli/pkg/constants"
```
davidcassany: Note the line below already adds constants. I am the one to blame here, I named them `cnst` to type less when I started coding elemental-cli in several places. That was before I started using VSCode and its automatic imports, back then I was using `Vim` as my single editor, so typing less was relevant 😅 

Feel free to use `cnst` or `constants` at your choice, but I'd stick to one in the same file.

pkg/types/v1/grub.go
```diff
@@ -0,0 +1,80 @@
...
+limitations under the License.
+*/
+
+package v1
+
+import (
+	"github.com/rancher/elemental-cli/pkg/constants"
+)
+
+func (i InstallSpec) GetGrubLabels() map[string]string {
```
davidcassany, line 23: Yes, loving it, probably later on we can think on extending this file with further methods and create an interface for the methods, so we can imagine something like:

```golang
utils.NewGrub(&u.config.Config, spec)
```

so we could compute all those non obvious grub parameters as part of grub internals.

davidcassany (approved):
Awesome, looks good :+1:

pkg/action/install.go
frelon: Oops, good catch! Yep, I will stick to the one used in the file!

pkg/types/v1/grub.go
frelon: That's a great idea!

pkg/types/v1/config.go
frelon: Hehe, took some time before I realized why I got nil for the persistent partition everywhere :sweat_smile:

pkg/config/config.go
```diff
@@ -274,7 +274,7 @@ func NewInstallElementalParitions() v1.ElementalPartitions {
 func getActivePassiveAndRecoveryState(state *v1.InstallState) (active, passive, recovery *v1.ImageState) {
 	recovery = &v1.ImageState{
 		FS:    constants.LinuxImgFs,
-		Label: constants.RecoveryLabel,
+		Label: constants.SystemLabel,
```
frelon, line 277: @davidcassany FYI, I'm pretty sure this is the correct label, but better with two pair of eyes.

davidcassany: Indeed... 🤦🏽 damn! how bad I did it...

davidcassany:
@frelon I consider this ready to be merged. Thanks for the patience and for few unrelated fixed in the former code :+1:

PR #417, opened 2023-01-30
Make ISO related constants consistent

davidcassany:
Some constants were using `Iso` prefix while others `ISO`. This PR just makes use of  `ISO` for all.


frelon (approved):
Nice! :+1:

fgiudici (approved):

PR #418, opened 2023-02-01
Restructure grub logic into independent methods

davidcassany:
In addition this commit also drops setting a tty feature, basically for three reasons:

* Parsing and changing the grub.cfg is an error prone approach as elemental-cli does not control grub.cfg contents.
* There are easier and more robust ways to add extra kernel parameters using grub2 persistent variables.
* The current approach might not be really functional in all cases as it is based on host's /dev analysis, which is not necessarily the same as the target system.

### DIFF
--- a/pkg/utils/grub.go
+++ b/pkg/utils/grub.go
@@ -25,13 +25,24 @@ import (
 	"strings"
 
 	efilib "github.com/canonical/go-efilib"
-	"github.com/rancher/elemental-cli/pkg/constants"
 	cnst "github.com/rancher/elemental-cli/pkg/constants"
 	eleefi "github.com/rancher/elemental-cli/pkg/efi"
 	v1 "github.com/rancher/elemental-cli/pkg/types/v1"
 )
 
-const bootEntryName = "elemental-shim"
+const (
+	bootEntryName   = "elemental-shim"
+	grubConfDir     = "grub2"
+	entryEFIPath    = "/EFI/elemental"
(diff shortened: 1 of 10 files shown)


davidcassany:
The diff looks terribly huge, however if you give it a proper look you will see that this basically splits the `grub.Install` logic into three methods `grub.InstallBIOS`, `grub.InstallEFI` and  `grub.InstallConfig`, each of this three methods keeps the original logic with just few clean ups to make use of the `utils.CopyFile` to copy files and several paths and literals converted into internal constants. Beyond that is a copy & paste of the same exact code and logic but in different sections of the same file.

See the unit tests are unchanged and the diff coverage is high. The logic is really the same.

davidcassany:
Easier to review by separate commits, the first does the grub code re-structure and the second deprecates the `tty` flag and removes the field from reset and install specs.

pkg/utils/grub.go
```diff
@@ -46,246 +57,240 @@ func NewGrub(config *v1.Config) *Grub {
 	return g
 }
 
-// Install installs grub into the device, copy the config file and add any extra TTY to grub
-func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool, stateLabel string, disableBootEntry bool, clearBootEntries bool) (err error) { // nolint:gocyclo
```
frelon, line 50: Really nice to be rid of all the bool arguments! :+1:

frelon (approved):
LGTM!

PR #420, opened 2023-02-02
Small refactor to include CommandExists method in Runner interface

davidcassany:

### DIFF
--- a/pkg/types/v1/runner.go
+++ b/pkg/types/v1/runner.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1
 
 import (
+	"fmt"
 	"os/exec"
 	"strings"
 )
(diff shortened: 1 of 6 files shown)


pkg/types/v1/runner.go
```diff
@@ -42,10 +49,13 @@ func (r RealRunner) RunCmd(cmd *exec.Cmd) ([]byte, error) {
...
-	if r.Logger != nil {
-		r.Logger.Debugf("Running cmd: '%s %s'", command, strings.Join(args, " "))
+	if !r.CommandExists(command) {
+		err := fmt.Errorf("Command %s not found", command)
+		r.error(err.Error())
+		return []byte{}, err
 	}
+	r.debug(fmt.Sprintf("Running cmd: '%s %s'", command, strings.Join(args, " ")))
+	cmd := r.InitCmd(command, args...)
 	return r.RunCmd(cmd)
```
frelon: RunCmd will return an ErrNotFound if the command is not found, so the new test passes with this change instead:

```
        bytes, err := r.RunCmd(cmd)
        if err != nil {
            r.Logger.Errorf("Error running command: %s", err.Error())
        }
        
	return bytes, err
```

frelon (commented):
I don't think it's needed to use the CommandExists method in RealRunner Run-method, otherwise looks fine :+1:

pkg/types/v1/runner.go
davidcassany: Right, you got a point, gonna adapt it

frelon (approved):
LGTM

PR #421, opened 2023-02-02
Refactor cloud-init paths setup

davidcassany:
This commit includes the default cloud-init paths as part of the default configuration. In addition the cloud init paths are moved to the main configuration struct, as the could potentially be used in build configuration struct too.

In fact, the default cloud-init paths are only applied for the runtime configuration and not for the build time configuration (at build time the host is not necessarily an Elemental based OS).


frelon (approved):
LGTM

PR #422, opened 2023-02-06
Build disk cmd recode

davidcassany:
This  a full rewrite of the `build-disk` command. It mostly introduces a new config schema for the command that perfectly aligned with the install/upgrade and reset commands. That means specific images for recovery and active can be set as well as custom partition size, filesystem and most of the details that are already possible within the install command.
In addition it also includes a `--recovery-only` flag to create expandable disk images. Expandable images only include a recovery partition and image plus a mini state partition including few config files, but no OS. The images are already created with a preloaded cloud-init setup to expand state partition to the requested size and also create a follow up persistent partition with the configured size. Once partitions are created the system runs a `reset` command to get a full defined system.

As a consequence of better coding reuse and the introduction of the `sgdisk` wrapper the build disk procedure can be done with a more traditional approach (loop device mount and dump data in it) requiring privileges when running inside a container or using a mount-less approach, so it can be build within a container without additional privileges under certain constraints. Both paths are sharing most of the code so from a maintainability perspective this has almost no impact.


davidcassany:
All unit tests are missing, so that this is a draft. This is one of the outcomes of my hackweek 😄 

I did few non extensive tests using this build command within the elemental-cli builder image and I could manage to build disk images in on privileged containers. A more traditional procedure using mounts is also possible and functional.

PR #425, opened 2023-02-16
Remove squashfs specific path

davidcassany:
This commit removes the existence and checks for recovery.squashfs. It always uses recovery.img as the recovery image file regardless of the underlying filesystem.

In addition it also removes the checks for a separate recovery image in ISOs. We are no longer building isos with separate recovery images. The same feature is also possible by setting recovery to the appropriate source (local or remote) in config.yaml.

Part of rancher/elemental#488


kkaempf (approved):
code cleanup, ftw !

PR #428, opened 2023-02-17
Add init command

frelon:
The init command replaces the `system/cos-setup`,
`system/base-dracut-modules` and `system/grub2-config` packages.

It does this by:

- Installing systemd units for the boot-process
- Installing dracut modules and configuration
- Setting GRUB2 configuration and next_entry config

Adapting `cloud-config/*` and `system/immutable-rootfs` packages to elemental-cli would allow us to write a derivative Dockerfile like the following:

```Dockerfile
FROM rancher/elemental-cli as CLI
FROM opensuse/tumbleweed as OS

# install kernel, systemd, dracut, grub2 and any other handy tool
RUN zypper install --non-interactive <list of system packages>

COPY --from=CLI /usr/bin/elemental /usr/bin/elemental
RUN elemental init
```

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -233,6 +233,24 @@ func ReadInstallSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.InstallSpec, er
 	return install, err
 }
 
+func ReadInitSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.InitSpec, error) {
+	init := config.NewInitSpec(r.Config)
+	vp := viper.Sub("init")
+	if vp == nil {
+		vp = viper.New()
+	}
+	// Bind install cmd flags
+	bindGivenFlags(vp, flags)
+	// Bind install env vars
+	viperReadEnv(vp, "INIT", constants.GetInitKeyEnvMap())
+
+	err := vp.Unmarshal(init, setDecoder, decodeHook)
+	if err != nil {
+		r.Logger.Warnf("error unmarshalling InitSpec: %s", err)
+	}
+	return init, err
+}
+
 func ReadResetSpec(r *v1.RunConfig, flags *pflag.FlagSet) (*v1.ResetSpec, error) {
 	reset, err := config.NewResetSpec(r.Config)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


davidcassany:
@frelon would you mind putting this PR on hold? I have concerns how to properly handle this.

IMHO this falls in the scope of merging elemental-cli and elemental-toolkit into a single repository. I think we should always keep the possibility of providing our own files too and I am not sure how to make this evident and flexible with the current implementation, probably subinit commands? something like `elemental init grub` (just for gurb) or `elemental init init-services` (just for the cloud-init like unit services)?

Then I am also not sure if this is something we want an elemental-cli sub command, what if this is an extra small utility (within this repo but a separare binary), so it is less risky that anyone in its host calls it (tells the man that executed ignition and combustion in its development host by mistake 😅 )

Probably we should have discussion around how to proceed towards that goal in some small organized steps in order to merge the two repositories without having a messy transition. I am thinking about it.

frelon:
> @frelon would you mind putting this PR on hold? I have concerns how to properly handle this.

Sure! Just something I came up with during hackweek to quicker iterate on the packages.

> IMHO this falls in the scope of merging elemental-cli and elemental-toolkit into a single repository. I think we should always keep the possibility of providing our own files too and I am not sure how to make this evident and flexible with the current implementation, probably subinit commands? something like `elemental init grub` (just for gurb) or `elemental init init-services` (just for the cloud-init like unit services)?

Yep, this is simply a wrapper around copying in files and running `mkinitrd` so anyone using wanting their own files could simply copy them and run `mkinitrd` in the same way as before. This was a timesaver when rebuilding a derivative (making changes to the "packages") and not having to use a slow package-manager.

> Then I am also not sure if this is something we want an elemental-cli sub command, what if this is an extra small utility (within this repo but a separare binary), so it is less risky that anyone in its host calls it (tells the man that executed ignition and combustion in its development host by mistake sweat_smile )

Fair enough, the command does check if we are running in a docker-build environment and errors out otherwise.

> Probably we should have discussion around how to proceed towards that goal in some small organized steps in order to merge the two repositories without having a messy transition. I am thinking about it.

Great, looking forward to it!

PR #432, opened 2023-02-21
Update go version to 1.18

frelon:
This PR also adds an updatecli config that keeps our github workflows and the Dockerfile up to date with the go-version specified in `go.mod`.


frelon:
Updatecli messes with the yaml indent but did not find a way to fix that...

PR #433, opened 2023-02-21
Support Device Mappers

paynejacob:
PROBLEM:

The library that was previously being used to detect block devices did not include mapped devices.  This meant that if a users were to have a custom disk layout some features such as reset would not work.  For example if  user were to set up their persistent partition with LUKS they would have a device mapped to the encrypted partition with the `COS_PERSISTENT` label.  This worked fine for normal operations but caused issues with reset because reset would not see the labeled map device and fall back to the named partition.  It would not recognize the filesystem and fail.

SOLUTION:

Rather than use the previous go library to detect block devices I switched to a subprocess call to lsblk.  lsblk is able to list mapped devices allowing elemental to format the labeled mapper device rather that the partition it's self.

### DIFF
--- a/pkg/types/v1/config.go
+++ b/pkg/types/v1/config.go
@@ -369,7 +369,7 @@ func NewElementalPartitionsFromList(pl PartitionList, state *InstallState) Eleme
 	ep.OEM = pl.GetByNameOrLabel(constants.OEMPartName, lm[constants.OEMPartName])
 	ep.Recovery = pl.GetByNameOrLabel(constants.RecoveryPartName, lm[constants.RecoveryPartName])
 	ep.State = pl.GetByNameOrLabel(constants.StatePartName, lm[constants.StatePartName])
-	ep.Persistent = pl.GetByNameOrLabel(constants.PersistentPartName, lm[constants.PersistentLabel])
+	ep.Persistent = pl.GetByNameOrLabel(constants.PersistentPartName, lm[constants.PersistentPartName])
 
 	return ep
 }
--- a/pkg/utils/getpartitions.go
+++ b/pkg/utils/getpartitions.go
@@ -17,8 +17,12 @@ limitations under the License.
 package utils
 
 import (
+	"bytes"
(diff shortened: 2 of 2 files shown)


davidcassany (commented):
Thanks @paynejacob for the contribution. 

So far I don't have a strong opinion on weather we should keep using the ghw library or move to lsblk. In any case this is a quite sensitive change as it impacts in many places and a bunch of tests need to be adapted, I believe there might be quite some here. If lsblk is more flexible and provides better support in a variety of cases then I'd say the change is reasonable.

We used to have a pretty similar implementation of the same here https://github.com/rancher/elemental-cli/commit/2dc11d6295b23f2c5c157121dbc33736b08cc79a#diff-b862f8327f8d9213360241fb4b09183f1eec026250afcd5a5a8673306002ff33 with few, but not sure if relevant, differences.

What I first notice is that you omit any loop and crypt device, which is the specific reason for that? This invalidates any installation over loop devices, which is really handy for testing purposes and probably something to explore in the future for a lightweight disk build process.

In our old implementation we used to only keep part and loop devices, only, so all others were ignored including the root device. If I understood it right one of the problems you have is the pkname device of a luks device is not the disk itself, there is one step in between. However is that required to fetch the root disk for the luks device? I feel more comfortable with  white list approach for types rather than a black list, so more in favor of explicitly detecting the particular cases supported and add few additional logic for those if required (like finding the appropriate root device as you did).


I am also curious about the lsblk output of for the use case you are targeting, do you have any at hand to share?

kkaempf:
ping @paynejacob - can you please fix the failing test ?

paynejacob:
The motivation for filtering out crypt devices was driven by how elemental detects partitions, it will favor the part label over the disk label.  When you have an encrypted partition it will have the elemental part label but the disk label will be on the mapper device.  Filtering out loop devices was an oversite and should be removed.

I also think there is a way to do this with out relying on lsblk.  I will create another pr with these changes.

PR #442, opened 2023-02-28
Bump ginkgo/gomega

frelon:


frelon:
The large amount of new vendored files seems to be because a change in ginkgo pulls in all their sub-packages as vendored, which they did not do before.

davidcassany (approved):
LGTM

PR #444, opened 2023-03-01
Delete new-cmd

frelon:
The new command is quite old and not used much anymore. It's also implemented with huge third-party dependencies that are only used by that command.

This commit removes the new-cmd and the dependencies.


frelon:
@Ludea I saw you implemented this command, is it something you are still using?

Ludea:
I don't use that command anymore

davidcassany (approved):
LGTM

PR #448, opened 2023-03-03
Fix updatecli workflow

frelon:
Do not update workflow files, since the bot does not have write permissions there.


davidcassany (approved):
LGTM

PR #453, opened 2023-03-08
Add cmdline argument to ISO

frelon:
Add the `rd.live.overlay.overlayfs=1` cmdline argument since it seems to be needed to boot when building a tumbleweed ISO.

### DIFF
--- a/pkg/live/common.go
+++ b/pkg/live/common.go
@@ -63,7 +63,7 @@ const (
 
 	menuentry "%s" --class os --unrestricted {
 		echo Loading kernel...
-		$linux ($root)` + constants.ISOKernelPath + ` cdroot root=live:CDLABEL=%s rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0 rd.cos.disable cos.setup=` + constants.ISOCloudInitPath + `
+		$linux ($root)` + constants.ISOKernelPath + ` cdroot root=live:CDLABEL=%s rd.live.dir=/ rd.live.squashimg=rootfs.squashfs rd.live.overlay.overlayfs=1 console=tty1 console=ttyS0 rd.cos.disable cos.setup=` + constants.ISOCloudInitPath + `
 		echo Loading initrd...
 		$initrd ($root)` + constants.ISOInitrdPath + `
 	}                                                                               


frelon:
Would be nice to be able to send in or override the cmdline params as well.. food for thought!

pkg/live/common.go
```diff
@@ -63,7 +63,7 @@ const (
 
 	menuentry "%s" --class os --unrestricted {
 		echo Loading kernel...
-		$linux ($root)` + constants.ISOKernelPath + ` cdroot root=live:CDLABEL=%s rd.live.dir=/ rd.live.squashimg=rootfs.squashfs console=tty1 console=ttyS0 rd.cos.disable cos.setup=` + constants.ISOCloudInitPath + `
+		$linux ($root)` + constants.ISOKernelPath + ` cdroot root=live:CDLABEL=%s rd.live.dir=/ rd.live.squashimg=rootfs.squashfs rd.live.overlay.overlayfs=1 console=tty1 console=ttyS0 rd.cos.disable cos.setup=` + constants.ISOCloudInitPath + `
```
davidcassany, line 66: I wonder which is the TW alternative to overlayfs 🤔

davidcassany (commented):
In fact, as you realized the ideal would be making all this configurable. I'd say this would be nice to configure it in two different parameters: predifined values and extra values. Predefined ones would be a parameter that defaults to the current kernel arguments. Extra values would default to an empty set. So from a user perspective you could append or overwrite the kernel arguments.

I'd like to bring a similar concept for the install and reset commands.

frelon:
> In fact, as you realized the ideal would be making all this configurable. I'd say this would be nice to configure it in two different parameters: predifined values and extra values. Predefined ones would be a parameter that defaults to the current kernel arguments. Extra values would default to an empty set. So from a user perspective you could append or overwrite the kernel arguments.
> 
> I'd like to bring a similar concept for the install and reset commands.

Yep, that sounds great! I will put up an issue for this instead. As you said we should also see how this plays out with install/upgrade :+1:

kkaempf:
:+1: on making these arguments better configurable - see also rancher/elemental#550

kkaempf:
Oh, there's also rancher/elemental-toolkit#1773

PR #457, opened 2023-03-16
Bump to yip v1.0.2

davidcassany:
This includes a fix in yip dependencies for an old datasource related issue, where the cdrom datasource was leaving behind all devices opened and hence elemental-cli was failing to partition on install due to busy devices.


cmd/cloud-init_test.go
```diff
@@ -49,7 +49,7 @@ var _ = Describe("cloud-init", Label("cloud-init", "cmd"), func() {
 					"-s",
 					"tests",
 					"-d",
-					`stages.tests=foo`,
+					`'stages.tests=foo'`,
```
fgiudici: nit: should we keep the same format for quoting ?
```suggestion
					"'stages.tests=foo'",
```

pkg/utils/runstage_test.go
```diff
@@ -34,6 +34,13 @@ import (
 	"github.com/twpayne/go-vfs/vfst"
 )
 
+const testingStages = `
+stages:
+  luke:
+  - commands:
+    - echo "I have a very bad feeling about this"
```
fgiudici, line 41: 🤣

fgiudici (approved):
LGTM

PR #460, opened 2023-03-23
Linting

frelon:


davidcassany (approved):
LGTM

pkg/types/v1/logger.go
```diff
@@ -166,4 +166,4 @@ func convert(args []interface{}) string {
 func (w *logrusWrapper) SetContext(string) {}
 func (w *logrusWrapper) Spinner()          {}
 func (w *logrusWrapper) SpinnerStop()      {}
-func (w *logrusWrapper) Screen(t string)   {}
+func (w *logrusWrapper) Screen(_ string)   {}
```
fgiudici, line 169: nit, format inconsistency: few lines above ( line 166), we  use unnamed parameter, here we use the blank identifier.

pkg/types/v1/logger.go
```diff
@@ -112,8 +112,8 @@ func (w *logrusWrapper) Ask() bool {
 
 func (w *logrusWrapper) Copy() (types.Logger, error) {
 	c := *w
-	copy := &c
-	return copy, nil
+	copied := &c
+	return copied, nil
```
fgiudici, line 116: 🤔 just curious, was the linter complaining here for the var name?

fgiudici (approved):
looks good here!

pkg/types/v1/logger.go
frelon: Yep, it's shadowing the builtin copy function.

PR #465, opened 2023-03-30
Update to go 1.19

frelon:
This should hopefully make our ci less brittle when new versions of ginkgo and gomega are released


davidcassany (approved):
LGTM

PR #466, opened 2023-03-31
Reduce default partition sizes

davidcassany:
This PR reduces default partition sizes for state and recovery partitions. State was set up to 15GB and it is mostly including active and passive images, which are about 1.3GB  each (they include a 256Mb overhead of free space). So there was way too much space reserved for them. Similar applies to Recovery image, originally set to 8GB and it only includes the recovery image which around 1.3GB again.

This PR reduces the sizes to the half, 8GB for state partition and 4GB for recovery partition. 

The theoretic minimum sizes are 4xOSsize for State partition and 3xOSsize on recovery.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -53,8 +53,8 @@ const (
 	BiosFs             = ""
 	EfiSize            = uint(64)
 	OEMSize            = uint(64)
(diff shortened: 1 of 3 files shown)


kkaempf (approved):
🤞🏻

PR #471, opened 2023-04-11
Deprecate disk related commands

### LINKED ISSUE #470: elemental-cli build-disk command
should be deprecated

davidcassany:
With the change the commands do not appear in `elemental --help` and if used they run as they used to but with the deprecation warning.

```
$ elemental build-disk
Command "build-disk" is deprecated, it and can be changed or removed without a major version bump
...
```

Fixes #470


frelon (approved):
LGTM

PR #473, opened 2023-04-11
fix typo in deprecation text

kkaempf:
Before:

> $ elemental build-disk
Command "build-disk" is deprecated, it and can be changed or removed without a major version bump

After:

> $ elemental build-disk
Command "build-disk" is deprecated, it can be changed or removed without a major version bump


davidcassany (approved):
ouch 🤦🏽  thanks @kkaempf

PR #477, opened 2023-04-21
Fix version command to proper show version and commit

davidcassany:
This ensures the elemental-cli binaries include the proper version form the tag they are build and the commit hash.

This was fixes long a go for elemental-operator, apparently we did not fix it for the elemental-cli. This PR is applying the same logic as in elemental-operator spec. Note that this will require a small arrangement in `_service` file in OBS too. OBS SR is also ready [here](https://build.opensuse.org/request/show/1081476).


kkaempf (approved):
🤩

PR #478, opened 2023-04-24
Add multi-arch support for pulling images

### LINKED ISSUE #452: Incorrect image is pulled when referencing a multi-arch manifest
Elemental CLI should use the context of the host machine's CPU architecture when pulling an image for upgrade or install. 

Instead, elemental-cli is defaulting to the first or perhaps the linux/amd64 image listed in the manifest and is not guaranteed to match the target system.

This limitation forces derivatives to split images into separate repositories within a container registry and is arguably more messy to maintain.

From a UX perspective, the consumer of a derivative should be able to call `elemental upgrade --system.uri=docker:myimage:latest` without worrying about whether they selected the correct architecture for the machine they are on.

For use-cases outside upgrade or install, the arch should be override-able to support building cloud images for ARM64 instances from an AMD64 builder or vice-versa. (although it's possible that install hooks may fail in this scenario so emulation/virtualization may be necessary anyway)

---

Screenshot Example
failed elemental upgrade due to exec format error

An exec format error is shown when after-upgrade hooks are attempted, because elemental cli pulled an amd64 image on an arm64 host.

---

P.S. I'm working on a derivative project that produces multi-arch image tags, but my images are not yet stable. So if anyone would like to reproduce this bug, you can try pulling one of my images but it's not guaranteed to leave you with a functioning system. (https://github.com/jtcressy-home/edgenode/pkgs/container/edgenode)

frelon:
This commit refactors the image pulling and extracting to be able to cross-build isos.

In order to do this we make use of go-containerregistry and containerd to pull and extract the image.

This refactor also removes alot of luet functionality mostly used for `build-disk`  and `convert-disk` commands that have been deprecated.


pkg/types/v1/image.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/daemon"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+type ImageExtractor interface {
+	ExtractImage(imageRef, destination, platformRef string, local bool) error
```
davidcassany, line 33: @frelon have you checked if the library easily supports pulling from a private registry. I guess it does. I'd say even if this is not coded now we should consider this use case and check no big API changes will be needed if such a feature is added later on.

pkg/types/v1/common.go
```diff
@@ -44,14 +43,10 @@ func (i ImageSource) Value() string {
 	return i.source
 }
 
-func (i ImageSource) IsDocker() bool {
+func (i ImageSource) IsImage() bool {
```
davidcassany, line 46: I like the rename

pkg/types/v1/image.go
```diff
@@ -0,0 +1,65 @@
...
+
+	reader := mutate.Extract(image)
+
+	_, err = archive.Apply(context.Background(), destination, reader)
+	return err
+}
+
+func image(ref name.Reference, platform v1.Platform, local bool) (v1.Image, error) {
+	if local {
+		return daemon.Image(ref)
```
davidcassany, line 64: Do we know how this works? Does it support containerd? Would be really cool if that would also be functional for containerd too.

tests/fixtures/config/config.yaml
```diff
@@ -22,6 +17,6 @@ reset:
 
 upgrade:
   system:
-    uri: channel:system/cos
+    uri: docker:some/image:latest
```
davidcassany: `some/image:latest` should also be fine, as `docker:` or `oci:` type is the expected default.

davidcassany (commented):
Really interesting, nice job! I like it. We have to carefully consider this change as it affects in several places and drops some (unused) features. 

Out of curiosity? Have you compared the size of the out coming binary? I'd be curious to check of there is a significant change.

It is relevant to note this change incompatible with the current elemental-toolkit repository structure. It would require adopting the restructure from davidcassany/elemental-toolkit#3 and the first half of rancher/elemental-toolkit#1738

frelon:
> Really interesting, nice job! I like it. We have to carefully consider this change as it affects in several places and drops some (unused) features.

Nice! Yep, it became bigger and bigger since the dropped dependencies used alot of old packages in turn. I will run more tests this week but an initial build and boot of elemental-teal was successful.

> Out of curiosity? Have you compared the size of the out coming binary? I'd be curious to check of there is a significant change.

I did a quick check and it seems we go from ~24MB to ~18MB.

> It is relevant to note this change incompatible with the current elemental-toolkit repository structure. It would require adopting the restructure from https://github.com/davidcassany/elemental-toolkit/pull/3 and the first half of https://github.com/rancher/elemental-toolkit/issues/1738

Hmm, I'm probably missing something but I think this change should be compatible with only some minor modifications to the toolkit?

pkg/types/v1/image.go
frelon: Hmm, it does not seem to support containerd out-of-the-box but might be able to just point it to the correct socket. The library also has abstracted this, so it's possible to write a custom client to fetch the image. I will experiment a bit with this!

davidcassany:
> Hmm, I'm probably missing something but I think this change should be compatible with only some minor modifications to the toolkit?

Well, depends on what minor means 😅 I bet the CI will have several breaks here and there with such a change included within toolkit, as the CI relies on channels in some places. And again, I don't think spending time to adapt the CI there makes much sense, better rearrange it not to be a Luet centric project first, then it should not much impact if any at all. You already know how maddening it can be the toolkit CI 😅

pkg/types/v1/image.go
frelon: It's the same library as used before, but a newer version so it should be possible to pull from any registry. Will have to bring back the authentication though, as I ripped that out in the refactor. Will test that also!

frelon:
> > Hmm, I'm probably missing something but I think this change should be compatible with only some minor modifications to the toolkit?
> 
> Well, depends on what minor means sweat_smile I bet the CI will have several breaks here and there with such a change included within toolkit, as the CI relies on channels in some places. And again, I don't think spending time to adapt the CI there makes much sense, better rearrange it not to be a Luet centric project first, then it should not much impact if any at all. You already know how maddening it can be the toolkit CI sweat_smile

Haha, fair enough! :sweat_smile:

pkg/types/v1/image.go
frelon: Yep, the latest commit adds authentication and I also tried running a local registry and pulling from that, which also works.

cmd/flags.go
```diff
@@ -139,10 +140,9 @@ func validateInstallUpgradeFlags(log v1.Logger, flags *pflag.FlagSet) error {
 	return validatePowerFlags(log, flags)
 }
 
-// addArchFlags adds the arch flag for build commands
-func addArchFlags(cmd *cobra.Command) {
-	archType := newEnumFlag([]string{"x86_64", "arm64"}, "x86_64")
-	cmd.Flags().VarP(archType, "arch", "a", "Arch to build the image for")
+// addPlatformFlags adds the arch flag for build commands
+func addPlatformFlags(cmd *cobra.Command) {
```
davidcassany, line 144: Shouldn't we add a conversion between to two and deprecate `arch`? Note this also affects the `manifest.yaml` and `config.yaml`.

cmd/config/config_test.go
```diff
@@ -76,15 +78,15 @@ var _ = Describe("Config", Label("config"), func() {
 		var flags *pflag.FlagSet
 		BeforeEach(func() {
 			flags = pflag.NewFlagSet("testflags", 1)
-			flags.String("arch", "", "testing flag")
-			flags.Set("arch", "arm64")
+			flags.String("platform", "", "testing flag")
+			flags.Set("platform", "linux/arm64")
```
davidcassany, line 100: wouldn't it be an option to always assume `linux/<arch>` and keep around the arch parameter? Probably this is easier than coding `arch` backward compatibility and intruducing the new platform parameter.

cmd/config/config_test.go
```diff
@@ -156,20 +155,10 @@ var _ = Describe("Config", Label("config"), func() {
 				Expect(err).ShouldNot(HaveOccurred())
 
 				// From config file
-				Expect(iso.Image[0].Value()).To(Equal("recovery/cos-img"))
+				Expect(iso.BootloaderInRootFs).To(BeTrue())
```
davidcassany: I'd keep it as a docker image instead. So we ensure only the ` channel` is gone.

tests/fixtures/config/manifest.yaml
```diff
@@ -1,51 +1,12 @@
 iso:
-  rootfs:
-    - channel:system/cos
-  uefi:
-    - channel:live/grub2-efi-image
-  image:
-    - channel:recovery/cos-img
```
davidcassany, line 7: I'd just change them from `channel:..` to `oci:...` or `dir:...`. So we still test `image`, `uefi` and `rootfs` parameters in config.

pkg/constants/constants.go
```diff
@@ -103,13 +103,6 @@ const (
 	PostUpgradeHook        = "post-upgrade"
 	BeforeUpgradeHook      = "before-upgrade"
 
-	// Luet constants
-	LuetCosignPlugin    = "luet-cosign"
-	LuetMtreePlugin     = "luet-mtree"
-	LuetDefaultRepoURI  = "quay.io/costoolkit/releases-green"
-	LuetRepoMaxPrio     = 1
-	LuetDefaultRepoPrio = 90
-
```
davidcassany, line 112: Also the `GetBuildDiskDefaultPackages` should be gone here.

pkg/types/v1/config.go
```diff
@@ -46,18 +46,17 @@ type Config struct {
...
-	Arch                      string       `yaml:"arch,omitempty" mapstructure:"arch"`
-	SquashFsCompressionConfig []string     `yaml:"squash-compression,omitempty" mapstructure:"squash-compression"`
-	SquashFsNoCompression     bool         `yaml:"squash-no-compression,omitempty" mapstructure:"squash-no-compression"`
-	CloudInitPaths            []string     `yaml:"cloud-init-paths,omitempty" mapstructure:"cloud-init-paths"`
-	Strict                    bool         `yaml:"strict,omitempty" mapstructure:"strict"`
+	Cosign                    bool     `yaml:"cosign,omitempty" mapstructure:"cosign"`
+	Verify                    bool     `yaml:"verify,omitempty" mapstructure:"verify"`
+	CosignPubKey              string   `yaml:"cosign-key,omitempty" mapstructure:"cosign-key"`
+	LocalImage                bool     `yaml:"local,omitempty" mapstructure:"local"`
+	Platform                  string   `yaml:"platform,omitempty" mapstructure:"platform"`
```
davidcassany: We need a migration path from `Arch` to  `Platform`. Note this impacts the ` config.yaml` and `manifest.yaml`. Is there any reason to change `Arch` to `Platform`? I still wonder if shouldn't we simply translate arch to platform internally, after all I doubt we will ever consider anything else than `linux` as a platform.

tests/fixtures/config/manifest.yaml
```diff
@@ -1,51 +1,12 @@
...
+  bootloader-in-rootfs: true
   label: "LIVE_LABEL"
 
 # Raw disk creation values start
 
-
 raw_disk:
```
davidcassany: this should be gone too I think, the raw_disk is the data structure associated with the `build-disk` command.

davidcassany (changes requested):
Looks good so far. However I am concerned about the platform vs arch change. We need to be backward compatible in that regard. The only non backward compatible changes we should consider are related to the channels usage.

I don't think we will ever consider a platform that is no linux, so I am wondering about keeping arch field instead of platform and code translators, we already have `ArchToGolangArch` and `GolangArchToArch` in utils. We could have something like that to translate arch to platfrom and vice versa. In we could simply use these methods to code a couple of new ones `PlatformToArch` and `ArchToPlatform`. I think this would be the easiest way to be backward compatible. What you think?

Beyond the arch vs platform change all looks good from my point of view, great job :+1:

pkg/types/v1/config.go
```diff
@@ -48,15 +49,17 @@ type Config struct {
...
-	SquashFsNoCompression     bool     `yaml:"squash-no-compression,omitempty" mapstructure:"squash-no-compression"`
-	CloudInitPaths            []string `yaml:"cloud-init-paths,omitempty" mapstructure:"cloud-init-paths"`
-	Strict                    bool     `yaml:"strict,omitempty" mapstructure:"strict"`
+	Platform                  *Platform `yaml:"-" mapstructure:"-"`
+	Cosign                    bool      `yaml:"cosign,omitempty" mapstructure:"cosign"`
+	Verify                    bool      `yaml:"verify,omitempty" mapstructure:"verify"`
+	CosignPubKey              string    `yaml:"cosign-key,omitempty" mapstructure:"cosign-key"`
+	LocalImage                bool      `yaml:"local,omitempty" mapstructure:"local"`
+	Arch                      string    `yaml:"arch,omitempty" mapstructure:"arch"`
+	PlatformStr               string    `yaml:"platform,omitempty" mapstructure:"platform"`
```
davidcassany: I am not sure I understand the presence of Arch, PlatformStr and Platform here 🤔 I wonder if this could be re-arranged to some extend by a using a `CustomtUnmarshal` as in `types/v1/common.go`  and directly map `platform` yaml or flags to a Platform struct and avoid having the PlatformStr here. Then in sanitize we could use `arch`  as `platform` if this was not defined. So once we completely remove `arch` parameter we are only left with the `Platform` in struct.

frelon: Yep, platformstr was superfluous, added custom unmarshalling to Platform in latest commit instead. :+1:

davidcassany (approved):
Looks good to me now :+1:  Good job

We should be try not to add new (unrelated) stuff on top of that until we have verified it in `Dev` CI. Once verified we should tag a new version, so we have a clean cut of the change.

PR #484, opened 2023-05-11
Add init cmd

### LINKED ISSUE #1766: Embed config files within elemental-cli
Embed config files within elemental-cli, that way a single RPM is required to build an OS on top of a compatible distro. Consider this other PoC derived from hackweek project.

frelon:
The init command is meant to be run in a container build step and extracts config-files needed for an elemental derivative.

This PR includes a way to turn on/off different features of the elemental-toolkit. The standard features included here are:
* immutable-rootfs
* grub-config
* elemental-setup
* dracut-config
* cloud-config

An example dockerfile:

```
FROM rancher/elemental-cli:latest AS elemental-cli
FROM opensuse/tumbleweed as default

RUN zypper install -y grub2 grub2-x86_64-efi grub2-x86_64-efi-extras shim dracut kernel kernel-firmware-all systemd \
                      bash syslinux lvm2 parted dosfstools e2fsprogs rsync util-linux-systemd coreutils \
                      squashfs NetworkManager device-mapper iproute2 tar curl ca-certificates ca-certificates-mozilla \
                      procps openssl openssh vim-small less iputils bind-utils wget jq

COPY --from=elemental-cli /usr/bin/elemental /usr/bin/elemental

RUN systemctl enable NetworkManager
RUN elemental init --force --features immutable-rootfs,grub-config,dracut-config
```

Fixes rancher/elemental-toolkit#1766


davidcassany:
Structure wise I am not so sure about the structure of this PR. I mean mostly about the files from toolkit, specially if both get merged.

I was thinking if something like renaming the `toolkit` folder of `elemental-tookit` repository to `resources` (or `features`) and having this folder directly under `pkg` of `elemental-cli`. So something like:

```
pkg/
  ...
  resources/
    dracut-config/
    grub/
    immutable-rootfs/
    init-setup/
    init-config/
  ...
```

So I'd try to keep the folder structure as some sort of _package_ concept. Probably we could pass packages as arguments instead of flags  and have some sort of `--all` flag to just include them all. We can also reconsider the names if we see better fit.

How this PR relates to rancher/elemental-toolkit#1763? Feels like this is on the opposite direction (I mean importing from toolkit to here). I'd say rancher/elemental-toolkit#1763 goes first and then somehow this is ported to the merged repository, is this how you imagine it?

frelon:
@davidcassany interesting take, rancher/elemental-toolkit#1763 does not really have a dependency on this PR, depending on which is merged first we would only have to `git mv` the `toolkit/` files to the embedded location to keep the history.

I kept the embedded files in the features-package since I thought it was a nicer interface to the rest of the system, just to run `features.Get` and not have to deal with raw files outside of the package.

Also `go:embed` can only embed files from the current dir or a child-dir, which means we would put the embedded FS in something like `pkg/resources.go` in the `pkg` package, which seems a bit confusing to me.

Let's have a chat today to move things forward! :+1:

davidcassany:
Something else I realized that would be nice if it is covered by the `init` command is the creation of certain elemental-specific folders. More specific in our green example we have:

```
# Create non FHS paths
RUN mkdir -p /oem /system

# Add elemental config dir
RUN mkdir -p /etc/elemental/config.d
```
Would be nice if `init` also covers those. So we remove these lines from the Dockerfile :wink:

frelon:
Closing this in favor of rancher/elemental-toolkit#1767

PR #491, opened 2023-05-12
Bump yip

davidcassany:
This new version of yip is already making use of rancher-sandbox/linuxkit and including a fix to not ignore steps having a non unique name.

With this version it will be possible to pull cloud-config from a vfat device labeled with `cidata` or `CIDATA`.


frelon (approved):
LGTM!

